PR #17, opened 2022-10-24
update musica-core lib; switch to util-only build

mattldawson:
updates the `musica-core` library and does a "utility only" build


sjsprecious (approved):
Thanks for this fix. Now I am able to compile MICM with nvhpc compiler.

PR #30, opened 2023-01-04
29 add documentation with doxygen breathe and sphinx

### LINKED ISSUE #29: Add documentation with doxygen, breathe, and sphinx

K20shores:
Closes #29

I setup several of the classes incorrectly. They should have been templated but were not. Adding the templates means that most of the files are now inherently header only so most of their source files were removed. This may be my misunderstanding of what is meant to be templated though.


include/micm/system/phase.hpp
```diff
@@ -10,18 +10,35 @@
...
     Phase(const Phase& other);
-    Phase(std::vector<Species> species);
+    /// @brief Create a phase with a set of species
+    /// @param species A unique list of species
+    Phase(std::vector<Species<T>> species);
 
+    /// @brief Assignment operator. This assumes ownership of the other's list of species
+    /// @param other Another micm::Phase
+    /// @return a reference to the created phase
     Phase& operator=(const Phase& other);
```
mattldawson, line 38: should these be defined here now that the `cpp` file has been removed?

K20shores: I hadn't because I figured defining these would have been writing code and I was planning on leaving that for later. They will have very simple implementations so I can define them now and fully write the tests for them if you'd like me to

test/unit/process/test_intraphase_process_builder.cpp
```diff
@@ -5,16 +5,16 @@
...
 TEST(IntraPhaseProcessBuilder, CanChainBuilder){
-  micm::IntraPhaseProcessBuilder builder{};
+  micm::IntraPhaseProcessBuilder<double> builder{};
 
   builder
-    .For(micm::Phase())
-    .Reacting(micm::Species())
-    .Producing(micm::Species())
+    .For(micm::Phase<double>())
+    .With(micm::Species<double>())
```
mattldawson, line 16:
```suggestion
    .Reacting(micm::Species<double>())
```

mattldawson (approved):

PR #31, opened 2023-01-05
28 create solver solverbuilder chapmanodesolver chapmanodesolverbuilder stubs

### LINKED ISSUE #28: Create Solver, SolverBuilder, ChapmanODESolver, ChapmanODESolverBuilder stubs

K20shores:
Closes #28


include/micm/solver/solver.hpp
```diff
@@ -0,0 +1,38 @@
...
+   protected:
+    /// @brief The size of the system
+    std::size_t size_;
+
+   public:
+    /// @brief Default constructor
+    Solver();
+    /// @brief A virtual function to be defined by any solver baseclass
+    /// @param T The current species concentrations of the system
+    virtual void Solve(T) = 0;
```
mattldawson, line 29:
```suggestion
    virtual void Solve(std::vector<T>) = 0;
```

include/micm/solver/chapman_ode_solver.hpp
```diff
@@ -0,0 +1,40 @@
...
+  template<typename T>
+  class ChapmanODESolver : public Solver<T>
+  {
+   private:
+   public:
+    /// @brief Default constructor
+    ChapmanODESolver();
+    /// @brief Move the system to the next state
+    /// @param The collection of species concentrations
+    void Solve(T);
```
mattldawson, line 26:
```suggestion
    void Solve(std::vector<T>);
```

include/micm/solver/solver.hpp
K20shores: If we're careful, it may not matter if it's a `double[]` or a `std::vector<T>`. I'm fine with enforcing one or the other, but I see no reason why we can't support both. Perhaps this is a policy decision?

mattldawson: ah, ok - yeah, I like that better

mattldawson (approved):

PR #35, opened 2023-01-20
32 create cfortran interface to solver and solverbuilder classes

K20shores:
This adds a dummy interface to fortran so that micm can be added to music box early on

Also allows us to add MICM to command-line MusicBox


mattldawson (approved):
looks great!

PR #37, opened 2023-01-27
36 add rate constant calculation to chapmanodesolver w test against micm pp code

### LINKED ISSUE #36: Add rate constant calculation to ChapmanODESolver w/ test against MICM-PP code

K20shores:
Closes #36. Adds some tests as both unit and regression using values from [JPL](https://jpldataeval.jpl.nasa.gov/pdf/JPL_00-03.pdf).


mattldawson (approved):
looks good to me!

PR #53, opened 2023-03-13
52 create rosenbrockodesolver based on chapmanodesolver

K20shores:
This reconfigures the chapman unit tests so that we can test the new rosenbrock solver in the same testing functions. Further, I started to flesh out the system builder (the thing that reads config files to make a system) and started trying to pull in the factory builder from the modern c++ design book.


test/unit/unit_configs/chapman/species.json
```diff
@@ -0,0 +1,1923 @@
+
+
+
+
+
+
+
+<!DOCTYPE html>
```
mattldawson, line 8: this file seems like it got overwritten with html somehow.

mattldawson (changes requested):
looks good! just one test file to take a look at

PR #67, opened 2023-05-18
Draft: public const data members

boulderdaze:


boulderdaze:
I wasn't able to update the following test cases (yellow dots) because the default constructors are not available now and I wasn't sure how to properly input the parameters. 
image

test/unit/process/test_arrhenius_rate_constant.cpp
```diff
@@ -7,40 +7,43 @@ TEST(ArrheniusRateConstant, DefaultConstructor){
   micm::ArrheniusRateConstant arrhenius{};
 }
 
-TEST(ArrheniusRateConstant, CalculateWithSystem){
-  micm::ArrheniusRateConstant zero{};
-  auto k = zero.calculate(micm::System());
-  EXPECT_NEAR(k, 1, 0.01);
+//
+// TODO: jiwon 5/18 - commented out because System doesn't have default constructor anymore
```
K20shores, line 11: @mattldawson initially we had said rate constants would calculate from a system. Should this now really be a state?

mattldawson: yes, I think so - I have been making this change for my issue also

include/micm/process/arrhenius_rate_constant.hpp
```diff
@@ -55,14 +55,12 @@ namespace micm
   };
 
   inline ArrheniusRateConstant::ArrheniusRateConstant()
-      : parameters_()
-  {
-  }
+    : parameters_()
+    {}
```
K20shores, line 59: I typically use clang-format to auto format all of our code. Did you happen to run that?

boulderdaze: No I didn't. I wanted to ask about formatting. Good to know. I will use the same auto format and update the code.

boulderdaze: @K20shores do you have specific config for clang-format or use the default?

mattldawson (approved):

include/micm/process/arrhenius_rate_constant.hpp
K20shores: There's a .clang-format file in the root of the repository. Feel free to make any additions to it that you like. Again, we don't really care about which format we have, so long as we have a format.

test/unit/process/test_arrhenius_rate_constant.cpp
boulderdaze: @K20shores Could you elaborate what changes need to be made?

K20shores: Actually, this can be done in a future task. I'll make a ticket for it.

K20shores (approved):

PR #76, opened 2023-06-01
Cmake fixes

K20shores:


cmake/dependencies.cmake
```diff
@@ -44,11 +46,10 @@ if(PROJECT_IS_TOP_LEVEL)
     set("CMAKE_${lang}_CLANG_TIDY" "")
   endforeach ()
 
-  include(FetchContent)
   FetchContent_Declare(googletest
     GIT_REPOSITORY https://github.com/google/googletest.git
-    GIT_TAG release-1.12.1
-    FIND_PACKAGE_ARGS NAMES GTest
+    GIT_TAG v1.13.0
+    # FIND_PACKAGE_ARGS GTest
```
mattldawson, line 52: is this supposed to be commented out?

mattldawson (approved):

boulderdaze (approved):

cmake/dependencies.cmake
K20shores: Yes, I think it doesn't play well with cmake 3.22.0. Not quite sure, but it's only important for if you have google test installed, and none of do. We can revisit when CISL adds cmake 3.24 to the machines and enforce that version and add this back

PR #77, opened 2023-06-02
Sparse matrix class

mattldawson:
Adds a `SparseMatrix` class. This will be used for the Jacobian and linear solver functions.

The data structure follows the CSR pattern with the underlying matrix data stored in a `std::vector<double>`. The matrix structure is block-diagonal where each block is square and of the same dimensions and sparse structure. For the chemical system, the blocks should correspond to individual grid cells.

The `VectorId()` function allows for lookups of the index in the underlying vector for matrix elements using their block, row, and column indices. This can be saved and used for fast access during solving. Normal array syntax `[block][row][column]` access to matrix data is also available for more convenient (but slower) access.

I wasn't sure what is the best way to handle attempts to access zero elements in the array. I set it up to throw exceptions if this happens, but maybe there is a better solution?

### DIFF
--- a/include/micm/util/sparse_matrix.hpp
+++ b/include/micm/util/sparse_matrix.hpp
@@ -28,6 +28,56 @@ namespace micm
     std::vector<std::size_t> row_start_;  // Index in data_ and row_ids_ of the start of each column in a block
 
     friend class SparseMatrixBuilder<T>;
+    friend class ProxyRow;
+    friend class Proxy;
+
+    class Proxy
+    {
+      SparseMatrix& matrix_;
+      std::size_t block_id_;
+      std::size_t row_id_;
+
+     public:
+      Proxy(SparseMatrix& matrix, std::size_t block_id, std::size_t row_id)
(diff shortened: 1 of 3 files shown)


include/micm/util/sparse_matrix.hpp
```diff
@@ -0,0 +1,205 @@
...
+
+    std::vector<T>& AsVector()
+    {
+      return data_;
+    }
+
+    std::size_t VectorIndex(std::size_t block, std::size_t row, std::size_t column) const
+    {
+      if (row >= row_start_.size() - 1 || column >= row_start_.size() - 1 || block >= number_of_blocks_)
+        throw std::invalid_argument("SparseMatrix element out of range");
```
K20shores, line 104: Normally I would say we shouldn't throw, but in this case it's an irrecoverable error, it seems like. The only other option I can think of is like what I did for the SolverResult, but I'm not really sure that would win us anything. I think this is fine.

boulderdaze (approved):

PR #81, opened 2023-06-07
80 micm cannot compile with nvhpc230

K20shores:
- Adds a dockerfile that compiles with the nvhpc compilers
- Excludes regression tests since musica core cannot compile with nvhpc
- adds a test to build micm with nvhpc


K20shores:
I had added nvhpc tests to github, but the image appears to be too large for github actions to handle

boulderdaze (approved):

mattldawson (approved):

PR #87, opened 2023-06-10
Vectorizable matrix class

mattldawson:
Adds a new matrix class `VectorMatrix` that organizes the matrix data for vectorization.

Includes common matrix policy tests that are applied to `Matrix` and `VectorMatrix`

Specializes the `ProcessSet::AddForcingTerms()` function for `VectorMatrix` using a `Vectorizable` concept with custom loops over the matrix data. I _think_ this is an acceptable way to do this, but maybe there is a better solution.


CMakeLists.txt
```diff
@@ -22,7 +22,7 @@ message(STATUS "CMake build configuration for micm(${CMAKE_BUILD_TYPE}) ${PROJEC
 include(cmake/StaticAnalyzers.cmake)
 
 # C++ options
-set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD 20)
```
K20shores, line 25: If you can, please remove this line. That sets the standard for all translation units and we only really need that on micm. I think I added this, but preferably we would set this only on the micm target in the src/CMakeLists.txt file: `target_compile_features(micm INTERFACE cxx_std_20)`

include/micm/process/process_set.hpp
```diff
@@ -158,9 +171,49 @@ namespace micm
...
+  template<template<class> typename M>
+    requires Vectorizable<M<double>>
+  inline void
+  ProcessSet::AddForcingTerms(const M<double>& rate_constants, const M<double>& state_variables, M<double>& forcing) const
+  {
+    const auto& v_rate_constants = rate_constants.AsVector();
+    const auto& v_state_variables = state_variables.AsVector();
+    auto& v_forcing = forcing.AsVector();
+    // loop over all rows
+    for (std::size_t i_block = 0; i_block < state_variables.NumberOfBlocks(); ++i_block)
```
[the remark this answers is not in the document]
mattldawson: I wasn't sure what terms to use for this new matrix structure, so we can change them if needed.

I used "block" here to refer to the highest level structure in the matrix data (all the columns for 4 or 8 or whatever number of rows is best for vectorization would be one block), so a matrix of 20 grid cells X 40 species would have 5 "blocks" if we choose to group 4 rows for vectorization (5 = 20 / 4).

The cells are the lowest-level structure - a single species for a specific grid cell in the 4 or 8 or whatever number of rows would be one "cell".

We can use whatever names you and John prefer, just let me know what you would like them to be.

include/micm/solver/rosenbrock.hpp
```diff
@@ -70,10 +70,10 @@ namespace micm
...
   };
 
-  /**
-   * @brief An implementation of the Chapman mechnanism solver
-   *
-   */
+   /// @brief An implementation of the Chapman mechnanism solver
+   ///
+   /// The template parameter is the type of matrix to use
+  template<template<class> class M>
```
K20shores, line 76: Would it be possible to replace `M` with `Matrix` everywhere, or is that not possible because of our own `Matrix` class?

K20shores (approved):

include/micm/util/vector_matrix.hpp
```diff
@@ -0,0 +1,202 @@
...
+      Proxy(VectorMatrix &matrix, std::size_t block_index, std::size_t row_index, std::size_t y_dim)
+          : matrix_(matrix),
+            block_index_(block_index),
+            row_index_(row_index),
+            y_dim_(y_dim)
+      {
+      }
+      Proxy &operator=(const std::vector<T> other)
+      {
+        assert(other.size() >= y_dim_ && "Matrix row size mismatch in assignment from vector");
```
K20shores, line 47: You'll need to do something like what I did in #84 to make these tests pass when we run them in release mode

https://github.com/NCAR/micm/pull/84/files#diff-78acddefff72c23089ee070cf969314ef0b55616b44fb5b03439896337a73913R40-R45

CMakeLists.txt
mattldawson: done!

include/micm/solver/rosenbrock.hpp
mattldawson: Yeah, I think it might at least be confusing to use `Matrix`, even if it is technically possible. In almost all the code I've seen that uses template parameters, the parameters are a single capital letter. Normally, I'll look at the [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html) for something like this, but I can't find anywhere where they specifically talk about template parameter naming.

I'm in favor of adopting a style for template parameter naming, but don't have a strong preference for what we choose.  Maybe we can choose a couple options and do pros/cons?

### Single capital letter
**Pros**: Easy to identify template parameters in code (if this is the only thing we allow to be a single capital letter). Consistent with many examples of template code.

**Cons**: Is not descriptive and might become confusing if multiple template parameters start to be used for a single class or function.

include/micm/util/vector_matrix.hpp
mattldawson: done!

include/micm/process/process_set.hpp
```diff
@@ -128,10 +141,10 @@ namespace micm
...
-      const Matrix<double>& rate_constants,
-      const Matrix<double>& state_variables,
-      Matrix<double>& forcing) const
+  template<template<class> typename M>
+    requires(!Vectorizable<M<double>>)
+  inline void
+  ProcessSet::AddForcingTerms(const M<double>& rate_constants, const M<double>& state_variables, M<double>& forcing) const
   {
     // loop over grid cells
     for (std::size_t i_cell = 0; i_cell < state_variables.size(); ++i_cell)
```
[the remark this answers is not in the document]
sjsprecious: Thanks. But the exact dimension order should depend on whether we loop **ncells** as the inner- or outermost place?

include/micm/process/process_set.hpp
```diff
@@ -158,9 +171,49 @@ namespace micm
...
   };
 
+  template<template<class> typename M>
+    requires Vectorizable<M<double>>
+  inline void
+  ProcessSet::AddForcingTerms(const M<double>& rate_constants, const M<double>& state_variables, M<double>& forcing) const
+  {
+    const auto& v_rate_constants = rate_constants.AsVector();
+    const auto& v_state_variables = state_variables.AsVector();
+    auto& v_forcing = forcing.AsVector();
```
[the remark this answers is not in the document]
sjsprecious: Ok.

include/micm/process/process_set.hpp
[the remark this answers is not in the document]
sjsprecious: Oh, I see. Thanks for the clarification. This is what John is looking for but my Fortran example shows that we may only need the innermost `i_cell` loop for the best CPU performance. We are iterating on the performance results and I could show you more details later.

CMakeLists.txt
sjsprecious: Is there a down effect to add this line here? Just try to learn more about CMake.

K20shores: Not necessarily bad to set `CMAKE_CXX_STANDARD`, but cmake is moving more towards things that set options on specific targets rather than doing something system wide.

include/micm/solver/rosenbrock.hpp
K20shores: How would you feel about `MatrixPolicy`? This indicates that the use of a thing is meant to provide different behavior, not just generation of code for a different data type. Behavior changes will have a suffix of `Policy` and data types will be a single letter?

CMakeLists.txt
sjsprecious: Thanks and glad to know it!

include/micm/process/process_set.hpp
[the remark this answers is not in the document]
mattldawson: Ah, good to know! Ok, I will look forward to hearing the results of your tests.

include/micm/process/process_set.hpp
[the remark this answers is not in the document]
mattldawson: Yes, the existing matrix is ordered like this, with grid cells being the slowest varying index.

The new VectorMatrix has the block-like structure John described, but it still implements the same matrix "policy" so that you can use `Matrix` and `VectorMatrix` interchangeably (primarily for testing), but write optimized functions that loop over the underlying vector in a way that is specific to the VectorMatrix ordering.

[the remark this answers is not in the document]
sjsprecious: Thanks for the clarification. That is clear to me now.

sjsprecious:
@mattldawson I checked out your branch on Gust but I could not build it with the NVHPC compiler. It failed with an error I reported to @K20shores before and I thought it should be resolved by his recent PR #84.

include/micm/solver/rosenbrock.hpp
mattldawson: I like that solution. Will make the change now

mattldawson:
> @mattldawson I checked out your branch on Gust but I could not build it with the NVHPC compiler. It failed with an error I reported to @K20shores before and I thought it should be resolved by his recent PR #84.

Are you able to build the main branch after PR #84 was merged?

include/micm/solver/rosenbrock.hpp
sjsprecious: Ok, another new terminology to me. What is `Policy` here?

mattldawson: It's a strategy for treating things by behaviors (like what functions they implement): https://en.wikipedia.org/wiki/Modern_C%2B%2B_Design

For our `AddForcingTerms` function, the `MatrixPolicy` type just has to have the `[]` syntax for 2D arrays (it should probably work if you passed it a `vector<vector<double>>`). Our specialized `AddForcingTerms` function that works with `VectorMatrix` gets used for things that have the additional functions that `VectorMatrix` provides:
```
  /// Concept for vectorizable matrices
  template<typename T>
  concept Vectorizable = requires(T t) {
    t.BlockSize();
    t.NumberOfBlocks();
    t.VectorSize();
  };

...

  template<template<class> typename MatrixPolicy>
    requires Vectorizable<MatrixPolicy<double>>
  void AddForcingTerms(const MatrixPolicy<double>& rate_constants, const MatrixPolicy<double>& state_variables, MatrixPolicy<double>& forcing)
      const;
```
So, the specialized function works with `VectorMatrix` and any other type that has the `BlockSize()`, `NumberOfBlocks()` and `VectorSize()` functions

mattldawson:
> > > @mattldawson I checked out your branch on Gust but I could not build it with the NVHPC compiler. It failed with an error I reported to @K20shores before and I thought it should be resolved by his recent PR #84.
> > 
> > 
> > Are you able to build the main branch after PR #84 was merged?
> 
> @mattldawson , yes, I just checked out the latest `main` branch and could build it on Gust with `NVHPC/23.1`. All the tests pass within 0.01 second.

Hi @sjsprecious - I think this branch should compile and pass the tests on Gust with `nvc++` now. Could you give it a try at some point and let me know?

I bypassed the function that is optimized for `VectorMatrix` when C++20 is not supported. We can decide after your testing if we want to make this available for the GPU code or not. If we do, we could ask CISL about getting the issue with the older g++ compilers fixed on Gust/Derecho

sjsprecious:
> > > > @mattldawson I checked out your branch on Gust but I could not build it with the NVHPC compiler. It failed with an error I reported to @K20shores before and I thought it should be resolved by his recent PR #84.
> > > 
> > > 
> > > Are you able to build the main branch after PR #84 was merged?
> > 
> > 
> > @mattldawson , yes, I just checked out the latest `main` branch and could build it on Gust with `NVHPC/23.1`. All the tests pass within 0.01 second.
> 
> Hi @sjsprecious - I think this branch should compile and pass the tests on Gust with `nvc++` now. Could you give it a try at some point and let me know?
> 
> I bypassed the function that is optimized for `VectorMatrix` when C++20 is not supported. We can decide after your testing if we want to make this available for the GPU code or not. If we do, we could ask CISL about getting the issue with the older g++ compilers fixed on Gust/Derecho

Hi @mattldawson , thanks for your update and efforts. I could confirm that I am able to build MICM on Gust with your `develop-vector-matrix` branch and `nvhpc/23.1` module now.

Which version of GCC do you need for C++20 support? Gust has `gcc/12.2.0` installed.

sjsprecious (approved):
Thanks @mattldawson for working on this new class. Hope we could restructure the loop easily with these changes later.

PR #89, opened 2023-06-13
Json parsing

boulderdaze:
Completed:
- Created solver config class to read and parse JSON objects to create solver parameters 
- Created unit tests 
- Auto-formatted
- Deleted files that are not used anymore
- Updated the integration test for `chapman` using configure files

Next step:
- Update the way to read configure files so it doesn't need to search species file first to read


include/micm/configure/solver_config.hpp
```diff
@@ -0,0 +1,495 @@
...
+#include <micm/process/arrhenius_rate_constant.hpp>
+#include <micm/process/photolysis_rate_constant.hpp>
+#include <micm/process/process.hpp>
+#include <micm/system/phase.hpp>
+#include <micm/system/property.hpp>
+#include <micm/system/species.hpp>
+#include <micm/system/system.hpp>
+#include <variant>
+
+#ifdef USE_JSON
```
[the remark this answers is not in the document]
boulderdaze: Yep, I added your suggestion to the code.

include/micm/process/photolysis_rate_constant.hpp
```diff
@@ -60,13 +61,14 @@ namespace micm
...
-      const
+  inline double PhotolysisRateConstant::calculate(
+      const Conditions& conditions,
+      const std::vector<double>::const_iterator& custom_parameters) const
   {
     return (double)*custom_parameters;
   }
 
-  inline std::size_t PhotolysisRateConstant::SizeCustomParameters() const
+  inline std::size_t PhotolysisRateConstant::SizeCustomParameters()
```
mattldawson, line 71: why remove `const` here?

test/unit/configure/test_solver_config.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+  short idx = 0;
+  for (const auto& p : solver_params.processes_)
+  {
+    EXPECT_EQ(p.reactants_.size(), num_reactants_in_each_process[idx]);
+    EXPECT_EQ(p.products_.size(), num_products_in_each_process[idx]);
+    EXPECT_EQ(p.products_[0].second, yield_value_of_first_product_in_each_process[idx]);
+    EXPECT_EQ(p.phase_.species_.size(), num_phase_in_each_process);
+    idx++;
+  }
```
mattldawson, line 78: would it be possible to include a check of the parameters for the Arrhenius rate constants and the name for the photolysis rate constants?

mattldawson (approved):
looks great! just a couple minor comments

We'll have to add a parser for Troe rate constants at some point, but the simple Chapman mechanism we're starting with doesn't have these, so I'll make a separate issue for this.

PR #92, opened 2023-06-15
creating a separate cuda and openacc target

K20shores:
Creates a separate target for cuda and openacc. This means that tests that need to use code for either must explicitly link to that library

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -57,11 +57,9 @@ if(ENABLE_OPENACC)
     PRIVATE micm
   )
 
-  if(ENABLE_OPENACC)
-    target_link_libraries(micm_openacc 
-      PUBLIC OpenACC::OpenACC_CXX
-    )
-  endif()
+  target_link_libraries(micm_openacc 
+    PUBLIC OpenACC::OpenACC_CXX
+  )
 endif()
 
 if(ENABLE_CUDA)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -28,9 +28,9 @@ if(ENABLE_MPI)
   target_link_libraries(micm INTERFACE MPI::MPI_CXX)
 endif()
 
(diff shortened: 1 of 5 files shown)


boulderdaze (approved):

mattldawson (approved):

src/CMakeLists.txt
```diff
@@ -51,23 +51,32 @@ if(ENABLE_CUDA OR ENABLE_OPENACC)
...
     PRIVATE micm
   )
 
-  if(ENABLE_CUDA)
-    target_link_libraries(micm_gpu 
-      PUBLIC cudart
-    )
-  endif()
-
   if(ENABLE_OPENACC)
```
sjsprecious, line 60: Is this `if(ENABLE_OPENACC)` redundant since it is already in a `if(ENABLE_OPENACC)` statemenet?

K20shores: Yes, fixed.

src/CMakeLists.txt
```diff
@@ -51,23 +51,32 @@ if(ENABLE_CUDA OR ENABLE_OPENACC)
   set(GPU_FLAGS -gpu=${GPU_ARCH},lineinfo -Minfo=accel)
   message(STATUS "GPU Flags: ${OpenACC_CXX_FLAGS} ${GPU_FLAGS}")
 
-  # target_compile_options(micm_gpu PRIVATE ${OpenACC_CXX_FLAGS} ${GPU_FLAGS})
+  target_compile_options(micm_openacc PRIVATE ${OpenACC_CXX_FLAGS} ${GPU_FLAGS})
```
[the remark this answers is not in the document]
K20shores: In this case, it's all about linking. PRIVATE here means that, even though we link `micm_openacc` to `micm`, these flags will only be added to the targets of `micm_openacc`. If we were to use `PUBLIC`, then all of the translation units needed by `micm` would add the GPU flags (to my understanding), which is not what we want.

PR #104, opened 2023-07-05
Draft:Map params

boulderdaze:
Completed

- Updated reading configure files by getting the path to the directory instead of the filename
- Added a mapping function for concentrations and corresponding species from user input
- Added a mapping function for photolysis rate constants  and its user input of them
- Updated the unit and integration tests for the mapping functions
- Updated the copy constructors for `Phase` and `Property`
- Updated variable length array to dynamic array in `process_set` because other compilers except for GCC don't support variable length array.
- Optimized reading and parsing configure files by reducing copy operations
- Update reading configure files focusing on configuring the given data for extensibility instead of just creating `SolverParameters`
- Moved `Conditions` class to its own header file (`conditions.hpp`) because `#include` header files induces circular dependency between `state.hpp` and `rate_constant.hpp`

To Discuss

- [ ] `Conditions` and `Condition` class

- [ ] heap allocation and de-allocation for every iteration in `process_set` 

Still to do

- [ ] Update unit tests ; Currently almost all unit tests are not updated because I wanted to discuss whether those changes make more sense and wanted to receive feedback before updating them because they are lots of work!


include/micm/process/process_set.hpp
```diff
@@ -218,6 +219,7 @@ namespace micm
         prod_id += number_of_products_[i_rxn];
         yield += number_of_products_[i_rxn];
       }
+      delete[] rate;
```
K20shores, line 222: On second thought, please just make rate a vector<double> and reserve `L` spaces. There's no reason to manage the memory ourselves.

K20shores (changes requested):

include/micm/process/process_set.hpp
boulderdaze: Sounds good. I updated `rate` to a vector

include/micm/solver/state.hpp
```diff
@@ -78,4 +89,48 @@ namespace micm
     for (auto& name : parameters.state_variable_names_)
       variable_map_[name] = index++;
   }
+
+  template<template<class> class MatrixPolicy>
+  inline void State<MatrixPolicy>::SetConcentrations(
+      const System& system,
+      const std::unordered_map<std::string, double>& species_to_concentration)
```
mattldawson, line 96: The species concentrations will be for a set of grid cells, so might be something like:
```suggestion
      const std::unordered_map<std::string, std::vector<double>>& species_to_concentration)
```
We'll want to try to come up with an efficient-as-possible solution eventually, but for now I think it's fine to just get something that works and can be optimized later.

include/micm/solver/state.hpp
```diff
@@ -78,4 +89,48 @@ namespace micm
...
+    {
+      auto species_ptr = species_to_concentration.find(species.name_);
+      if (species_ptr == species_to_concentration.end())
+      {
+        throw std::invalid_argument("Concentration value for '" + species.name_ + "' must be given.");
+      }
+      concentrations.push_back(species_ptr->second);
+    }
+
+    variables_[0] = concentrations;
```
mattldawson, line 111: This is just setting the values for the first grid cell.

include/micm/solver/state.hpp
```diff
@@ -78,4 +89,48 @@ namespace micm
...
+  }
+
+  template<template<class> class MatrixPolicy>
+  inline void State<MatrixPolicy>::SetPhotolysisRate(
+      const std::vector<PhotolysisRateConstant>& photolysis_rate_arr,
+      const std::unordered_map<std::string, double>& photolysis_rate)
```
mattldawson, line 117: same comment as for `SetConcentrations()`

include/micm/system/conditions.hpp
```diff
@@ -0,0 +1,16 @@
...
+#pragma once
+
+namespace micm
+{
+  /// @brief Environemental conditions
+  struct Conditions
+  {
+    double temperature_{ 0.0 };
+    double pressure_{ 0.0 };
+    double air_density_{ 1.0 };
```
mattldawson, line 14:
```suggestion
    double temperature_{ 0.0 }; // K
    double pressure_{ 0.0 }; // Pa;
    double air_density_{ 1.0 }; // mol m-3
```

mattldawson (changes requested):
Looks great to me! I just have one recommendation for setting concentrations and rate constants.

include/micm/solver/state.hpp
boulderdaze: Okay that makes sense. I addressed this in the latest commit.

test/unit/configure/test_solver_config.cpp
```diff
@@ -127,4 +124,40 @@ TEST(SolverConfig, ReadAndParseProcess)
...
+
+TEST(SolverConfig, GettingPhotolysisRateConstantThrowsExceptionWithFailedParsing)
+{
+  micm::SolverConfig<micm::JsonReaderPolicy, micm::NoThrowPolicy> solverConfig;
+  bool is_parse_success = solverConfig.ReadAndParse("not_a_config_file_directory");
+  EXPECT_FALSE(is_parse_success);
+  // EXPECT_ANY_THROW(solverConfig.GetPhotolysisRateConstants());
+}
+
+TEST(SolverConfig, GetPhotolysisRateConstants)
```
K20shores, line 144: Can you add a test that creates a solver using the solverConfig?

K20shores (changes requested):

K20shores (approved):

PR #135, opened 2023-07-20
Example in README

### LINKED ISSUE #96: Automate documentation builds
Automate documentation deployment to github.io using GH Actions. (Can follow the approach used for [TUV-x](https://github.com/NCAR/tuv-x)) and include a link in the README

### LINKED ISSUE #123: Create simple hello-world example in README
Create a small example in the README that a new user can follow to get started with MICM.

Make any changes needed to simplify the user code (Maybe move call to `UpdateState()` inside `Solve()`

mattldawson:
Adds a short example to the README. Also adds a documentation-building GH action and moves the GPU instructions to the documentation from the README. (I think most people just getting started with MICM will use the CPU version, and then can look up in the documentation how to use it with GPUs if needed)

closes #123
closes #96


include/micm/solver/rosenbrock.hpp
```diff
@@ -131,7 +132,7 @@ namespace micm
     /// @brief Builds a Rosenbrock solver for the given system, processes, and solver parameters
     /// @param system The chemical system to create the solver for
     /// @param processes The collection of chemical processes that will be applied during solving
-    RosenbrockSolver(const System& system, std::vector<Process>&& processes, const RosenbrockSolverParameters parameters);
+    RosenbrockSolver(const System& system, const std::vector<Process>& processes, const RosenbrockSolverParameters parameters);
 
     virtual ~RosenbrockSolver();
```
boulderdaze, line 137: Do you think we can pass `RosenbrockSolverParameters` as const ref to avoid copy?

include/micm/solver/state.hpp
```diff
@@ -51,6 +51,12 @@ namespace micm
         const System& system,
         const std::unordered_map<std::string, std::vector<double>>& species_to_concentration);
 
+    /// @brief Set a single species concentration
+    /// @param species the species to set the concentration for
+    /// @param concentration concentration(s) [mol m-3]
+    void SetConcentration(const Species& species, double concentration);
+    void SetConcentration(const Species& species, std::vector<double> concentration);
+
```
boulderdaze, line 59: Do you think we can pass `concentration` as const ref to avoid copy?

boulderdaze (approved):

test/tutorial/test_README_example.cpp
```diff
@@ -0,0 +1,54 @@
+#include <iomanip>
```
K20shores, line 1: If we wanted, we could have cmake write the content for this test and the readme. If we ever changed the contents of that cmake file, the tests and the readme would be updated together. That's probably more work than is necessary though

K20shores (approved):

include/micm/solver/rosenbrock.hpp
mattldawson: done!

include/micm/solver/state.hpp
mattldawson: done!

test/tutorial/test_README_example.cpp
mattldawson: maybe revisit this when we start writing tutorials in the documentation?

K20shores: fine with me

PR #139, opened 2023-07-24
Address compiler warnings

### LINKED ISSUE #138: Cleanup all compiler warnings, where reasonable
Check the build logs for any of the successful github actions and address compiler warnings

mattldawson:
I think there is only one remaining warning. It's with nvhpc, and I'm pretty sure it's incorrectly applied:
```
"/micm/test/regression/RosenbrockChapman/regression_test_solve.cpp", line 57: warning: ignoring return value type with "nodiscard" attribute [nodiscard_return_type]
      fixed_results[i] = fixed_solver.Solve(0.0, 500.0, fixed_state);
                       ^

Remark: individual warnings can be suppressed with "--diag_suppress <warning-name>"
```

I opted to leave our code unchanged, instead of trying to get rid of the warning.

closes #138

### DIFF
--- a/include/micm/configure/solver_config.hpp
+++ b/include/micm/configure/solver_config.hpp
@@ -39,7 +39,7 @@ namespace micm
       {
       }
 
-      virtual const char* what()
+      virtual const char* what() const noexcept override
       {
         return msg_;
       }
--- a/include/micm/configure/solver_config.hpp
+++ b/include/micm/configure/solver_config.hpp
@@ -39,7 +39,7 @@ namespace micm
       {
       }
 
-      virtual const char* what() const noexcept override
+      virtual const char* what() const noexcept
       {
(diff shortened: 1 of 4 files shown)


include/micm/configure/solver_config.hpp
```diff
@@ -39,7 +39,7 @@ namespace micm
       {
       }
 
-      virtual const char* what()
+      virtual const char* what() const noexcept override
       {
         return msg_;
```
boulderdaze, line 44: I think `override` specifier is for a function that overrides the virtual version of it, so I think we can remove it

mattldawson: done!

K20shores (approved):

boulderdaze (approved):

PR #173, opened 2023-08-03
KPP to MICM

dwfncar:
PR for initial implementation.
Need still to test MICM with the generated JSON from the KPP Chapman example ...

Description:
    kpp_to_micm.py translates KPP config files to MICM JSON config files
    (desginated by the suffixes .kpp, .spc, .eqn, .def)
    from a single directory specified by the --kpp_dir argument.

    In the initial implementation,
    the KPP sections #ATOMS (not yet used), #DEFFIX (not yet used),
    #DEFVAR, and #EQUATIONS are read and parsed.
    Equations with the hv reactant are MICM PHOTOLYSIS reactions,
    all others are assumed to be ARRHENIUS reactions.

TODO:
    (1) Parse both A and B Arrhenius coefficients from KPP equations
    Currently a single rate coeffient is assigned to A and B is set to 0.
    (2) Translate stoichiometric coefficients in the equation string
    with more than one digit.
    (3) Add method unit tests with pytest.
    (4) Add support for many more reaction types ...


mattldawson (approved):
Looks good! I like the idea of adding some unit tests - this text parsing is always tricky because of all the edge cases

boulderdaze (approved):

K20shores (approved):

PR #174, opened 2023-08-05
Add LLVM support for JIT compiling

### LINKED ISSUE #172: Add LLVM libraries to build and create JIT class for generating functions

mattldawson:
Adds an option to include LLVM in the build with `JitCompiler` and `JitFunction` convenience classes for JIT compiling functions. Hopefully the tests can be used as examples for working with these classes, but we can add more documentation if it would be helpful.

closes #172 

(Also, I removed the Intel build from the GH Actions because it was running out of space)


boulderdaze (approved):

include/micm/jit/jit_function.hpp
```diff
@@ -0,0 +1,307 @@
...
+// SPDX-License-Identifier: Apache-2.0
+#pragma once
+
+#include <stdlib.h>
+
+#include <iostream>
+#include <memory>
+
+#include "jit_compiler.hpp"
+#include "llvm//IR/BasicBlock.h"
```
K20shores, line 12:
```suggestion
#include "llvm/IR/BasicBlock.h"
```

include/micm/jit/jit_function.hpp
```diff
@@ -0,0 +1,307 @@
...
+      arg.alloca_ = CreateEntryBlockAlloca(arg.type_, arg.name_);
+    for (auto& arg : arguments_)
+      builder_->CreateStore(arg.arg_, arg.alloca_);
+    for (auto& arg : arguments_)
+      arg.ptr_ = builder_->CreateLoad(arg.type_, arg.alloca_);
+  }
+
+  std::pair<llvm::orc::ResourceTrackerSP, llvm::JITTargetAddress> JitFunction::Generate()
+  {
+    std::pair<llvm::orc::ResourceTrackerSP, llvm::JITTargetAddress> ret_val;
```
K20shores, line 202: can we add an assert or some other check to ensure this is only called once?

cmake/test_util.cmake
```diff
@@ -30,6 +30,11 @@ function(create_standard_test)
     target_compile_definitions(test_${TEST_NAME} PUBLIC USE_JSON)
   endif()
 
+  if(ENABLE_LLVM)
+    target_link_libraries(test_${TEST_NAME} PRIVATE ${llvm_libs})
+    target_compile_definitions(test_${TEST_NAME} PUBLIC USE_LLVM)
+  endif()
+
```
K20shores, line 37:
```suggestion
  endif()

  if(ENABLE_LLVM)
    target_link_libraries(test_${TEST_NAME} PRIVATE ${llvm_libs})
  endif()

```

Do we even need these compile definitions?

K20shores (changes requested):

cmake/test_util.cmake
mattldawson: good point, I don't think we do.

include/micm/jit/jit_function.hpp
mattldawson: done!

K20shores (approved):

PR #180, opened 2023-08-11
158 add integration tests for simple chemistry systems

### LINKED ISSUE #158: Add integration tests for simple chemistry systems
Create a set of integration tests for each of the reaction types that uses a simple system (e.g., A -> B -> C) that can be solved analytically and compared to the solver output.

Similar tests exist for CAMP and can be used as a template (e.g., https://github.com/open-atmos/camp/blob/main/test/unit_rxn_data/test_rxn_troe.F90)

K20shores:
closes #158 

Checks our numerical solutions to analytical solutions for all of the reaction rates, except the surface reaction which I made into an issue for #182.
There's also a stiff system check.


test/integration/analytical.cpp
```diff
@@ -0,0 +1,1539 @@
...
+  size_t _c = map.at("C");
+
+  for (size_t i = 0; i < model_concentrations.size(); ++i)
+  {
+    EXPECT_NEAR(model_concentrations[i][_a], analytical_concentrations[i][0], 1e-3)
+        << "Arrays differ at index (" << i << ", " << 0 << ")";
+    EXPECT_NEAR(model_concentrations[i][_b], analytical_concentrations[i][1], 1e-3)
+        << "Arrays differ at index (" << i << ", " << 1 << ")";
+    EXPECT_NEAR(model_concentrations[i][_c], analytical_concentrations[i][2], 1e-3)
+        << "Arrays differ at index (" << i << ", " << 2 << ")";
```
K20shores, line 1380: @mattldawson for some reason I had to drop the tolerance for this reaction to 1e-3. I suspect it's another floating point thing, but I'm not sure if it indicates something else that may be wrong

boulderdaze (approved):

mattldawson (approved):
nice!

PR #188, opened 2023-08-15
Add jacobian terms cuda

qinatan:
Parallelize grid cells working with vectorized(flipped) memory layout

I changed the function return type from void to double in AddForcingTerms and AddJacobianTerms in CUDAProcessSet class in order to measure time performance of kernel function.


include/micm/process/process_set.hpp
```diff
@@ -21,7 +21,9 @@ namespace micm
     std::vector<std::size_t> number_of_products_;
     std::vector<std::size_t> product_ids_;
     std::vector<double> yields_;
-    std::vector<std::size_t> jacobian_flat_ids_;
+    //std::vector<std::size_t> jacobian_flat_ids_;
```
K20shores, line 24: Please either make it private and add a getter or just delete this commented out line if it needs to be public

K20shores (changes requested):

include/micm/process/cuda_process_set.cuh
```diff
@@ -21,5 +21,22 @@ namespace micm
         int product_ids_size,
         const double* yields,
         int yields_size);
+
+    void AddJacobianTerms_kernelSetup(
+        const double* rate_constants, 
+        const double* state_variables, 
+        size_t n_grids, 
+        size_t n_reactions, 
+        size_t n_species,
```
sjsprecious, line 30: In the `AddForcingTerms_kernelSetup` function, different variable names and types are used. Either way works for me but I would suggest using the consistent declaration.

include/micm/process/process_set.hpp
sjsprecious, line 24: remove this line

include/micm/process/process_set.hpp
```diff
@@ -304,15 +303,15 @@ namespace micm
             d_rate_d_ind[i_cell] = v_rate_constants[offset_rc + i_rxn * L + i_cell];
           for (std::size_t i_react = 0; i_react < number_of_reactants_[i_rxn]; ++i_react)
           {
-            if (i_react == i_ind)
+             if (i_react == i_ind)
```
sjsprecious, line 306: Is this space intended?

src/process/process_set.cu
```diff
@@ -7,7 +7,7 @@ namespace micm
     // flipped memory layout
     __global__ void AddForcingTerms_kernel(
         double* rate_constants,
-        double* state_variables,
+        double* state_variables,
```
sjsprecious, line 10: remove the space?

test/unit/process/test_cuda_process_set.cpp
```diff
@@ -99,17 +174,17 @@ using Group100000VectorMatrix = micm::VectorMatrix<T, 100000>;
...
-  testRandomSystem<Group100000VectorMatrix>(100000, 500, 400);
-  testRandomSystem<Group1000000VectorMatrix>(1000000, 500, 400);
-
-  std::cout << "system with 100 reactions and 80 species" << std::endl;
-  testRandomSystem<Group1000VectorMatrix>(1000, 100, 80);
-  testRandomSystem<Group10000VectorMatrix>(10000, 100, 80);
-  testRandomSystem<Group100000VectorMatrix>(100000, 100, 80);
-  testRandomSystem<Group1000000VectorMatrix>(1000000, 100, 80);
+   testRandomSystem_AddForcingTerms<Group10000VectorMatrix>(10000, 500, 400);
+   testRandomSystem_AddJacobianTerms<Group10000VectorMatrix, Group10000SparseVectorMatrix>(10000, 500, 400);
```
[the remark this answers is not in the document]
sjsprecious: Also, shall we move this test set to the separate repo where we perform verification and performance evaluation? Or we only do performance evaluation in that repo?

sjsprecious (changes requested):
Thanks @qinatan for working on this PR. I have some minor comments and once you address them, I would test this PR by myself to double check the results.
Also, would you issue a separate PR in the performance repository to show how to measure the performance of jacobian function between CPU and GPU?

include/micm/process/process_set.hpp
qinatan: Thanks for catching this. I will make it to be private.

include/micm/process/cuda_process_set.cuh
qinatan: I updated the variable names and data type the same in both AddJacobianTerms and AddForcingTerms for consistency.

include/micm/process/process_set.hpp
qinatan: I made std::vector<std::size_t> jacobian_flat_ids__ as protected member data, and removed the line.

include/micm/process/process_set.hpp
qinatan: No. Just the format difference. I removed the space.

include/micm/process/process_set.hpp
```diff
@@ -304,15 +303,15 @@ namespace micm
...
               continue;
             for (std::size_t i_cell = 0; i_cell < L; ++i_cell)
               d_rate_d_ind[i_cell] *= v_state_variables[offset_state + react_id[i_react] * L + i_cell];
           }
           for (std::size_t i_dep = 0; i_dep < number_of_reactants_[i_rxn]; ++i_dep)
           {
             for (std::size_t i_cell = 0; i_cell < L; ++i_cell)
-              v_jacobian[offset_jacobian + *flat_id + i_cell] -= d_rate_d_ind[i_cell];
+              v_jacobian[offset_jacobian + *flat_id + i_cell] -= d_rate_d_ind[i_cell];
```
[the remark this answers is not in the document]
qinatan: done.

src/process/process_set.cu
```diff
@@ -48,6 +48,126 @@ namespace micm
       }    // if check for valid CUDA threads
     }      // end of AddForcingTerms_kernel
 
+
+  __global__ void AddJacobianTerms_kernel(
```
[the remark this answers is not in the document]
qinatan: On the CPU version, AddJacobianTerms and AddForcingTerms are in the same class called ProcessSet. I think it should be in the same class. But I open to hear for their advice. @K20shores @mattldawson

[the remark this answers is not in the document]
sjsprecious: Gotcha. Somehow I got the impression that each function would have its own file but probably it was wrong. Sorry for the confusion.

test/unit/process/test_cuda_process_set.cpp
```diff
@@ -99,17 +174,17 @@ using Group100000VectorMatrix = micm::VectorMatrix<T, 100000>;
 template<class T>
 using Group1000000VectorMatrix = micm::VectorMatrix<T, 1000000>;
 
+template<class T>
+using Group1000SparseVectorMatrix = micm::SparseMatrix<T, micm::SparseMatrixVectorOrdering<1000>>;
```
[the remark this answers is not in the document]
qinatan: AddJacobianTerms function works with Jacobian, State_Variable and Forcing. Jacobian is a member of class SparseMatrix; State_Variables and Forcing are members of class Matrix. So AddJacobianTerms work with both sparse matrix and regular dense matrix.

test/unit/process/test_cuda_process_set.cpp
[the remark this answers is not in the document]
qinatan: I am not sure. I only have accuracy test for now.

src/process/process_set.cu
qinatan: done

test/unit/process/test_cuda_process_set.cpp
```diff
@@ -90,6 +92,79 @@ void testRandomSystem(std::size_t n_cells, std::size_t n_reactions, std::size_t
...
+    elem = get_double();
+
+  auto non_zero_elements = cpu_set.NonZeroJacobianElements();
+  auto builder = SparseMatrixPolicy<double>::create(n_species).number_of_blocks(n_cells).initial_value(100.0);
+  for (auto& elem : non_zero_elements)
+    builder = builder.with_element(elem.first, elem.second);
+  SparseMatrixPolicy<double> cpu_jacobian{ builder };
+  SparseMatrixPolicy<double> gpu_jacobian{builder};
+
+  cpu_set.SetJacobianFlatIds(cpu_jacobian);
```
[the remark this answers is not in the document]
sjsprecious: When you say first sub-block, do you mean the matrix block for the first grid cell? @mattldawson can you confirm this?

test/unit/process/test_cuda_process_set.cpp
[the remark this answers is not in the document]
sjsprecious: @mattldawson @K20shores can you clarify on this?

test/unit/process/test_cuda_process_set.cpp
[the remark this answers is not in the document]
qinatan: Jacobian is an instance of SparseMatrix class. AddJacobianTerms function work for both SparseMatrix and Matrix class. I just want to make sure we using names accurately and communicating on the same page.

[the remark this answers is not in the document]
qinatan: It has to be an instance of SparseMatrix class because they are different data structure in high level and in the underline implementation.

test/unit/process/test_cuda_process_set.cpp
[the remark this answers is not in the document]
qinatan: I tried but not sure how to make "2 tests shown" .  I test 2 functions separately and both functions passed. I would like to hear from @mattldawson if there's a way to make "2 test shown".

K20shores (approved):

qinatan:
I changed the function return type from void to double in AddForcingTerms and AddJacobianTerms in CUDAProcessSet class in order to measure time performance of kernel function.

include/micm/process/cuda_process_set.cuh
```diff
@@ -4,22 +4,37 @@ namespace micm
...
         const size_t* number_of_products,
-        int number_of_products_size,
         const size_t* product_ids,
-        int product_ids_size,
+        size_t product_ids_size,
         const double* yields,
-        int yields_size);
+        size_t yields_size);
+
+    double AddJacobianTerms_kernelSetup(
```
K20shores, line 23:
```suggestion
    std::chrono::nanoseconds AddJacobianTerms_kernelSetup(
```

Please return a `std::chrono` object, like I suggested above, so that we don't have to keep track of the actual units of time. We can always cast from nanoseconds to microseconds in the driver code and c++ will properly convert for us without us having to really care

K20shores (changes requested):

include/micm/process/cuda_process_set.cuh
qinatan: done.

include/micm/process/cuda_process_set.cuh
```diff
@@ -4,22 +4,37 @@ namespace micm
 {
   namespace cuda
   {
-    void AddForcingTerms_kernelSetup(
+   std::chrono::nanoseconds AddForcingTerms_kernelSetup(
```
mattldawson, line 7: We've been trying to use CamelCase for function names (e.g., `AddForcingTermsKernelSetup`) here and below.

Also, I'm not a CUDA programmer but it was a little confusing to me that this function is called "setup" but also calls the kernel function. If this is a standard naming convention, I'm fine with that, but if not maybe consider something like`AddForcingTermsKernelDriver`. @sjsprecious - what do you think?

include/micm/process/cuda_process_set.hpp
```diff
@@ -26,10 +25,17 @@ namespace micm
 #ifdef USE_CUDA
```
mattldawson, line 25: why not wrap the entire class in the `USE_CUDA` block? I'm just wondering about the case where someone doesn't compile in CUDA support but then is able to create and use a `CudaProcessSet`.

test/unit/process/test_cuda_process_set.cpp
```diff
@@ -99,17 +174,21 @@ using Group100000VectorMatrix = micm::VectorMatrix<T, 100000>;
...
 
-TEST(RandomProcessSet, Matrix)
+template<class T>
+using Group1000SparseVectorMatrix = micm::SparseMatrix<T, micm::SparseMatrixVectorOrdering<1000>>;
+template<class T>
+using Group10000SparseVectorMatrix = micm::SparseMatrix<T, micm::SparseMatrixVectorOrdering<10000>>;
+template<class T>
+using Group100000SparseVectorMatrix = micm::SparseMatrix<T, micm::SparseMatrixVectorOrdering<100000>>;
+template<class T>
+using Group1000000SparseVectorMatrix = micm::SparseMatrix<T, micm::SparseMatrixVectorOrdering<1000000>>;
```
mattldawson, line 184: If some of these aren't being used, I think you could remove them.

mattldawson (changes requested):
Looks great! Just a couple minor things.

Please add the copyright and license text to the top of all the header and source code files (I think they're missing from some of the `.cu` and `.cuh` files):
```
// Copyright (C) 2023 National Center for Atmospheric Research,
//
// SPDX-License-Identifier: Apache-2.0
```

The number of arguments for some of these functions is getting pretty large, and future functions will probably require even more. I think before moving on to the next function, we should explore what options are available to avoid passing a large number of primitive type arguments that are easy to mix up.

Also, can we add the `.cu` and `.cuh` files to those that clang-format reformats?

qinatan:
> Looks great! Just a couple minor things.
> 
> Please add the copyright and license text to the top of all the header and source code files (I think they're missing from some of the `.cu` and `.cuh` files):
> 
> ```
> // Copyright (C) 2023 National Center for Atmospheric Research,
> //
> // SPDX-License-Identifier: Apache-2.0
> ```
> 
> The number of arguments for some of these functions is getting pretty large, and future functions will probably require even more. I think before moving on to the next function, we should explore what options are available to avoid passing a large number of primitive type arguments that are easy to mix up.
> 
> Also, can we add the `.cu` and `.cuh` files to those that clang-format reformats?

I need help on adding .cu and .cuh file to those that clang-format reformats.

qinatan:
> > Looks great! Just a couple minor things.
> > Please add the copyright and license text to the top of all the header and source code files (I think they're missing from some of the `.cu` and `.cuh` files):
> > ```
> > // Copyright (C) 2023 National Center for Atmospheric Research,
> > //
> > // SPDX-License-Identifier: Apache-2.0
> > ```
> > done! 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The number of arguments for some of these functions is getting pretty large, and future functions will probably require even more. I think before moving on to the next function, we should explore what options are available to avoid passing a large number of primitive type arguments that are easy to mix up.
> > Also, can we add the `.cu` and `.cuh` files to those that clang-format reformats?
> 
> I need help on adding .cu and .cuh file to those that clang-format reformats.

test/unit/process/test_cuda_process_set.cpp
qinatan: removed.

include/micm/process/cuda_process_set.hpp
qinatan: done!

include/micm/process/cuda_process_set.cuh
qinatan: changed it to AddForcingTermsKernelDriver for now. Let me know if Jian has any suggestion. @sjsprecious

mattldawson (approved):
very nice work!

(we can add the clang-format as a separate issue)

include/micm/process/cuda_process_set.cuh
sjsprecious: Thanks @mattldawson and @qinatan for pointing this out and fixing it. This is a good point as conventionally, only the code that actually runs on the GPU is called `CUDA kernel`. Other subroutines that prepare the input and output data are really `driver` subroutines as @mattldawson suggested. Changing the name makes it more clear now.

sjsprecious:
> Looks great! Just a couple minor things.
> 
> Please add the copyright and license text to the top of all the header and source code files (I think they're missing from some of the `.cu` and `.cuh` files):
> 
> ```
> // Copyright (C) 2023 National Center for Atmospheric Research,
> //
> // SPDX-License-Identifier: Apache-2.0
> ```
> 
> The number of arguments for some of these functions is getting pretty large, and future functions will probably require even more. I think before moving on to the next function, we should explore what options are available to avoid passing a large number of primitive type arguments that are easy to mix up.
> 
> Also, can we add the `.cu` and `.cuh` files to those that clang-format reformats?

After comparing CPU to GPU code, it seems they share the same vector/matrix input, but GPU code just requires more scalar variable input. Can we group them into a new class called `CudaScalarParam` and add those scalar variables as members? In this way, we just pass the instance of `CudaScalarParam` as an argument to the driver function and we could increase the length of `CudaScalarParam` class as needed, without changing the argument length. What do you think?

K20shores (approved):

Issue #129, opened 2023-07-19
Review the naming across micm and ensure the style of naming of member variables is consistent


mattldawson:
[Use Google C++ style](https://google.github.io/styleguide/cppguide.html)
Add style to README
Create GH Action to report bad styling
Use `typename` instead of `class` in template arguments, [if possible](https://stackoverflow.com/questions/2023977/what-is-the-difference-between-typename-and-class-template-parameters/2024173#2024173)

- [x] #480
- [x] #494
- [x] #473
- [x] #512
- [x] Update micm code in MUSICA and MUSICA-Performance to reflect the changes in the names and the style 
-> I created an issue to each repository. 

**Ideas**
- Modify the `.clang.format` file to apply the [changes to include order](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#includecategories)
- Or modify the `.clang.tidy` file to apply the [changes to include order](https://clang.llvm.org/extra/clang-tidy/checks/llvm/include-order.html)
- See if clang-tidy can handle the `typename` vs. `class` usage 
- Update `.clang.tidy` to flag class, function, [data member naming](https://clang.llvm.org/extra/clang-tidy/checks/readability/identifier-naming.html)

PR #197, opened 2023-08-23
Parameter fixes

K20shores:
Fixed a few logic issues in the solver

- calculating the jacobian exactly once for each phase of the solver
- calculating an initial forcing which I had removed previously, not realizing that when rejecting an `H`, the initial forcing is used
- setting some of the same parameters as KPP
- setting the right size for `Yerror`
- added a `NaN` check

### DIFF
--- a/include/micm/solver/rosenbrock.hpp
+++ b/include/micm/solver/rosenbrock.hpp
@@ -588,7 +588,7 @@ namespace micm
    protected:
     /// @brief Computes the scaled norm of the vector errors
     /// @param Y the original vector
-    /// @param new_number_densities the new vector
+    /// @param Ynew the new vector
     /// @param errors The computed errors
     /// @return
     double NormalizedError(
--- a/include/micm/solver/rosenbrock.hpp
+++ b/include/micm/solver/rosenbrock.hpp
@@ -76,7 +76,7 @@ namespace micm
     // Gamma_i = \sum_j  gamma_{i,j}
     std::array<double, 6> gamma_{};
 
-    double absolute_tolerance_{ 1e-12 };
+    double absolute_tolerance_{ 1e-3 };
     double relative_tolerance_{ 1e-4 };
 
     size_t number_of_grid_cells_{ 1 };  // Number of grid cells to solve simultaneously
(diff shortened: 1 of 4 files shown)


include/micm/solver/rosenbrock.hpp
```diff
@@ -76,7 +76,7 @@ namespace micm
     // Gamma_i = \sum_j  gamma_{i,j}
     std::array<double, 6> gamma_{};
 
-    double absolute_tolerance_{ 1e-12 };
+    double absolute_tolerance_{ 1e-3 };
```
mattldawson, line 79: I'm going to create a separate issue to figure out how to handle absolute tolerances. They will depend on the units, which for just the gas-phase we can maybe base them on our default units of `mol m-3`, but if we add aerosol-phase species, we might have to rethink this.

include/micm/solver/rosenbrock.hpp
```diff
@@ -585,14 +587,14 @@ namespace micm
 
    protected:
     /// @brief Computes the scaled norm of the vector errors
-    /// @param original_number_densities the original number densities
-    /// @param new_number_densities the new number densities
+    /// @param Y the original vector
+    /// @param new_number_densities the new vector
```
mattldawson, line 591:
```suggestion
    /// @param Ynew the new vector
```

include/micm/solver/rosenbrock.hpp
```diff
@@ -585,14 +587,14 @@ namespace micm
...
     /// @param errors The computed errors
     /// @return
     double NormalizedError(
-        MatrixPolicy<double> original_number_densities,
-        MatrixPolicy<double> new_number_densities,
-        MatrixPolicy<double> errors);
+      const MatrixPolicy<double>& Y,
+      const MatrixPolicy<double>& Ynew,
```
mattldawson, line 596: if we're changing the names, we should probably follow our style of:
```suggestion
      const MatrixPolicy<double>& y,
      const MatrixPolicy<double>& y_new,
```

mattldawson: or we can do this with #129

mattldawson (approved):
looks great! just a couple very minor comments

boulderdaze (approved):

PR #206, opened 2023-09-01
Add inline files

boulderdaze:
Solver class has numerical implementation details that could be separated from the interface which makes the header files a little easier to read. There are two files that I worked on as a demo to discuss whether to have `inl` (inline) files separated from its header files.


mattldawson:
I like this. It does seem easier to read with the function definitions in a separate file, kind of like normal `.cpp` files but for templated classes/functions. Would it make sense to put these in a separate root folder, like `inl/` or `inline/`, or maybe just in `src/`?

boulderdaze:
The projects I've seen that maintain header-only library with additional inline files seem to put them in the same directory but I think we could put them in the separate directory such as `/inl`, or `inline`. One thing I've noticed that most of `micm` don't seem to need to separate header file from inline functions. It may only be applicable to a few with math-heavy header files.

mattldawson:
That sounds good to me - let's keep the `.inl` files in the `include/` folder if that's most commonly done.

K20shores (approved):

mattldawson (approved):

Issue #211, opened 2023-09-06
Post-development design review


mattldawson:
- [x] #604
- [x] #492
- [ ] review user interactions with multiple grid-cell solving to identify difficulties and easily made errors
- [ ] User defined rate constants have an easy API with the state to set their values. Maybe we can come up with something similar for the surface reaction rate constants so that the index of the values doesn't need to be remembered.
- [x] #620
- [x] #587
- [x] #619
- [x] #621
- [ ] Could we redesign the error schemes we have so that we don't need so much boilerplate code?
- [ ] Could we redesign the builders so that parameters don't need to be specified both in the template argument and then passes in a real instance later on?

CUDA-related things:
- [x] #622
- [ ] Change the storage of CUDA sparse matrix to leverage the CuSparse library (replacing the existing CUDA LU Decomposition and Linear Solver).
- [ ] Provide Allison an ill-conditioned matrix and see if she has a better suggestion of a different numerical solver (i.e., not LU).
- [ ] Change the CUDA class so that we only allocate device memory (no host memory allocation). This is specific for arrays of index or constants.
- [ ] Consider moving all temporary matrices onto the `State` so that memory for these temporary variables are only made once
  - https://github.com/NCAR/micm/blob/6b1c58a9be14095e3f3c6df403c91c4e800f23de/include/micm/solver/rosenbrock.inl#L14-L25

PR #229, opened 2023-09-14
228 hires

### LINKED ISSUE #228: HIRES
- [defined here](https://link.springer.com/article/10.1007/BF00276015)
  - [the forcing and jacobian functions](https://www.unige.ch/~hairer/testset/stiff/hires/equation.f)

K20shores:
Closes #228

### DIFF
--- a/test/integration/analytical.cpp
+++ b/test/integration/analytical.cpp
@@ -18,20 +18,21 @@
 #include <utility>
 #include <vector>
 
+#include "hires.hpp"
 #include "oregonator.hpp"
 
 constexpr size_t nsteps = 1000;
 
-double relative_difference(double a, double b){
-  return abs(a - b) / ((a+b)/ 2);
+double relative_difference(double a, double b)
+{
+  return abs(a - b) / ((a + b) / 2);
 }
 
 void writeCSV(
     const std::string& filename,
     const std::vector<std::string>& header,
     const std::vector<std::vector<double>>& data,
-    const std::vector<double>& times
-    )
+    const std::vector<double>& times)
 {
   std::ofstream file(filename);
   if (file.is_open())
@@ -1592,7 +1593,7 @@ TEST(AnalyticalExamples, Robertson)
    * Hairer, E., Wanner, G., 1996. Solving Ordinary Differential Equations II: Stiff and Differential-Algebraic Problems, 2nd
    * edition. ed. Springer, Berlin ; New York. Page 3
    *
-   * solutions are provided here https://www.unige.ch/~hairer/testset/stiff/rober/res_exact_pic
+   * solutions are provided here
    * https://www.unige.ch/~hairer/testset/testset.html
    */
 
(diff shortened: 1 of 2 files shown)


include/micm/solver/rosenbrock.hpp
```diff
@@ -153,8 +154,8 @@ namespace micm
       double final_time_{};
     };
 
-    const System system_;
-    const std::vector<Process> processes_;
+    System system_;
+    std::vector<Process> processes_;
```
mattldawson, line 158: can these no longer be `const`?

include/micm/solver/rosenbrock.hpp
```diff
@@ -189,7 +193,7 @@ namespace micm
     /// @param rate_constants List of rate constants for each needed species
     /// @param number_densities The number density of each species
     /// @param forcing Vector of forcings for the current conditions
-    void CalculateForcing(
+    virtual void CalculateForcing(
```
mattldawson, line 196: is it necessary to make this virtual? it looks like you are using the extending class directly in the tests

include/micm/solver/rosenbrock.hpp
```diff
@@ -210,7 +214,7 @@ namespace micm
     /// @param rate_constants List of rate constants for each needed species
     /// @param number_densities The number density of each species
     /// @param jacobian The matrix of partial derivatives
-    void CalculateJacobian(
+    virtual void CalculateJacobian(
```
mattldawson, line 217: same as above

boulderdaze (approved):

mattldawson (changes requested):
looks good! just a couple minor questions

include/micm/solver/rosenbrock.hpp
K20shores: No, because you can't set base class member variables with this type of syntax:

```
class Derived : Base {
    Derived(int a, int b): a_(a), b_(b){}
};
```

At least, I wasn't able to. I think normally you would just call the base class constructor, but in this case I didn't want the behavior of the base class constructor (setting the jacobian and stuff based off of the processes, which were redundant and not needed here. It would be better to figure out how to make that base solver class like you mentioned in #211.

include/micm/solver/rosenbrock.hpp
K20shores: Well, when I didn't have virtual here, the functions I defined weren't called, I'll try removing this now and see if it breaks

K20shores: Yeah, removing `virtual` here broke the tests

include/micm/solver/rosenbrock.hpp
mattldawson: ah, ok - that makes sense

include/micm/solver/rosenbrock.hpp
K20shores: That's fine. I'll have to translate the HIRES and E5 equations though.

mattldawson: oh, well maybe it's not worth the change then. we can just keep the virtual functions and always change later on if we want to

mattldawson (approved):

PR #233, opened 2023-09-14
lowering computed rate constant

### LINKED ISSUE #232: Tests fail on dercho with nvhpc
```
module load cmake/3.26.3 nvhpc/23.5
```

The analytical test fails

K20shores:
Closes #232. Lowers the rate constant from something on the order of 10^308 to 10^6, which is still huge, but not outrageous. Changing that allowed the test to pass on derecho


sjsprecious (approved):
Thanks @K20shores for sorting this issue out.

mattldawson (approved):

boulderdaze (approved):

test/unit/process/test_troe_rate_constant.cpp
```diff
@@ -85,5 +85,5 @@ TEST(TroeRateConstant, AnalyticalTroeExampleBC)
   double k1 =
       k_0 * 42.2 / (1.0 + k_0 * 42.2 / k_inf) * pow(0.9, 1.0 / (1.0 + (1.0 / 0.8) * pow(log10(k_0 * 42.2 / k_inf), 2)));
 
-  EXPECT_EQ(k, k1);
+  EXPECT_NEAR(k, k1, 1e-8);
```
sjsprecious, line 88: I did not understand why we need to use `expect_near` rather than `expect_eq`. I saw some constants were declared as integer for `troe_params` (e.g., .k0_B_), but a real number was used for the analytical solution (e.g., 167.0 for k_0 calculation). Could this be a problem?

test/regression/RosenbrockChapman/regression_test_p_force.cpp
```diff
@@ -41,7 +41,7 @@ void testRateConstants()
     EXPECT_EQ(state.rate_constants_[i].size(), fixed_state.rate_constants_[0].size());
     for (size_t j{}; j < state.rate_constants_[i].size(); ++j)
     {
-      EXPECT_EQ(state.rate_constants_[i][j], fixed_state.rate_constants_[0][j]);
+      EXPECT_NEAR(state.rate_constants_[i][j], fixed_state.rate_constants_[0][j], 1e-6);
```
sjsprecious, line 44: the same question: why can't we use `expect_eq` here?

sjsprecious (changes requested):

test/unit/process/test_troe_rate_constant.cpp
sjsprecious: According to Kyle, this issue only happens when using the `DEBUG` mode. I checked the relative difference here and it is smaller than 1e-16, which to me could be safely ignored.

test/unit/process/test_ternary_chemical_activation_rate_constant.cpp
```diff
@@ -17,7 +17,7 @@ TEST(TernaryChemicalActivationRateConstant, CalculateWithMinimalArugments)
   auto k = ternary.calculate(state.conditions_[0], params);
   double k0 = 1.0;
   double kinf = 1.0;
-  EXPECT_NEAR(k, k0 / (1.0 + 42.4 * k0 / kinf) * pow(0.6, 1.0 / (1 + pow(log10(42.2 * k0 / kinf), 2))), 0.001);
+  EXPECT_NEAR(k, k0 / (1.0 + 42.4 * k0 / kinf) * std::pow(0.6, 1.0 / (1 + std::pow(std::log10(42.2 * k0 / kinf), 2))), 0.001);
```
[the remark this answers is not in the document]
K20shores: this now matches exactly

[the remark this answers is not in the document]
sjsprecious: Fixed with a later commit

test/unit/process/test_ternary_chemical_activation_rate_constant.cpp
```diff
@@ -37,7 +37,7 @@ TEST(TernaryChemicalActivationRateConstant, CalculateWithAllArugments)
       .Fc_ = 0.9,
       .N_ = 1.2 } };
   auto k = ternary.calculate(state.conditions_[0], params);
-  double k0 = 1.2 * exp(302.3 / temperature) * pow(temperature / 300.0, 2.3);
-  double kinf = 2.6 * exp(402.1 / temperature) * pow(temperature / 300.0, -3.1);
-  EXPECT_NEAR(k, k0 / (1.0 + 42.2 * k0 / kinf) * pow(0.9, 1.0 / (1.0 + 1.0 / 1.2 * pow(log10(42.2 * k0 / kinf), 2))), 0.001);
+  double k0 = 1.2 * std::exp(302.3 / temperature) * std::pow(temperature / 300.0, 2.3);
+  double kinf = 2.6 * std::exp(402.1 / temperature) * std::pow(temperature / 300.0, -3.1);
+  EXPECT_NEAR(k, k0 / (1.0 + 42.2 * k0 / kinf) * std::pow(0.9, 1.0 / (1.0 + 1.0 / 1.2 * std::pow(std::log10(42.2 * k0 / kinf), 2))), 0.001);
```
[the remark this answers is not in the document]
sjsprecious: Fixed with a later commit

sjsprecious (approved):

mwaxmonsky (approved):
Looks good!  Thanks for fixing this!

Dockerfile.nvhpc
```diff
@@ -30,6 +31,14 @@ COPY . /micm/
 # build the library and run the tests
 RUN mkdir /build \
       && cd /build \ 
+      && cmake \
+        -D CMAKE_BUILD_TYPE=release \
```
K20shores, line 35: @mwaxmonsky nvhpc tests seem to fail only in the debug builds of micm, from what I've gathered. Our tests now run the nvidia compiled version of micm, but only as a release build.

PR #247, opened 2023-09-19
Develop 193 jit rate constants

### LINKED ISSUE #193: Add JITed rate constant function

mattldawson:
Adds a rate constant calculation function for `VectorMatrix`. I didn't add a JITed version of this function because there is no indirect addressing and so it probably wouldn't benefit from JIT compiling.

closes #193

### DIFF
--- a/test/unit/process/test_process.cpp
+++ b/test/unit/process/test_process.cpp
@@ -1,0 +1,87 @@
+// Copyright (C) 2023 National Center for Atmospheric Research,
+// SPDX-License-Identifier: Apache-2.0
+#include <gtest/gtest.h>
+
+#include <micm/process/arrhenius_rate_constant.hpp>
+#include <micm/process/process.hpp>
+#include <micm/process/surface_rate_constant.hpp>
+#include <micm/process/user_defined_rate_constant.hpp>
+#include <micm/util/matrix.hpp>
+#include <micm/util/vector_matrix.hpp>
+#include <random>
+
+template<template<class> class MatrixPolicy>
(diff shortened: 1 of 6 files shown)


test/unit/process/test_process.cpp
```diff
@@ -0,0 +1,87 @@
...
+  std::vector<double> params = { 0.0, 0.0, 0.0 };
+  auto get_double = std::bind(std::lognormal_distribution(0.0, 0.01), std::default_random_engine());
+
+  for (std::size_t i_cell = 0; i_cell < number_of_grid_cells; ++i_cell)
+  {
+    state.conditions_[i_cell].temperature_ = get_double() * 285.0;
+    state.conditions_[i_cell].pressure_ = get_double() * 101100.0;
+    params[0] = get_double() * 1.0e-8;
+    params[1] = get_double() * 1.0e5;
+    params[2] = get_double() * 1.0e-2;
```
K20shores, line 44: At some point, I'd like to see if we could come up with a different API for this. Perhaps we could do something that sets these parameters similar to how the user defined rate constants are set on the state.

K20shores (approved):

boulderdaze (approved):

test/unit/process/test_process.cpp
mattldawson: That's what these parameters are - I was just saving them in a vector so I could directly calculate the expected results. But I think they are being used as you describe here: https://github.com/NCAR/micm/blob/b5a2c6bb9550c01c9fe0d0a1b41d1b72a367731b/test/unit/process/test_process.cpp#L45-L49

PR #257, opened 2023-09-26
Decompose cuda

qinatan:
CUDA decompose function passed test. 

Added time measurement for Kernel function.


include/micm/solver/cuda_lu_decomposition.hpp
```diff
@@ -0,0 +1,144 @@
...
+    template<typename T, template<class> typename SparseMatrixPolicy>
+    requires VectorizableSparse<SparseMatrixPolicy<T>>
+    void Decompose(
+        const SparseMatrixPolicy<T>&A, 
+        SparseMatrixPolicy<T>& L, 
+        SparseMatrixPolicy<T>& U) const; 
+    }; 
+
+    template<typename T, typename OrderingPolicy>
+    inline CUDALuDecomposition::CUDALuDecomposition(const SparseMatrix<T, OrderingPolicy>& matrix){
```
sjsprecious, line 29: Why is this function named `CUDAxxx`? It does not seem to run anything on GPU.

include/micm/solver/lu_decomposition.hpp
```diff
@@ -34,12 +33,13 @@ namespace micm
...
   class LuDecomposition
   {
+    public:
     /// number of elements in the middle (k) loops for lower and upper triangular matrices, respectively,
     /// for each iteration of the outer (i) loop
     std::vector<std::pair<std::size_t, std::size_t>> niLU_;
     /// True when A[i][k] is non-zero for each iteration of the middle (k) loop for the upper
     /// triangular matrix; False otherwise
-    std::vector<bool> do_aik_;
+    std::vector<char> do_aik_;
```
sjsprecious, line 42: why change from `bool` to `char`?

test/unit/solver/test_cuda_lu_decomposition.cpp
```diff
@@ -0,0 +1,93 @@
...
+    for (std::size_t j = 0; j < 10; ++j)
+      if (!A.IsZero(i, j))
+        for (std::size_t i_block = 0; i_block < n_grids; ++i_block)
+          A[i_block][i][j] = get_double();
+
+  micm::CUDALuDecomposition gpu_lud(A); 
+  auto gpu_LU = micm::CUDALuDecomposition::GetLUMatrices(A, 1.0e-30); 
+  gpu_lud.Decompose<double, SparseMatrixPolicy>(A, gpu_LU.first, gpu_LU.second); 
+  check_results<double, SparseMatrixPolicy>(
+      A, gpu_LU.first, gpu_LU.second, [&](const double a, const double b) -> void { EXPECT_NEAR(a, b, 1.0e-5); });
```
sjsprecious, line 75: why not use `expect_eq` here?

sjsprecious (changes requested):
Looks good and it is a lot of work! I have a few questions about the code and there are a few conflicts needed to be resolved first.

include/micm/solver/cuda_lu_decomposition.hpp
qinatan: This is the constructor of the derived class CUDALuDecomposition. Constructor in C++ has the same name as their class name.

include/micm/solver/lu_decomposition.hpp
qinatan: I could not use .data() on vector<bool> do_aik to retrieve the memory address of vector because it is a vector<bool> is a special template. The test for functions both on CPU and GPU still passed.

test/unit/solver/test_cuda_lu_decomposition.cpp
qinatan: I followed the test template on CPU code. I can make it expect_eq to if it is still pass.

include/micm/solver/cuda_lu_decomposition.hpp
sjsprecious: Thanks for the clarification. But since the whole function is not running on GPU, can we just remove it or move it to another class defined for CPU? Ideally anything in the `cuda_xxx.hpp` should be only GPU-related.

test/unit/solver/test_cuda_lu_decomposition.cpp
sjsprecious: Yes, please do so. If it fails the `expect_eq` test, I think there is a bug in the GPU code.

K20shores (changes requested):

include/micm/solver/lu_decomposition.hpp
K20shores: The c++ standard chose a poor representation for the `vector<bool>`. The [standard recommends a space-efficient implementation](https://en.cppreference.com/w/cpp/container/vector_bool), so using bits instead of bytes. This is different than how a vector works for other types. I don't really know what this means throughout our code, but I can understand why it may not be supported on the GPU: there would need to be a template specialization for bools only which the NVIDIA engineers probably didn't care to support (rightly)

sjsprecious: Thanks @K20shores for the clarification. Then probably we just update the comments here to explain why to use `char` instead of `bool`?

src/solver/lu_decomposition.cu
```diff
@@ -0,0 +1,177 @@
...
+            size_t lkj_uji_offset = 0; 
+            size_t uii_offset = 0; 
+            
+            if (tid < n_grids){
+                //loop through every element in niLU 
+                for (size_t i = 0; i < niLU_size; i++){
+                    //upper triangular matrix 
+                    auto inLU = device->niLU[i]; 
+                    for (size_t iU = 0; iU < inLU.second; ++iU){
+                        if(device->do_aik[do_aik_offset++]){
```
sjsprecious, line 51: Is `do_aik` here the type of `char`? If so, shall we use something like `device->do_aik[do_aik_offset++] == 'true'`?

include/micm/solver/lu_decomposition.hpp
qinatan: I could keep it bool for CPU code, and copy the bool vector to char vector separately for GPU. Is that better to eliminate uncertainty?

sjsprecious: Hi @qinatan , thanks for proposing this idea. As long as the `char` vector works on CPU, I am fine with your current implementation and I would suggest just updating the comments to make it clear.

test/unit/solver/test_cuda_lu_decomposition.cpp
sjsprecious: @mattldawson , oh, thanks for the additional information. That is helpful. So do you mean that we can not get BFB result even on the CPU end when comparing L * U to the original square matrix?

include/micm/solver/lu_decomposition.hpp
qinatan: updated comment.

include/micm/solver/cuda_lu_decomposition.hpp
qinatan: I eliminated the constructor, and used the constructor from the base class.

include/micm/solver/cuda_lu_decomposition.hpp
```diff
@@ -0,0 +1,144 @@
...
+#ifdef USE_CUDA
+namespace micm{
+    class CUDALuDecomposition: public LuDecomposition{
+    public: 
+    CUDALuDecomposition(){}; 
+    
+    /// @brief Construct an LU decomposition algorithm for a given sparse matrix
+    /// @param matrix Sparse matrix
+    template<typename T, typename OrderingPolicy>
+    CUDALuDecomposition(const SparseMatrix<T, OrderingPolicy>& matrix);
```
K20shores, line 18: Don't you still need to call the parent constructor here?

include/micm/process/cuda_process_set.cuh
```diff
@@ -1,40 +1,17 @@
 #pragma once
```
mattldawson, line 1: Please add the copyright and license text to the top of all the files in `src/` and `include/`

include/micm/solver/cuda_lu_decomposition.hpp
```diff
@@ -0,0 +1,71 @@
...
+#include<micm/solver/lu_decomposition.hpp>
+#include<micm/util/cuda_param.hpp>
+#include <stdexcept>
+#ifdef USE_CUDA
+#include <micm/solver/cuda_lu_decomposition.cuh>
+#endif 
+
+#ifdef USE_CUDA
+namespace micm{
+    class CUDALuDecomposition: public LuDecomposition{
```
mattldawson, line 11:
```suggestion
    class CudaLuDecomposition: public LuDecomposition{
```
please rename for consistency with `CudaProcessSet` and to follow style of capitalizing only first character of acronyms in class/stuct/function names

include/micm/util/cuda_param.hpp
```diff
@@ -0,0 +1,62 @@
+#include<vector>
+
+#ifndef CUDA_PARAM_HPP
+#define CUDA_PARAM_HPP
+  //member data of class CUDAProcessSet grouped in struct passing to kernel driver function 
+  struct CUDAProcessSetParam{
```
mattldawson, line 6:
```suggestion
  struct CudaProcessSetParam{
```

include/micm/util/cuda_param.hpp
```diff
@@ -0,0 +1,62 @@
...
+    const size_t* number_of_products; 
+    const size_t* product_ids; 
+    size_t product_ids_size; 
+    const double* yields; 
+    size_t yields_size; 
+    const size_t* jacobian_flat_ids; 
+    size_t jacobian_flat_ids_size; 
+}; 
+
+  struct CUDASolverParam{
```
mattldawson, line 19:
```suggestion
  struct CudaSolverParam{
```

include/micm/util/cuda_param.hpp
```diff
@@ -0,0 +1,62 @@
...
+    size_t aki_size;
+    const std::pair<size_t, size_t>* lki_nkj;
+    size_t lki_nkj_size; 
+    const std::pair<size_t, size_t>* lkj_uji;  
+    size_t lkj_uji_size;  
+    const size_t* uii; 
+    size_t uii_size; 
+  };
+ //different matrix data grouped in struct passing to kernel driver function 
+    struct CUDAMatrixParam{
```
mattldawson, line 42:
```suggestion
    struct CudaMatrixParam{
```

include/micm/util/cuda_param.hpp
```diff
@@ -0,0 +1,62 @@
...
+     const double* rate_constants; 
+     const double* state_variables; 
+     double* forcing; 
+     size_t n_grids; 
+     size_t n_reactions; 
+     size_t n_species; 
+}; 
+//sparseMatrix data grouped in struct passing to kernel driver function 
+struct CUDASparseMatrixParam{
```
mattldawson, line 51:
```suggestion
struct CudaSparseMatrixParam{
```

include/micm/util/cuda_param.hpp
```diff
@@ -0,0 +1,62 @@
...
+    const size_t* number_of_reactants; 
+    const size_t* reactant_ids; 
+    size_t reactant_ids_size; 
+    const size_t* number_of_products; 
+    const size_t* product_ids; 
+    size_t product_ids_size; 
+    const double* yields; 
+    size_t yields_size; 
+    const size_t* jacobian_flat_ids; 
+    size_t jacobian_flat_ids_size;
```
mattldawson, line 16: please update to use a trailing underscore for class/struct data members

include/micm/util/cuda_param.hpp
```diff
@@ -0,0 +1,62 @@
...
+    const char* do_aki;
+    size_t do_aki_size; 
+    const size_t* aki; 
+    size_t aki_size;
+    const std::pair<size_t, size_t>* lki_nkj;
+    size_t lki_nkj_size; 
+    const std::pair<size_t, size_t>* lkj_uji;  
+    size_t lkj_uji_size;  
+    const size_t* uii; 
+    size_t uii_size;
```
mattldawson, line 39: please update to use a trailing underscore for class/struct data members

include/micm/util/cuda_param.hpp
```diff
@@ -0,0 +1,62 @@
...
+    size_t uii_size; 
+  };
+ //different matrix data grouped in struct passing to kernel driver function 
+    struct CUDAMatrixParam{
+     const double* rate_constants; 
+     const double* state_variables; 
+     double* forcing; 
+     size_t n_grids; 
+     size_t n_reactions; 
+     size_t n_species;
```
mattldawson, line 48: please update to use a trailing underscore for class/struct data members

include/micm/util/cuda_param.hpp
```diff
@@ -0,0 +1,62 @@
...
+struct CUDASparseMatrixParam{
+   double* jacobian; 
+   size_t jacobian_size; 
+   const double* A; 
+   size_t A_size; 
+   double* L; 
+   size_t L_size; 
+   double* U;
+   size_t U_size; 
+   size_t n_grids;
```
mattldawson, line 60: please update to use a trailing underscore for class/struct data members

src/process/process_set.cu
```diff
@@ -4,71 +4,92 @@
 
 #include <chrono>
 #include <iostream>
+#include <micm/util/cuda_param.hpp>
+
+//device pointers passing to AddForcingTermsKernel()
+typedef struct forcingDevice{
```
mattldawson, line 10:
```suggestion
typedef struct ForcingDevice{
```

src/process/process_set.cu
```diff
@@ -4,71 +4,92 @@
...
+  double* rate_constants; 
+  double* state_variables; 
+  double* forcing; 
+  size_t* number_of_reactants; 
+  size_t* reactant_ids; 
+  size_t* number_of_products; 
+  size_t* product_ids; 
+  double* yields;
```
mattldawson, line 18: please update to use a trailing underscore for class/struct data members

src/process/process_set.cu
```diff
@@ -4,71 +4,92 @@
...
+  double* state_variables; 
+  double* forcing; 
+  size_t* number_of_reactants; 
+  size_t* reactant_ids; 
+  size_t* number_of_products; 
+  size_t* product_ids; 
+  double* yields; 
+};
+//device pointers passing to AddJacobianTermsKernel() 
+typedef struct jacobianDevice{
```
mattldawson, line 21:
```suggestion
typedef struct JacobianDevice{
```

src/process/process_set.cu
```diff
@@ -4,71 +4,92 @@
...
+  double* rate_constants; 
+  double* state_variables; 
+  double* jacobian;
+  size_t* number_of_reactants; 
+  size_t* reactant_ids; 
+  size_t* number_of_products; 
+  double* yields; 
+  size_t* jacobian_flat_ids;
```
mattldawson, line 29: please update to use a trailing underscore for class/struct data members

src/process/process_set.cu
```diff
@@ -4,71 +4,92 @@
...
+  double* state_variables; 
+  double* jacobian;
+  size_t* number_of_reactants; 
+  size_t* reactant_ids; 
+  size_t* number_of_products; 
+  double* yields; 
+  size_t* jacobian_flat_ids; 
+};
+
+const size_t BLOCK_SIZE = 320;
```
mattldawson, line 32: is this a constant that applies to all the CUDA code? if so, should it be defined in a common location?

src/process/process_set.cu
```diff
@@ -4,71 +4,92 @@
...
+        forcingDevice* device, 
         size_t n_grids,
         size_t n_reactions,
-        size_t n_species,
-        size_t* number_of_reactants_,
-        size_t* reactant_ids_,
-        size_t* number_of_products_,
-        size_t* product_ids_,
-        double* yields_)
+        size_t n_species)
```
mattldawson, line 43: would it make sense to also include these in the `ForcingDevice` struct?

src/process/process_set.cu
```diff
@@ -4,71 +4,92 @@
...
+        jacobianDevice* device,
         size_t n_grids,
-        size_t n_reactions,
-        double* jacobian,
-        size_t* number_of_reactants,
-        size_t* reactant_ids,
-        size_t* number_of_products,
-        double* yields,
-        size_t* jacobian_flat_ids)
+        size_t n_reactions)
```
mattldawson, line 82: would it make sense to also include these in the `JacobianDevice` struct?

src/solver/lu_decomposition.cu
```diff
@@ -0,0 +1,177 @@
+#include <iostream> 
+#include <vector>
+#include <micm/util/cuda_param.hpp> 
+const size_t BLOCK_SIZE = 320; 
+struct decomposeDevice{
```
mattldawson, line 5:
```suggestion
struct DecomposeDevice{
```

src/solver/lu_decomposition.cu
```diff
@@ -0,0 +1,177 @@
...
+    char* do_aik; 
+    size_t* aik; 
+    char* do_aki;
+    size_t* aki;  
+    size_t* uii; 
+    std::pair<size_t,size_t>* niLU;
+    std::pair<size_t, size_t>* uik_nkj; 
+    std::pair<size_t, size_t>* lij_ujk;
+    std::pair<size_t, size_t>* lki_nkj; 
+    std::pair<size_t, size_t>* lkj_uji;
```
mattldawson, line 18: please update to use a trailing underscore for class/struct data members

src/solver/lu_decomposition.cu
mattldawson: `do_aik[do_aik_offset]` is a single `char`, not a string. `true` will be stored as `1` and `false` as `0` (or really the ASCII character whose index is 1 or 0). It's ok to use as it is written here.

mattldawson (changes requested):
looks great! just some minor style things, but really nice work!

src/solver/lu_decomposition.cu
mattldawson: `do_aik[do_aik_offset]` is a single `char`, not a string. `true` will be stored as `1` and `false` as `0` (or really the ASCII character whose index is 1 or 0). It's ok to use as it is written here.

sjsprecious: Thanks for the clarification.

include/micm/solver/cuda_lu_decomposition.hpp
qinatan: No. I believe it is called when creating a instance of this class. This occurs at the testing function.

K20shores: I don't think that's true. I think the only constructor called by default is the one with no arguments. Since we have arguments I believe you need to call it explicitly. You can check if this is true with a print statement

qinatan: Yes you are right. I meant this constructor is explicitly called when creating an instance of the class in the test function:  micm::LuDecomposition gpu_lud(A);  Does it need to be call inside the class definition, too?

test/unit/solver/test_cuda_lu_decomposition.cpp
qinatan: Ok. I will write a test for that.

include/micm/solver/cuda_lu_decomposition.hpp
```diff
@@ -0,0 +1,144 @@
...
+            do_aki_.push_back(true);
+            aki_.push_back(matrix.VectorIndex(0, k, i));
+            }
+            uii_.push_back(LU.second.VectorIndex(0, i, i));
+            lki_nkj_.push_back(std::make_pair(LU.first.VectorIndex(0, k, i), nkj));
+            ++(iLU.first);
+        }
+        niLU_.push_back(iLU);
+    }
+  }
```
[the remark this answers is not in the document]
qinatan: Yes. I eliminated this constructor. Called the parent constructor explicitly when creating an instance.

include/micm/process/cuda_process_set.cuh
qinatan: Done.

include/micm/solver/cuda_lu_decomposition.hpp
qinatan: Done.

include/micm/util/cuda_param.hpp
qinatan: Done.

include/micm/util/cuda_param.hpp
qinatan: Done.

include/micm/util/cuda_param.hpp
qinatan: Done.

src/process/process_set.cu
qinatan: My initial thought was to only group the vectors. It does make sense to include these variables. Changes made.

src/process/process_set.cu
qinatan: changes made.

src/solver/lu_decomposition.cu
qinatan: changes made.

src/solver/lu_decomposition.cu
qinatan: added trailing lines to all struct members include AddForcing and AddJacobians.

include/micm/solver/lu_decomposition.hpp
```diff
@@ -52,9 +53,10 @@ namespace micm
     std::vector<std::pair<std::size_t, std::size_t>> lij_ujk_;
     /// True when A[k][i] is non-zero for each iteration of the middle (k) loop for the lower
     /// triangular matrix; False otherwise
-    std::vector<bool> do_aki_;
+    std::vector<char> do_aki_;
     /// Index in A.data_ for A[k][i] for each iteration of the middle (k) loop for the lower
-    /// triangular matrix when A[k][i] is non-zero
+    /// triangular matrix when A[k][i] is non-zero. Used data type char instead of bool because vector<bool> representation
```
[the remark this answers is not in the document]
qinatan: Thanks for catching that! Change made.

test/unit/solver/test_cuda_lu_decomposition.cpp
qinatan: test for checking GPU code against CPU code added, and test passed.

sjsprecious:
@qinatan can you resolve the conflicting files in your PR before I test it on Derecho?

K20shores (approved):

sjsprecious:
@qinatan The following two tests failed with your latest branch:
```
7 - troe_rate_constant (Failed)
10 - cuda_process_set (SEGFAULT)
```

K20shores:
> @qinatan The `troe_rate_constant` test still fails somehow with your latest branch.
> 
> ```
> 7 - troe_rate_constant (Failed)
> ```

@qinatan we've ran into this failing before, specifically with nvhpc. Does this test happen to pass if you make a release build?

mattldawson:
> > @qinatan The `troe_rate_constant` test still fails somehow with your latest branch.
> > ```
> > 7 - troe_rate_constant (Failed)
> > ```
> 
> @qinatan we've ran into this failing before, specifically with nvhpc. Does this test happen to pass if you make a release build?

@sjsprecious - were you building with `Debug` or `Release`?

sjsprecious:
> > > @qinatan The `troe_rate_constant` test still fails somehow with your latest branch.
> > > ```
> > > 7 - troe_rate_constant (Failed)
> > > ```
> > 
> > 
> > @qinatan we've ran into this failing before, specifically with nvhpc. Does this test happen to pass if you make a release build?
> 
> @sjsprecious - were you building with `Debug` or `Release`?

@mattldawson @K20shores @qinatan My bad! Yes, I recalled this issue now and I was using the `DEBUG` build option before. Switching to the `release` option passes the `troe_rate_constant` test. However, it then fails the `regression_test_rosenbrock_p_force` test somehow.

mattldawson:
> > > > @qinatan The `troe_rate_constant` test still fails somehow with your latest branch.
> > > > ```
> > > > 7 - troe_rate_constant (Failed)
> > > > ```
> > > 
> > > 
> > > @qinatan we've ran into this failing before, specifically with nvhpc. Does this test happen to pass if you make a release build?
> > 
> > 
> > @sjsprecious - were you building with `Debug` or `Release`?
> 
> @mattldawson @K20shores @qinatan My bad! Yes, I recalled this issue now and I was using the `DEBUG` build option before. Switching to the `release` option passes the `troe_rate_constant` test. However, it then fails the `regression_test_rosenbrock_p_force` test somehow.

@sjsprecious Would you be able to run this test using the `main` branch? I think it's likely this is unrelated to this particular PR, and just slipped by because we have no automated testing with the NVidia compiler on GPUs. If `main` fails in the same way, let's create a separate issue for this (and see if we can come up with some solution for automated testing on GPUs, if possible)

sjsprecious:
> > > > > @qinatan The `troe_rate_constant` test still fails somehow with your latest branch.
> > > > > ```
> > > > > 7 - troe_rate_constant (Failed)
> > > > > ```
> > > > 
> > > > 
> > > > @qinatan we've ran into this failing before, specifically with nvhpc. Does this test happen to pass if you make a release build?
> > > 
> > > 
> > > @sjsprecious - were you building with `Debug` or `Release`?
> > 
> > 
> > @mattldawson @K20shores @qinatan My bad! Yes, I recalled this issue now and I was using the `DEBUG` build option before. Switching to the `release` option passes the `troe_rate_constant` test. However, it then fails the `regression_test_rosenbrock_p_force` test somehow.
> 
> @sjsprecious Would you be able to run this test using the `main` branch? I think it's likely this is unrelated to this particular PR, and just slipped by because we have no automated testing with the NVidia compiler on GPUs. If `main` fails in the same way, let's create a separate issue for this (and see if we can come up with some solution for automated testing on GPUs, if possible)

Thanks @mattldawson . Yes, the same test fails for the `main` branch, either. I agree that this is unrelated to Qina's work here and should be addressed separately. In terms of testing MICM on GPU, we are using a self-hosted runner on Derecho to test another code on the GPU through the GitHub workflow. It is not 100% automatic but it could be a solution temporarily. Happy to chat offline about this idea.

sjsprecious (approved):
Nice work and thank you for addressing all my comments! @qinatan

mattldawson (approved):
looks great! very nice work!

PR #289, opened 2023-10-05
putting docker files into their own folder

K20shores:
I put all of the docker files into a folder. This kinda cleans up the repository a little but is not necessary. Feel free to reject


mattldawson (approved):
I like it - there are getting to be a lot of them

PR #294, opened 2023-10-06
292 add a docker image publish

### LINKED ISSUE #292: Add a docker image publish

K20shores:
Closes #292 

Only one platform is published. When I tried to add `linux/arm64` the action would take forever, and likely will end up timing out. 

I filed an issue: https://github.com/docker/build-push-action/issues/982. Maybe it'll get resolved, but I kinda doubt it.

So for now, we publish for one platform


mattldawson (approved):
nice!

PR #295, opened 2023-10-10
Test kpp micm

dwfncar:
A Chapman test example which uses MICM config files output from the KPP to MICM translator.


etc/scripts/kpp_to_micm.py
```diff
@@ -129,12 +135,14 @@ def micm_species_json(lines, fixed=False, tolerance=1.0e-12):
...
 
-def parse_kpp_arrhenius(kpp_str):
+def parse_kpp_arrhenius(kpp_str, to_si_units=False):
     """
     Parse KPP Arrhenius reaction
 
     Parameters
         (str) kpp_str: Arrhenius reaction string
+        (bool) to_si_units: convert A coefficient
+        from cm^3 molecule-1 s-1 to m^3 mol-1 s-1
```
mattldawson, line 145: I think this should convert from `(cm^3 molecule-1)^(N-1) s-1` to `(m^3 mol-1)^(N-1) s-1` where `N` is the number of reactants, because we don't restrict Arrhenius reactions to have exactly two reactants

etc/scripts/kpp_to_micm.py
```diff
@@ -188,16 +196,20 @@ def parse_kpp_arrhenius(kpp_str):
         arr_dict['D'] = 300.0
     else:
         logging.error('unrecognized KPP Arrhenius syntax')
+    if to_si_units:
+        arr_dict['A'] *= (N_Avogadro * 1.0e-6)
```
mattldawson, line 200: see above comment

etc/scripts/kpp_to_micm.py
```diff
@@ -232,12 +244,15 @@ def micm_equation_json(lines):
...
-            equation_dict = parse_kpp_arrhenius(coeffs)
+            equation_dict = parse_kpp_arrhenius(coeffs,
+                to_si_units=to_si_units)
         else:
             # default to Arrhenius with a single coefficient
             coeffs = coeffs.replace('(', '').replace(')', '')
             equation_dict['type'] = 'ARRHENIUS' 
             equation_dict['A'] = float(coeffs)
+            if to_si_units:
+                equation_dict['A'] *= (N_Avogadro * 1.0e-6)
```
mattldawson, line 255: see above comment

mattldawson (changes requested):

K20shores (approved):

mattldawson (approved):

PR #307, opened 2023-10-12
Actions

K20shores:
Reorganized the actions so that we have explicit files for mac, windows, ubuntu, and the docker files.

For mac and ubuntu, I test both debug and release versions for gcc and clang. mac also tests 4 versions of xcode across 2 versions of mac. Windows has many tests that I stole from [nlohmann/json](https://github.com/nlohmann/json/blob/develop/.github/workflows/windows.yml). 

I'm not sure which we should enable for PRs only.


mattldawson (approved):
nice! they all seem to run pretty fast, so maybe we just move some to only run on PRs if, in the future, the tests take longer?

PR #314, opened 2023-10-13
Develop 245 terminator

### LINKED ISSUE #245: Create Terminator chemistry configuration and test

mattldawson:
closes #245 

Includes the analytical solution from CAM-Chem in the test for the Terminator configuration.

Also, I updated slightly the logic around min/max/initial solver time steps to avoid changing solver parameters during the solve and to allow users to be able to specify these (which I did in the terminator test to get closer results to the analytical solution).

And, I added the ability to include species whose concentrations are not solved for, but are parameterized on environmental conditions. At first I did this because I mistakenly thought the third-body species "M" was included in the Terminator configuration. I left this option in the PR because it will be useful for other mechanisms that include "M" and for setting constant values for species like O2 and N2

### DIFF
--- a/include/micm/process/process.hpp
+++ b/include/micm/process/process.hpp
@@ -36,13 +36,11 @@ namespace micm
     /// @param processes The set of processes being solved
     /// @param state The solver state to update
     template<template<class> class MatrixPolicy>
-    requires(!VectorizableDense<MatrixPolicy<double>>) static void UpdateState(
-        const std::vector<Process>& processes,
-        State<MatrixPolicy>& state);
+      requires(!VectorizableDense<MatrixPolicy<double>>)
+    static void UpdateState(const std::vector<Process>& processes, State<MatrixPolicy>& state);
     template<template<class> class MatrixPolicy>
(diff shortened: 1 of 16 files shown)


K20shores (approved):

K20shores:
@mattldawson should we also update the configuration parser to be able to make the parameterized species?

mattldawson:
> @mattldawson should we also update the configuration parser to be able to make the parameterized species?

Yes, I think that makes sense, but I'm wondering how we should go about this, because the open-atmos schema doesn't include this option currently. I feel like this is more of a solver option than a description of the chemical system, so maybe we do a MICM-specific addition to the configuration format?

K20shores:
> > @mattldawson should we also update the configuration parser to be able to make the parameterized species?
> 
> Yes, I think that makes sense, but I'm wondering how we should go about this, because the open-atmos schema doesn't include this option currently. I feel like this is more of a solver option than a description of the chemical system, so maybe we do a MICM-specific addition to the configuration format?

This feels like something we should add to the open atmos format though. It feels useful

PR #317, opened 2023-10-17
164 prevent non standard keys in json

### LINKED ISSUE #164: Prevent non-standard keys in JSON?
For TUV-x we decided that non-standard keys can only be used if they start with `__`. This helps avoid the situation where someone misspells an optional key name and gets unexpected results. Should we do the same for MICM?

K20shores:
Closes #164 

- Detects any usage of nonstandard keys for all reaction types, products, reactants
- Added configuration tests for arrhenius and troe reactions
- removed the MZ whatever configuration. That was supposed to be a TS1 reaction but it wasn't. That's supposed to be made in a future task anyways


test/unit/unit_configs/process/tunneling/valid/reactions.json
```diff
@@ -5,6 +5,7 @@
       "type": "MECHANISM",
       "reactions": [
         {
+          "__my comment": {},
```
boulderdaze, line 8: Just curious, what is this information for?

boulderdaze (approved):

mattldawson (approved):

dwfncar (approved):

PR #326, opened 2023-10-24
Command-line driver for MICM

boulderdaze:
Closes [#308](https://github.com/NCAR/micm/issues/308)

Created try-out example configurations with a single grid cell and a set of initial conditions from a CSV file

- chapman
- TS1
- robertson
- carbon bond 5

PR #356, opened 2023-11-13
182 surface reaction example

### LINKED ISSUE #182: Add surface reaction analytical example
In #180, all analytical examples that CAMP has for the same reactions we have were implemented except for the surface reactions. Do those. 

https://github.com/open-atmos/camp/blob/main/test/unit_rxn_data/test_rxn_surface.F90

dwfncar:
Waiting for https://github.com/NCAR/micm/pull/326 merge, to add the new state printer header and print state methods.
Should this test be made into a full science example?
Setup was copied from https://github.com/open-atmos/camp/blob/main/test/unit_rxn_data/test_rxn_surface.F90

closes #182


K20shores:
@dwfncar could you move this to the [analytical tests](https://github.com/NCAR/micm/blob/main/test/integration/analytical_policy.hpp)? That's where the other tests like this are

dwfncar:
> @dwfncar could you move this to the [analytical tests](https://github.com/NCAR/micm/blob/main/test/integration/analytical_policy.hpp)? That's where the other tests like this are

Yes, hadn't ever looked at that file, so wasn't sure where this belonged.  Will add the EXPECT_NEAR assertions as well.

dwfncar:
Added TEST(AnalyticalExamplesRosenbrock, SurfaceRxn) and JIT version.  PR ready for review.

K20shores (approved):

PR #369, opened 2023-11-20
153 config parser

### LINKED ISSUE #153: Status messages for config parser
Should we add a message string to the status object returned by the config parser? For a large mechanism, it might be helpful to know which reaction, species, etc. had an invalid configuration and what specifically was missing or incorrectly formatted.

dwfncar:
Capture last species, tolerance or reaction JSON.
Add to the error message when any return status is not SUCCESS.

closes #153


test/unit/configure/test_solver_config.cpp
```diff
@@ -16,6 +16,20 @@ TEST(SolverConfig, NoConfigFilesFound)
...
+  micm::SolverConfig solverConfig{};
+  auto status = solverConfig.ReadAndParse("./unit_configs/CAMP/camp_unknown_key/config.json");
+  EXPECT_EQ(micm::ConfigParseStatus::UnknownKey, status);
+  try
+  {
+    micm::SolverParameters solver_params = solverConfig.GetSolverParams();
+  }
+  catch (...)
+  {
+  }
```
boulderdaze, line 30: I think it'd be nice to specify `catch` block with `std::runtime_error` since we know what will be thrown

boulderdaze (approved):

K20shores (approved):

Issue #387, opened 2024-01-04
Move device data management of `CudaLuDecomposition` data members to constructor/destructor


mattldawson:
Move device memory allocation, deallocation, and copying to `CudaLuDecomposition`.

Acceptance Criteria:
- device pointers for all index array data members of `CudaLuDecomposition` are created during construction
- all device pointer data members are freed during deconstruction

**Ideas**
- create a struct that can hold the device pointer for LU decomposition, and make an instance of this struct a data member of `CudaLuDecomposition`
- pass a reference to the device pointer struct to the CUDA solve kernel
- possibly create functions in `cuda_lu_decomposition.cuh` and `lu_decomposition.cu` for the allocation/memcpy and for deallocation of members of the device pointer struct (because `cudaMalloc`, `cudaMemcpy`, and `CudaFree` will not be callable in the `*.hpp` files because these aren't compiled with the cuda compiler). These functions can then just be called in the `CudaLuDecomposition` constructor and deconstructor.

PR #399, opened 2024-01-08
Create constructor/destructor for `CudaLuDecomposition` class to handle data movement between host and device

sjsprecious:
This PR introduces a new constructor/destructor to the `CudaLuDecomposition` class that handles the data movement between host and device. It also cleans up some interfaces and adds more comments for future information.

```
Currently Loaded Modules:
```
the unit test shows that only the following test fails:
```
97% tests passed, 1 tests failed out of 39

Total Test time (real) =  17.67 sec

The following tests FAILED:
	 18 - cuda_rosenbrock (SEGFAULT)
```
However, I do not understand why revising the `CudaLuDecomposition` class will fail the test of `cuda_rosenbrock`. @mattldawson @K20shores do you have any ideas here?

Fixed https://github.com/NCAR/micm/issues/387.


sjsprecious:
@mattldawson @K20shores Just for your information: all the tests passed for commit `72987a1`. The failed test showed up since commit `4414751` but I did not see why that change would affect the `cuda_rosenbrock` test. If it does not make sense to you, either, I prefer to ignore this failure and work on other issues first (the error may disappear "magically").

include/micm/solver/cuda_lu_decomposition.hpp
```diff
@@ -11,26 +11,85 @@
...
+        hostptr->aki_size_     = this->aki_.size();
+        hostptr->lki_nkj_size_ = this->lki_nkj_.size();
+        hostptr->lkj_uji_size_ = this->lkj_uji_.size();
+        hostptr->uii_size_     = this->uii_.size(); 
+        
+        // Copy the data from host struct to device struct
+        micm::cuda::CopyConstData(hostptr, this->devptr);
+
+        /// Free the host memory for "hostptr"
+        delete hostptr;
```
K20shores, line 63:
```suggestion
        LuDecomposeConst hostptr = LuDecomposeConst;
        hostptr->niLU_         = this->niLU_.data();
        hostptr->do_aik_       = this->do_aik_.data();
        hostptr->aik_          = this->aik_.data();
        hostptr->uik_nkj_      = this->uik_nkj_.data();
        hostptr->lij_ujk_      = this->lij_ujk_.data();
        hostptr->do_aki_       = this->do_aki_.data();
        hostptr->aki_          = this->aki_.data();
        hostptr->lki_nkj_      = this->lki_nkj_.data();
        hostptr->lkj_uji_      = this->lkj_uji_.data();
        hostptr->uii_          = this->uii_.data();
        hostptr->niLU_size_    = this->niLU_.size();
        hostptr->do_aik_size_  = this->do_aik_.size();
        hostptr->aik_size_     = this->aik_.size();
        hostptr->uik_nkj_size_ = this->uik_nkj_.size();
        hostptr->lij_ujk_size_ = this->lij_ujk_.size();
        hostptr->do_aki_size_  = this->do_aki_.size();
        hostptr->aki_size_     = this->aki_.size();
        hostptr->lki_nkj_size_ = this->lki_nkj_.size();
        hostptr->lkj_uji_size_ = this->lkj_uji_.size();
        hostptr->uii_size_     = this->uii_.size(); 
        
        // Copy the data from host struct to device struct
        micm::cuda::CopyConstData(&hostptr, this->devptr);
```

Since it's a temporary object, there's no reason to manage the memory ourselves; a stack variable seems fine to me here

src/solver/lu_decomposition.cu
```diff
@@ -4,190 +4,199 @@
...
             ++uii_offset;
           }
         }
       }
-    }  // end of kernel
+    }  // end of CUDA kernel
+
+    /// This is the function that will copy the constant data
+    ///   members of class "CudaLuDecomposition" to the device
+    void CopyConstData(LuDecomposeConst* hostptr, LuDecomposeConst*& devptr)
```
K20shores, line 92:
```suggestion
    void CopyConstData(LuDecomposeConst* hostptr, LuDecomposeConst* devptr)
```

Only the `*` is needed here

include/micm/solver/cuda_lu_decomposition.cuh
```diff
@@ -3,13 +3,27 @@
...
-}  // namespace micm
+            LuDecomposeConst* devptr);
+
+    /// This is the function that will copy the constant data
+    ///   members of class "CudaLuDecomposition" to the device;
+    /// Note that if we want to allocate "devptr" inside this function,
+    ///   it must be declared as "LuDecomposeConst*&"; otherwise,
+    ///   passing "devptr->d_niLU_" as an argument to the CUDA kernel
+    ///   will trigger a segmentation fault;
+    void CopyConstData(LuDecomposeConst* hostptr, LuDecomposeConst*& devptr);
```
K20shores, line 22:
```suggestion
    void CopyConstData(LuDecomposeConst* hostptr, LuDecomposeConst* devptr);
```

Only the `*` is needed here

include/micm/solver/cuda_lu_decomposition.cuh
```diff
@@ -3,13 +3,27 @@
...
+   
+    /// This is the function that will delete the constant data
+    ///   members of class "CudaLuDecomposition" on the device
+    void FreeConstData(LuDecomposeConst*& devptr);
```
K20shores, line 26:
```suggestion
    void FreeConstData(LuDecomposeConst* devptr);
```

Only the `*` is needed here

src/solver/lu_decomposition.cu
```diff
@@ -4,190 +4,199 @@
...
+      cudaMemcpy(devptr->lki_nkj_, hostptr->lki_nkj_, lki_nkj_bytes, cudaMemcpyHostToDevice);
+      cudaMemcpy(devptr->lkj_uji_, hostptr->lkj_uji_, lkj_uji_bytes, cudaMemcpyHostToDevice);
+      cudaMemcpy(devptr->uii_,     hostptr->uii_,     uii_bytes,     cudaMemcpyHostToDevice);
+      devptr->niLU_size_ = hostptr->niLU_size_;
+    }
 
-    std::chrono::nanoseconds DecomposeKernelDriver(CudaSparseMatrixParam& sparseMatrix, CudaSolverParam& solver)
+    /// This is the function that will delete the constant data
+    ///   members of class "CudaLuDecomposition" on the device
+    void FreeConstData(LuDecomposeConst*& devptr)
```
K20shores, line 142:
```suggestion
    void FreeConstData(LuDecomposeConst* devptr)
```

Only the `*` is needed here

K20shores (changes requested):
Just a few small syntactic changes seem to be needed

include/micm/solver/cuda_lu_decomposition.hpp
sjsprecious: So if we make it a stack variable, then shall we access it through something like `hostptr.niLU_`?

src/solver/lu_decomposition.cu
sjsprecious: Actually no. Because I will allocate the memory space for `devptr` inside the `CopyConstData` function, I have to add the `&` here to make sure that it is reflected outside this functions. Otherwise I will get a segmentation fault when calling the CUDA kernel. See my comments at https://github.com/NCAR/micm/blob/update_cuda_class/include/micm/solver/cuda_lu_decomposition.cuh#L16.

include/micm/solver/cuda_lu_decomposition.cuh
sjsprecious: see above.

include/micm/solver/cuda_lu_decomposition.cuh
sjsprecious: see above.

src/solver/lu_decomposition.cu
sjsprecious: see above.

include/micm/solver/cuda_lu_decomposition.hpp
K20shores: yes

src/solver/lu_decomposition.cu
mattldawson: I think you probably want to make this a return value of the function, and I'm not sure if you really need to allocate memory for this. I would suggest something like:
```
LuDecomponseConst CopyConstData(LuDecomposeConst& hostptr)
```
and then just have the `CudaLuDecomposition` data member be:
```
LuDecompositionConst devptr_;
```
(instead of being a pointer)

include/micm/solver/cuda_lu_decomposition.cuh
```diff
@@ -3,13 +3,27 @@
...
   namespace cuda
   {
+    /// This is the host function that will call the CUDA kernel
+    ///   to perform LU decomposition on the device
     std::chrono::nanoseconds DecomposeKernelDriver(
             CudaSparseMatrixParam& sparseMatrix, 
-            CudaSolverParam& solver);
-  }  // namespace cuda
-}  // namespace micm
+            LuDecomposeConst* devptr);
```
mattldawson, line 14:
```suggestion
            LuDecomposeConst& devptr);
```
I think, in most cases, it's best to prefer reference over pointer arguments

src/solver/lu_decomposition.cu
```diff
@@ -4,190 +4,199 @@
...
+    __global__ void DecomposeKernel(const double* d_A, double* d_L, double* d_U,
+                                    std::pair<size_t, size_t>* d_niLU,
+                                    char* d_do_aik, size_t* d_aik,
+                                    std::pair<size_t, size_t>* d_uik_nkj,
+                                    std::pair<size_t, size_t>* d_lij_ujk,
+                                    char* d_do_aki, size_t* d_aki,
+                                    std::pair<size_t, size_t>* d_lki_nkj,
+                                    std::pair<size_t, size_t>* d_lkj_uji,
+                                    size_t* d_uii,
+                                    size_t niLU_size, size_t ngrids)
```
mattldawson, line 22: Could the kernel just take as an argument?:
```
LuDecomposeConst& devptr
```

src/solver/lu_decomposition.cu
sjsprecious: Thanks @mattldawson . I am actually struggling with pointers in C++ a lot and I try to avoid using it unless necessary. I think what you suggest makes sense and I am working on it now.

mattldawson (changes requested):
Looks great! A couple minor suggestions:

In the case of `hostptr` and `devptr`, as Kyle mentioned, I would suggest not having them be pointers that we have to manage memory for.

I would also suggest passing function arguments as copies, or by reference, as opposed to pointers, when possible, and returning output from a function, as opposed to updating reference or pointer arguments (unlike Fortran =)

src/solver/lu_decomposition.cu
sjsprecious: The reason I switch to the long list here is that the original implementation had `devptr->xxx` inside the loops and I thought there would be lots of overhead time by doing the deference. But if doing `devptr->xxx` or `devptr.xxx` is as efficient as accessing `xxx` directly, I could try the one argument version as suggested.

K20shores: A compromise could be passing in a `LuDecomposeConst&` and saving all of the arguments here as local variables. It increases readability while preventing the performance hit that you're concerned about since all data access would then be stack variables

sjsprecious: Thanks @K20shores . I just did what you suggest (hopefully I understand it correctly) and it worked.

One thing I need to mention: using `LuDecomposeConst&` passes the compilation but the execution of this CUDA test will hang. Switching to passing by value (i.e., `LuDecomposeConst`) fixes the runtime problem but I do not understand why.

include/micm/solver/cuda_lu_decomposition.cuh
sjsprecious: I just update the pointers to be references.

src/solver/lu_decomposition.cu
sjsprecious: @mattldawson @K20shores I just updated the `CopyConstData` to return the `LuDecomponseConst` type. I still generate a local `LuDecomponseConst` struct inside the `CopyConstData` function but let me know if this is not what I am supposed to do.

include/micm/solver/cuda_lu_decomposition.hpp
sjsprecious: Done.

sjsprecious:
@mattldawson @K20shores So I have replaced the pointers by references in this PR and guess what, now all the 39 tests pass on Derecho with the same NVHPC env.

I still have no clues why pointers or references matter here. I thought they were just different programming styles.

include/micm/solver/cuda_lu_decomposition.cuh
```diff
@@ -3,13 +3,27 @@
...
+    /// Note that referencing "devstruct" as "LuDecomposeConst&"
+    ///   will cause an error like `error: qualifiers dropped in 
+    ///   binding reference of type "LuDecomposeConst &" to 
+    ///   initializer of type "const LuDecomposeConst"`
     std::chrono::nanoseconds DecomposeKernelDriver(
             CudaSparseMatrixParam& sparseMatrix, 
-            CudaSolverParam& solver);
-  }  // namespace cuda
-}  // namespace micm
+            LuDecomposeConst devstruct);
```
sjsprecious, line 18: @mattldawson @K20shores I got a compilation error when using `LuDecomposeConst&` and the error message is listed in the comment. I switch to passing by value to avoid this error but let me know if there is a better way to do so. I also do not understand the error message here.

mattldawson:
> @mattldawson @K20shores So I have replaced the pointers by references in this PR and guess what, now all the 39 tests pass on Derecho with the same NVHPC env.
> 
> I still have no clues why pointers or references matter here. I thought they were just different programming styles.

That's interesting - maybe there was some condition where the pointer wasn't defined?

include/micm/solver/cuda_lu_decomposition.cuh
mattldawson: I don't see the error message - is it attached to another comment?

sjsprecious: Haha, it is not clear in the comment but here it is: `error: qualifiers dropped in binding reference of type "LuDecomposeConst &" to initializer of type "const LuDecomposeConst"`

mattldawson: ah, ok - I think you can change the function argument to be `const LuDecomposeConst&` and it should work (as long as you aren't modifying it in the function, which I don't think is the case)

sjsprecious:
> > @mattldawson @K20shores So I have replaced the pointers by references in this PR and guess what, now all the 39 tests pass on Derecho with the same NVHPC env.
> > I still have no clues why pointers or references matter here. I thought they were just different programming styles.
> 
> That's interesting - maybe there was some condition where the pointer wasn't defined?

Probably but I do not know. The failed test is for `AlphaMinusJacobian` and I do not think my revision touches that function. Maybe mixed usage of pointers and references in the same code confuse the compiler?

include/micm/solver/cuda_lu_decomposition.cuh
sjsprecious: so in this case, do I have to declare `const LuDecomposeConst devstruct_` in the `CudaLuDecomposition` class, too?

mattldawson:
> > > @mattldawson @K20shores So I have replaced the pointers by references in this PR and guess what, now all the 39 tests pass on Derecho with the same NVHPC env.
> > > I still have no clues why pointers or references matter here. I thought they were just different programming styles.
> > 
> > 
> > That's interesting - maybe there was some condition where the pointer wasn't defined?
> 
> Probably but I do not know. The failed test is for `AlphaMinusJacobian` and I do not think my revision touches that function. Maybe mixed usage of pointers and references in the same code confuse the compiler?

I'm not sure, but maybe we could do some memory checking on the CUDA tests in the future. Do you know if valgrind works with CUDA code, or if there is some equivalent tool?

include/micm/solver/cuda_lu_decomposition.cuh
mattldawson: no, if the function takes a `const` it will also work a non-`const` argument - it's essentially just a promise that the function won't try to modify the referenced argument, so it's safe to pass things that can't be modified

sjsprecious: Oh I see. Thanks for the clarification. I thought C++ must match the variable type exactly but  the keyword `const` seems to be an exception?

sjsprecious:
> > > > @mattldawson @K20shores So I have replaced the pointers by references in this PR and guess what, now all the 39 tests pass on Derecho with the same NVHPC env.
> > > > I still have no clues why pointers or references matter here. I thought they were just different programming styles.
> > > 
> > > 
> > > That's interesting - maybe there was some condition where the pointer wasn't defined?
> > 
> > 
> > Probably but I do not know. The failed test is for `AlphaMinusJacobian` and I do not think my revision touches that function. Maybe mixed usage of pointers and references in the same code confuse the compiler?
> 
> I'm not sure, but maybe we could do some memory checking on the CUDA tests in the future. Do you know if valgrind works with CUDA code, or if there is some equivalent tool?

I do not know but it is worth trying. I have used `valgrind` to do the memory check for TUV-x on Derecho. Regarding CUDA tool, `compute-sanitizer` should be the one to use. We can make a separate issue for this.

include/micm/solver/cuda_lu_decomposition.cuh
mattldawson: I think the difference is that `const` is a qualifier as opposed to being part of the type itself

src/solver/lu_decomposition.cu
mattldawson: A potential source of the intermittent failures with pointers and references could be that there is a default constructor that doesn't define `CudaLuDecomposition::devptr_` [here](https://github.com/NCAR/micm/blob/2686dc09f057eb32d9e28fdb96a9cc1cc4ded8a8/include/micm/solver/cuda_lu_decomposition.hpp#L22-L23) but `CudaLuDecomposition::Decompose` assumes that it is in a valid state [here](https://github.com/NCAR/micm/blob/2686dc09f057eb32d9e28fdb96a9cc1cc4ded8a8/include/micm/solver/cuda_lu_decomposition.hpp#L98). If the default constructor is needed, maybe there could be some check that `devptr_` is in a valid state?

sjsprecious: Oh, that could be the case. But do we really need the default constructor for MICM? Or it has to exist?

include/micm/solver/cuda_lu_decomposition.cuh
sjsprecious: I add the `const` qualifier to the `LuDecomposeConst` type and it works for the `DecomposeKernelDriver` function. However, if I do the same thing for the CUDA kernel `LuDecomposeConst`, the compilation passes but the gpu test of `cuda_lu_decomposition` will hang. Thus I still use passing by value for that kernel.

Is passing by reference a new feature in C++ standard? I see a compilation warning saying `nvcc warning : The -std=c++20 flag is not supported with the configured host compiler. Flag will be ignored.` and maybe this is the reason why CUDA kernel does not work with reference?

src/solver/lu_decomposition.cu
K20shores: @sjsprecious it doesn't necessarily have to exist. You can delete it or make it private and see what happens.

include/micm/solver/cuda_lu_decomposition.cuh
K20shores: @sjsprecious pass by value and pass by reference is not new in c++. To my knowledge it has been in the language from the beginning. The flag you are getting seems to suggest that yet again the configuration of the nividia compiler on the HPC machines is incorrect.

sjsprecious: According to this stackflow https://stackoverflow.com/questions/17794492/reference-argument-on-cuda), we need a pointer/copy to the device memory for the CUDA kernel argument. Thus the reference does not work for CUDA kernel in this case.

src/solver/lu_decomposition.cu
sjsprecious: Kyle: C/C++ prefers to return the value rather than passing the argument as value/pointer.

mattldawson (approved):

K20shores (approved):

Issue #390, opened 2024-01-04
Move device data management of `CudaProcessSet` data members to constructor/destructor


mattldawson:
Move device memory allocation, deallocation, and copying to `CudaProcessSet`.

Acceptance Criteria:

- device pointers for all index array data members of `CudaProcessSet` are created during construction
- all device pointer data members are freed during deconstruction

PR #403, opened 2024-01-15
Create constructor/destructor for CudaProcessSet class to handle data movement between host and device

sjsprecious:
This PR introduces a new constructor/destructor to the `CudaProcessSet` class that handles the data movement between host and device. It also cleans up some interfaces and adds more comments for future information.

```
Currently Loaded Modules:
```
All the 39 unit tests passed.

Fixed https://github.com/NCAR/micm/issues/390


src/process/process_set.cu
```diff
@@ -3,288 +3,223 @@
...
       auto endTime = std::chrono::high_resolution_clock::now();
       auto kernel_duration = std::chrono::duration_cast<std::chrono::nanoseconds>(endTime - startTime);
 
+      // copy the data from device to host
       cudaMemcpy(sparseMatrix.jacobian_, d_jacobian, sizeof(double) * sparseMatrix.jacobian_size_, cudaMemcpyDeviceToHost);
+
       // clean up
       cudaFree(d_rate_constants);
       cudaFree(d_state_variables);
       cudaFree(d_jacobian);
```
K20shores, line 192: Do these need to be allocated and freed each time? Couldn't we also allocate space for this at the beginning and just overwrite the data at the beginning of this kernel?

src/process/process_set.cu
```diff
@@ -296,13 +231,8 @@ namespace micm
       cudaFree(d_rate_constants);
       cudaFree(d_state_variables);
       cudaFree(d_forcing);
```
K20shores, line 233: Same question here as above

src/process/process_set.cu
sjsprecious: Agreed. It is more efficient to do what you suggested. However, there variables are either input variables for MICM or temporary matrix-type variables. Thus I think it should be handled in the `CudaMatrix` class or at the beginning of the MICM solver instead.

K20shores (approved):

include/micm/process/cuda_process_set.hpp
```diff
@@ -57,23 +84,13 @@ namespace micm
...
-
-    std::chrono::nanoseconds kernel_duration = micm::cuda::AddForcingTermsKernelDriver(matrix, processSet);
+    std::chrono::nanoseconds kernel_duration = micm::cuda::AddForcingTermsKernelDriver(matrix, this->devstruct_);
     return kernel_duration;  // time performance of kernel function
   }
 
   template<template<class> class MatrixPolicy, template<class> class SparseMatrixPolicy>
   requires VectorizableDense<MatrixPolicy<double>> && VectorizableSparse<SparseMatrixPolicy<double>>
-  inline std::chrono::nanoseconds CudaProcessSet::AddJacobianTerms(
+  inline std::chrono::nanoseconds CudaProcessSet::FormJacobianMatrix(
```
mattldawson, line 93: This function overrides `ProcessSet::AddJacobianTerms` [here](https://github.com/NCAR/micm/blob/be23ff66a96e72dd3dc945b83865ef42994bc5bf/include/micm/process/process_set.hpp#L69-L73), so it needs to have the same name

sjsprecious: Thanks @mattldawson . I have reverted the function name back to `AddJacobianTerms` and it makes more sense if we will consider aerosol chemistry in MICM in the future besides the gas-phase chemistry.

sjsprecious: And now the verification test for `cuda_process_set` fails. Let me debug more on this.

mattldawson (approved):
looks great!

sjsprecious:
@K20shores @mattldawson somehow the `cuda_process_set` unit test fails now for Jacobian matrix and I am debugging it. Please do not merge this PR until I find the problem. Thanks.

sjsprecious:
@K20shores @mattldawson After adding the `CopyJacobiFlatId` function that sets the `jacobian_flat_ids_` correctly on the device, all the tests now pass on Derecho.

Issue #394, opened 2024-01-04
Override `ForEach()` function for CUDA Matrix classes


mattldawson:
Acceptance Criteria:
- `ForEach()` function for `CudaMatrix` and `CudaSparseMatrix` executes on device
- @sjsprecious - should there be performance criteria as well?

sjsprecious:
According to the discussion with @mattldawson and @K20shores , instead of using the `foreach` function, we could also use some standard libraries like KPP does (https://github.com/KineticPreProcessor/KPP/blob/main/int/rosenbrock.c#L538-L570). There should be both CPU and GPU versions available.

Need to evaluate its impact on the performance in some ways.

sjsprecious:
The CUDA version of `Foreach` is introduced into `CudaVectorMatrix` class through https://github.com/NCAR/micm/pull/422.

sjsprecious:
I think we still need this issue open for the `CudaSparseMatrix` class unless it could inherit the same function from `CudaVectorMatrix` class. Am I wrong here? @mattldawson

mattldawson:
> I think we still need this issue open for the `CudaSparseMatrix` class unless it could inherit the same function from `CudaVectorMatrix` class. Am I wrong here? @mattldawson

ah! sorry, I'll reopen it

sjsprecious:
After discussing with @mattldawson , the `Foreach` function may not be used for the `SparseMatrix` or `CudaSparseMatrix` class. Close this issue for now.

PR #408, opened 2024-01-30
Initial CudaVectorMatrix implementation.

### LINKED ISSUE #392: Create `CudaVectorMatrix` class
Extend the `VectorMatrix` class for use with CUDA solver functions.

Acceptance Criteria:

- `CudaVectorMatrix` class should allocate/deallocate device memory for matrix data on construction/destruction
- Matrix data is readable/writable from host and device

**Ideas**
- It might be easiest to add something like `SyncHost` and `SyncDevice` functions that would explicitly copy matrix data from device-to-host and host-to-device

mwaxmonsky:
Adds new `CudaVectorMatrix` class which extends the base `VectorMatrix` class.

A few caveats are:
1) The class can only be used with doubles when trying to move memory to the GPU and vice-versa and throws a build error if using non-`double`s.
2) Currently does no work on GPU for `ForEach`.  This will come in with #394.
3) Some of the tests from `VectorMatrix` had to be copied into the `test_cuda_vector_matrix` file due to the base tests using `int`s instead of `double`s.

Closes #392

### DIFF
--- a/include/micm/util/cuda_vector_matrix.cuh
+++ b/include/micm/util/cuda_vector_matrix.cuh
@@ -1,0 +1,15 @@
+#include <vector>
+
+namespace micm
+{
+  namespace cuda
+  {
+    int malloc_vector(double *d_data, std::size_t num_elements);
+
+    int free_vector(double *d_data);
+
+    int copy_to_device(double *d_data, const double *h_data, std::size_t num_elements);
+
+    int copy_to_host(double *d_data, double *h_data, std::size_t num_elements);
+  }
+}
--- a/include/micm/util/cuda_vector_matrix.hpp
+++ b/include/micm/util/cuda_vector_matrix.hpp
@@ -1,0 +1,65 @@
+#include <type_traits>
+#include <micm/util/vector_matrix.hpp>
+#include <micm/util/cuda_vector_matrix.cuh>
+
+namespace micm {
+
+    template<class T, std::size_t L = DEFAULT_VECTOR_SIZE>
+    class CudaVectorMatrix : public VectorMatrix<T, L> {
+    private:
+        double* d_data_;
+
+    public:
+        CudaVectorMatrix() requires(std::is_same_v<T, double>)
+         : VectorMatrix<T,L>()
+        {
+          micm::cuda::malloc_vector(d_data_, static_cast<unsigned int>(0));
+        }
+        CudaVectorMatrix()
+         : VectorMatrix<T,L>()
+        {}
+
+        CudaVectorMatrix(std::size_t x_dim, std::size_t y_dim) requires(std::is_same_v<T, double>)
+          : VectorMatrix<T,L>(x_dim, y_dim)
+        {
+          micm::cuda::malloc_vector(d_data_, this->data_.size());
+        }
+        CudaVectorMatrix(std::size_t x_dim, std::size_t y_dim)
+          : VectorMatrix<T,L>(x_dim, y_dim)
+        {}
+
+        CudaVectorMatrix(std::size_t x_dim, std::size_t y_dim, T initial_value) requires(std::is_same_v<T, double>)
+          : VectorMatrix<T,L>(x_dim, y_dim, initial_value)
+        {
+          micm::cuda::malloc_vector(d_data_, this->data_.size());
+        }
+        CudaVectorMatrix(std::size_t x_dim, std::size_t y_dim, T initial_value)
+          : VectorMatrix<T,L>(x_dim, y_dim, initial_value)
+        {}
+
+        CudaVectorMatrix(const std::vector<std::vector<T>> other) requires(std::is_same_v<T, double>)
+          : VectorMatrix<T,L>(other)
+        {
+          micm::cuda::malloc_vector(d_data_, this->data_.size());
+        }
+        CudaVectorMatrix(const std::vector<std::vector<T>> other)
+          : VectorMatrix<T,L>(other)
+        {}
+
+        ~CudaVectorMatrix()
+        {
+          micm::cuda::free_vector(d_data_);
+        }
+
+        int CopyToDevice()
+        {
+          static_assert(std::is_same_v<T, double>);
+          return micm::cuda::copy_to_device(d_data_, this->data_.data(), this->AsVector().size());
+        }
+        int GetFromDevice()
+        {
+          static_assert(std::is_same_v<T, double>);
+          return micm::cuda::copy_to_host(d_data_, this->data_.data(), this->AsVector().size());
+        }
+    };
+}
--- a/src/util/cuda_vector_matrix.cu
+++ b/src/util/cuda_vector_matrix.cu
@@ -1,0 +1,25 @@
+#include <vector>
+//#include <micm/util/cuda_vector_matrix.cuh>
+
+namespace micm
+{
+  namespace cuda
+  {
+    int malloc_vector(double *d_data, std::size_t num_elements)
+    {
+      return cudaMalloc(&d_data, sizeof(double) * num_elements);
+    }
+    int free_vector(double *d_data)
+    {
+      return cudaFree(d_data);
+    }
+    int copy_to_device(double *d_data, const double *h_data, std::size_t num_elements)
+    {
+      return cudaMemcpy(d_data, h_data, sizeof(double) * num_elements, cudaMemcpyHostToDevice);
+    }
+    int copy_to_host(double* d_data, double *h_data, std::size_t num_elements)
+    {
+      return cudaMemcpy(h_data, d_data, sizeof(double) * num_elements, cudaMemcpyDeviceToHost);
+    }
+  }
+}
--- a/test/unit/util/test_cuda_vector_matrix.cpp
+++ b/test/unit/util/test_cuda_vector_matrix.cpp
@@ -1,0 +1,187 @@
+#include <gtest/gtest.h>
+
+#include <micm/util/cuda_vector_matrix.hpp>
+
+#include "test_matrix_policy.hpp"
+
+template<class T>
+using Group1MatrixAlias = micm::CudaVectorMatrix<T,1>;
+template<class T>
+using Group2MatrixAlias = micm::CudaVectorMatrix<T,2>;
+template<class T>
+using Group3MatrixAlias = micm::CudaVectorMatrix<T,3>;
+template<class T>
+using Group4MatrixAlias = micm::CudaVectorMatrix<T,4>;
+
+
+
+TEST(VectorMatrix, SmallVectorMatrix)
+{
+  auto matrix = testSmallMatrix<Group2MatrixAlias>();
+
+  matrix.CopyToDevice();
+  matrix.GetFromDevice();
+
+  EXPECT_EQ(matrix[1][3], 64.7);
+  EXPECT_EQ(matrix[0][0], 41.2);
+  EXPECT_EQ(matrix[2][4], 102.3);
+
+  std::vector<double>& data = matrix.AsVector();
+
+  EXPECT_EQ(data.size(), 4 * 5);
+  EXPECT_EQ(matrix.GroupSize(), 2 * 5);
+  EXPECT_EQ(matrix.NumberOfGroups(), 2);
+  EXPECT_EQ(matrix.GroupVectorSize(), 2);
+  EXPECT_EQ(data[0], 41.2);
+  EXPECT_EQ(data[2 * 5 + 0 + 2 * 4], 102.3);
+  EXPECT_EQ(data[1 + 2 * 3], 64.7);
+}
+
+TEST(VectorMatrix, SmallConstVectorMatrix)
+{
+  auto matrix = testSmallConstMatrix<Group4MatrixAlias>();
+
+  const std::vector<double>& data = matrix.AsVector();
+
+  EXPECT_EQ(data.size(), 4 * 5);
+  EXPECT_EQ(matrix.GroupSize(), 4 * 5);
+  EXPECT_EQ(matrix.NumberOfGroups(), 1);
+  EXPECT_EQ(matrix.GroupVectorSize(), 4);
+  EXPECT_EQ(data[0], 41.2);
+  EXPECT_EQ(data[2 + 4 * 4], 102.3);
+  EXPECT_EQ(data[1 + 4 * 3], 64.7);
+}
+
+TEST(VectorMatrix, InitializeVectorMatrix)
+{
+  auto matrix = testInializeMatrix<Group1MatrixAlias>();
+  matrix.CopyToDevice();
+  matrix.GetFromDevice();
+
+  EXPECT_EQ(matrix[0][0], 12.4);
+  EXPECT_EQ(matrix[1][0], 12.4);
+  EXPECT_EQ(matrix[1][2], 12.4);
+}
+
+TEST(VectorMatrix, InitializeConstVectorMatrix)
+{
+  auto matrix = testInializeConstMatrix<Group2MatrixAlias>();
+  matrix.CopyToDevice();
+  matrix.GetFromDevice();
+
+  EXPECT_EQ(matrix[0][0], 12.4);
+  EXPECT_EQ(matrix[1][0], 12.4);
+  EXPECT_EQ(matrix[1][2], 12.4);
+}
+
+TEST(VectorMatrix, LoopOverVectorMatrix)
+{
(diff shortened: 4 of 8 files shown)


include/micm/util/cuda_vector_matrix.hpp
```diff
@@ -0,0 +1,65 @@
...
+          micm::cuda::malloc_vector(d_data_, this->data_.size());
+        }
+        CudaVectorMatrix(std::size_t x_dim, std::size_t y_dim)
+          : VectorMatrix<T,L>(x_dim, y_dim)
+        {}
+
+        CudaVectorMatrix(std::size_t x_dim, std::size_t y_dim, T initial_value) requires(std::is_same_v<T, double>)
+          : VectorMatrix<T,L>(x_dim, y_dim, initial_value)
+        {
+          micm::cuda::malloc_vector(d_data_, this->data_.size());
```
sjsprecious, line 34: should `initial_value` be passed as an argument to the CUDA kernel?

include/micm/util/cuda_vector_matrix.hpp
```diff
@@ -0,0 +1,65 @@
...
+        CudaVectorMatrix(std::size_t x_dim, std::size_t y_dim) requires(std::is_same_v<T, double>)
+          : VectorMatrix<T,L>(x_dim, y_dim)
+        {
+          micm::cuda::malloc_vector(d_data_, this->data_.size());
+        }
+        CudaVectorMatrix(std::size_t x_dim, std::size_t y_dim)
+          : VectorMatrix<T,L>(x_dim, y_dim)
+        {}
+
+        CudaVectorMatrix(std::size_t x_dim, std::size_t y_dim, T initial_value) requires(std::is_same_v<T, double>)
```
sjsprecious, line 31: What is the difference between the constructors at line 27 and 31?

test/unit/util/test_cuda_vector_matrix.cpp
```diff
@@ -0,0 +1,187 @@
...
+  const std::vector<double>& data = matrix.AsVector();
+
+  EXPECT_EQ(data.size(), 4 * 5);
+  EXPECT_EQ(matrix.GroupSize(), 4 * 5);
+  EXPECT_EQ(matrix.NumberOfGroups(), 1);
+  EXPECT_EQ(matrix.GroupVectorSize(), 4);
+  EXPECT_EQ(data[0], 41.2);
+  EXPECT_EQ(data[2 + 4 * 4], 102.3);
+  EXPECT_EQ(data[1 + 4 * 3], 64.7);
+}
```
sjsprecious, line 53: What is this test for?

sjsprecious (commented):
Thanks @mwaxmonsky for working on this PR. It looks great and I just have a few clarification questions.

include/micm/util/cuda_vector_matrix.hpp
mwaxmonsky: I believe we are going with the design choice to make any memory movement to the device an explicit call by the user.  So in this case, the user would call `auto matrix = CudaVectorMatrix<...>(...)` and then `matrix.CopyToDevice()` manually.  Which matches what the process will be when we start calling `ForEach(...)` as well.

test/unit/util/test_cuda_vector_matrix.cpp
mwaxmonsky: Ah, good catch!  Glossed over this one initially.  There's supposed to be copies to and from the device and verifying the contents afterwards. similar to the test above.

src/util/cuda_vector_matrix.cu
```diff
@@ -0,0 +1,25 @@
+#include <vector>
+//#include <micm/util/cuda_vector_matrix.cuh>
```
mattldawson, line 2: please delete commented out code

test/unit/util/test_cuda_vector_matrix.cpp
```diff
@@ -0,0 +1,194 @@
...
+  EXPECT_EQ(data.size(), 4 * 5);
+  EXPECT_EQ(matrix.GroupSize(), 2 * 5);
+  EXPECT_EQ(matrix.NumberOfGroups(), 2);
+  EXPECT_EQ(matrix.GroupVectorSize(), 2);
+  EXPECT_EQ(data[0], 41.2);
+  EXPECT_EQ(data[2 * 5 + 0 + 2 * 4], 102.3);
+  EXPECT_EQ(data[1 + 2 * 3], 64.7);
+}
+
+TEST(VectorMatrix, SmallConstVectorMatrix)
```
mattldawson, line 40:
```suggestion
TEST(CudaVectorMatrix, SmallConstVectorMatrix)
```

test/unit/util/test_cuda_vector_matrix.cpp
```diff
@@ -0,0 +1,194 @@
...
+  EXPECT_EQ(data.size(), 4 * 5);
+  EXPECT_EQ(matrix.GroupSize(), 4 * 5);
+  EXPECT_EQ(matrix.NumberOfGroups(), 1);
+  EXPECT_EQ(matrix.GroupVectorSize(), 4);
+  EXPECT_EQ(data[0], 41.2);
+  EXPECT_EQ(data[2 + 4 * 4], 102.3);
+  EXPECT_EQ(data[1 + 4 * 3], 64.7);
+}
+
+TEST(VectorMatrix, InitializeVectorMatrix)
```
mattldawson, line 62:
```suggestion
TEST(CudaVectorMatrix, InitializeVectorMatrix)
```

test/unit/util/test_cuda_vector_matrix.cpp
```diff
@@ -0,0 +1,194 @@
...
+  auto matrix = testInializeMatrix<Group1MatrixAlias>();
+  matrix.CopyToDevice();
+  matrix.GetFromDevice();
+
+  EXPECT_EQ(matrix[0][0], 12.4);
+  EXPECT_EQ(matrix[1][0], 12.4);
+  EXPECT_EQ(matrix[1][2], 12.4);
+}
+
+TEST(VectorMatrix, InitializeConstVectorMatrix)
```
mattldawson, line 73:
```suggestion
TEST(CudaVectorMatrix, InitializeConstVectorMatrix)
```

test/unit/util/test_cuda_vector_matrix.cpp
```diff
@@ -0,0 +1,194 @@
...
+  auto matrix = testInializeConstMatrix<Group2MatrixAlias>();
+  matrix.CopyToDevice();
+  matrix.GetFromDevice();
+
+  EXPECT_EQ(matrix[0][0], 12.4);
+  EXPECT_EQ(matrix[1][0], 12.4);
+  EXPECT_EQ(matrix[1][2], 12.4);
+}
+
+TEST(VectorMatrix, LoopOverVectorMatrix)
```
mattldawson, line 84:
```suggestion
TEST(CudaVectorMatrix, LoopOverVectorMatrix)
```

test/unit/util/test_cuda_vector_matrix.cpp
```diff
@@ -0,0 +1,194 @@
...
+  matrix.CopyToDevice();
+  matrix.GetFromDevice();
+
+  EXPECT_EQ(matrix[0][0], 0);
+  EXPECT_EQ(matrix[1][2], 102);
+  EXPECT_EQ(matrix[2][3], 203);
+  EXPECT_EQ(matrix[0][3], 3);
+}
+
+TEST(VectorMatrix, LoopOverConstVectorMatrix)
```
mattldawson, line 109:
```suggestion
TEST(CudaVectorMatrix, LoopOverConstVectorMatrix)
```

test/unit/util/test_cuda_vector_matrix.cpp
```diff
@@ -0,0 +1,194 @@
...
+  matrix.CopyToDevice();
+  matrix.GetFromDevice();
+
+  EXPECT_EQ(matrix[0][0], 0);
+  EXPECT_EQ(matrix[1][2], 102);
+  EXPECT_EQ(matrix[2][3], 203);
+  EXPECT_EQ(matrix[0][3], 3);
+}
+
+TEST(VectorMatrix, ConversionToVector)
```
mattldawson, line 136:
```suggestion
TEST(CudaVectorMatrix, ConversionToVector)
```

test/unit/util/test_cuda_vector_matrix.cpp
```diff
@@ -0,0 +1,194 @@
...
+  matrix.GetFromDevice();
+
+  auto slice = matrix[1];
+
+  EXPECT_EQ(slice[0], 13.2);
+  EXPECT_EQ(slice[1], 31.2);
+  EXPECT_EQ(slice[2], 314.2);
+}
+
+TEST(VectorMatrix, ConstConversionToVector)
```
mattldawson, line 149:
```suggestion
TEST(CudaVectorMatrix, ConstConversionToVector)
```

test/unit/util/test_cuda_vector_matrix.cpp
```diff
@@ -0,0 +1,194 @@
...
+  matrix.GetFromDevice();
+
+  auto slice = matrix[1];
+
+  EXPECT_EQ(slice[0], 13.2);
+  EXPECT_EQ(slice[1], 31.2);
+  EXPECT_EQ(slice[2], 314.2);
+}
+
+TEST(VectorMatrix, ConversionFromVector)
```
mattldawson, line 162:
```suggestion
TEST(CudaVectorMatrix, ConversionFromVector)
```

test/unit/util/test_cuda_vector_matrix.cpp
```diff
@@ -0,0 +1,194 @@
...
+  EXPECT_EQ(matrix[0][0], 412.3);
+  EXPECT_EQ(matrix[0][1], 32.4);
+  EXPECT_EQ(matrix[0][2], 41.3);
+  EXPECT_EQ(matrix[1].size(), 3);
+  EXPECT_EQ(matrix[1][0], 5.33);
+  EXPECT_EQ(matrix[1][1], -0.3);
+  EXPECT_EQ(matrix[1][2], 31.2);
+}
+
+TEST(VectorMatrix, AssignmentFromVector)
```
mattldawson, line 183:
```suggestion
TEST(CudaVectorMatrix, AssignmentFromVector)
```

test/unit/util/test_cuda_vector_matrix.cpp
```diff
@@ -0,0 +1,194 @@
...
+using Group4MatrixAlias = micm::CudaVectorMatrix<T,4>;
+
+
+
+TEST(VectorMatrix, SmallVectorMatrix)
+{
+  auto matrix = testSmallMatrix<Group2MatrixAlias>();
+
+  matrix.CopyToDevice();
+  matrix.GetFromDevice();
```
mattldawson, line 23: Would it be worth adding a kernel function just for tests that modified the data on the device before copying it back to the host? As these tests are currently set up, `CopyToDevice()` and `CopyFromDevice()` could do nothing and it seems like the tests would pass.

include/micm/util/cuda_vector_matrix.hpp
```diff
@@ -0,0 +1,65 @@
+#include <type_traits>
+#include <micm/util/vector_matrix.hpp>
+#include <micm/util/cuda_vector_matrix.cuh>
+
+namespace micm {
+
+    template<class T, std::size_t L = DEFAULT_VECTOR_SIZE>
+    class CudaVectorMatrix : public VectorMatrix<T, L> {
```
mattldawson, line 8: I suggest adding a description in doxygen format for at least the class as a whole, describing what it's purpose is, how it allows transfer to/from device memory, etc.

include/micm/util/cuda_vector_matrix.cuh
```diff
@@ -0,0 +1,15 @@
...
+{
+  namespace cuda
+  {
+    int MallocVector(double *d_data, std::size_t num_elements);
+
+    int FreeVector(double *d_data);
+
+    int CopyToDevice(double *d_data, const double *h_data, std::size_t num_elements);
+
+    int CopyToHost(double *d_data, double *h_data, std::size_t num_elements);
```
mattldawson, line 13: maybe add a short description for each of these function in doxygen format? it's pretty clear from the names what they do, and we haven't been very consistent with adding these comments, but it would be nice to move in this direction

mattldawson (changes requested):
Looks really nice! I think this is going to end up working very well in the solver functions. Just a couple very minor suggestions

test/unit/util/test_cuda_vector_matrix.cpp
mwaxmonsky: I was thinking about that as well and am definitely interested in adding that functionality in so we can test this.  I wasn't sure how to best add it without creating a dummy function in the cuda files though.  Not sure the current `cuda_vector_matrix` class is the right place but more than open to ideas on how to best go about it.

test/unit/util/test_cuda_vector_matrix.cpp
mwaxmonsky: Sorry about that!  Updated!

test/unit/util/test_cuda_vector_matrix.cpp
mwaxmonsky: Updated

test/unit/util/test_cuda_vector_matrix.cpp
mwaxmonsky: Updated!

test/unit/util/test_cuda_vector_matrix.cpp
mwaxmonsky: Updated!

test/unit/util/test_cuda_vector_matrix.cpp
mwaxmonsky: Updated!

test/unit/util/test_cuda_vector_matrix.cpp
mwaxmonsky: Updated!

test/unit/util/test_cuda_vector_matrix.cpp
mwaxmonsky: Updated!

test/unit/util/test_cuda_vector_matrix.cpp
mwaxmonsky: Updated!

test/unit/util/test_cuda_vector_matrix.cpp
mwaxmonsky: Updated!

test/unit/util/test_cuda_vector_matrix.cpp
mattldawson: Maybe adding it to the test folder? something like:
```
test/unit/util/cuda/cuda_matrix_utils.cu
test/unit/util/cuda/cuda_matrix_utils.cuh
```
and you could just create a wrapper function that takes a `CudaMatrixParam&` and passes the device pointer (which points to an array of doubles) to a kernel function that then adds some specific values to certain elements? (I think it's ok for this function to assume some size for the array, since it's just for the tests).

src/util/cuda_vector_matrix.cu
mwaxmonsky: Good catch!  Forgot to re-add this but my intention was to force the `.cu` implementation to follow the signature in the header file that way it is shared between the `.cu` file and the `.cpp` files that include it.  If this isn't the case, I can remove this.

test/unit/util/test_cuda_vector_matrix.cpp
mwaxmonsky: Added in the test that we discussed!

include/micm/util/cuda_vector_matrix.hpp
mwaxmonsky: Added documentation, just let me know if it could be clearer or needs some formatting updates.

include/micm/util/cuda_vector_matrix.cuh
mwaxmonsky: Documentation added!

sjsprecious (approved):
Thanks @mwaxmonsky for working on the revision. It looks great.

I just ran the tests on Derecho's A100 GPU and all the 40 tests passed as expected.

mattldawson (approved):
looks great!

PR #410, opened 2024-02-12
reseting matrices

K20shores:
Resets the lower and upper matrices to zero at the beginning of the solve function


mattldawson (approved):
nice!

PR #413, opened 2024-02-13
prefixing all options with micm_

### LINKED ISSUE #412: Prefix all cmake build options with MICM_
To prevent conflicts with other packages and make it easy to identify options specific to micm, prefix all options with `MICM_`

**Acceptance criteria**

- All options are prefixed with `MICM_` or something similar.
- All tests and documentation use the new options

K20shores:
To prevent name conflicts with other packages and to make it explicit when using micm from other packages, all micm cmake options are prefixed with `MICM_`

closes #412


src/CMakeLists.txt
```diff
@@ -9,49 +9,51 @@ add_library(musica::micm ALIAS micm)
...
-    message(FATAL_ERROR "GPU_TYPE is not set or is empty. Please provide a GPU type.")
+if(MICM_ENABLE_OPENACC OR MICM_ENABLE_CUDA)
+  if(NOT MICM_GPU_TYPE)
+    message(FATAL_ERROR "MICM_GPU_TYPE is not set or is empty. Please provide a GPU type.")
   endif()
 
   set(GPU_ARCH "")
-  if (GPU_TYPE STREQUAL a100)
+  if (MICM_GPU_TYPE STREQUAL a100)
     set(GPU_ARCH "80")
```
mwaxmonsky, line 44: We should also set `GPU_ARCH` to `MICM_GPU_ARCH` and then I can re-arrange how it's used when I clean up the cuda project cmake to use the newer `project(...)` syntax.

mwaxmonsky (changes requested):
Great work @K20shores!  Just one little update and should be good to go!  Thanks!

sjsprecious (approved):
Thanks @K20shores . I tested the new options for a GPU build on Derecho and all the tests passed.

mattldawson (approved):

src/CMakeLists.txt
mattldawson: If this would break the current scripts, I would suggest holding off until you submit the PR with the cleaned-up CUDA build

K20shores: @mwaxmonsky done

mwaxmonsky (approved):
Looks great!

PR #420, opened 2024-02-22
416 add assignment overridescopy constructors to cudavectormatrix so new matrix has separate device memory

mwaxmonsky:
Adds assignment and move operators and constructors so a CudaVectorMatrix can be assigned to another  which generates new device data for the newly assigned vector or moves the other vector into the new vector while in that case taking ownership of the other matrix's device memory.

### DIFF
--- a/include/micm/util/cuda_vector_matrix.cuh
+++ b/include/micm/util/cuda_vector_matrix.cuh
@@ -27,5 +27,11 @@ namespace micm
     /// @param h_data Host data to copy data to
     /// @returns Error code from copying from the device to the host, if any
     int CopyToHost(CudaVectorMatrixParam& vectorMatrix, std::vector<double>& h_data);
+
+    /// @brief Copies data to the destination device memory block from the source device memory block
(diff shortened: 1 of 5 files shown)


sjsprecious (approved):
Thanks @mwaxmonsky for fixing this issue. All the tests passed on Derecho's GPU.

Would it be helpful to add some comments about the overrided operator?

mattldawson (approved):
nice!

mwaxmonsky:
> Thanks @mwaxmonsky for fixing this issue. All the tests passed on Derecho's GPU.
> 
> Would it be helpful to add some comments about the overrided operator?

Is this in regards to the 4 move/copy assignment/constructors?  If so, after talking with @mattldawson, it sounds like we don't want to add comments stating what each operator/constructor is since they are inferred by their signature.

But if this is in relation to different comments, just let me know and I'd be happy to add it!

sjsprecious (approved):
All the tests still passed on Derecho's GPU.

PR #422, opened 2024-02-24
Add cublasDaxpy routine to perform foreach-like calculation on the GPU

sjsprecious:
This PR introduces the `cublasDaxpy` routine in the `CudaVectorMatrix` class to perform the `foreach` function on the GPU.

It should be added to the `CudaSparseMatrix` class (once it is generated) as well.

```
Currently Loaded Modules:
```
all the unit tests passed on Derecho's GPU.


src/util/cuda_vector_matrix.cu
```diff
@@ -16,10 +17,13 @@ namespace micm
...
     int CopyToDevice(CudaVectorMatrixParam& param, std::vector<double>& h_data)
     {
-      return cudaMemcpy(param.d_data_, h_data.data(), sizeof(double) * param.num_elements_, cudaMemcpyHostToDevice);
+      cudaError_t err = cudaMemcpy(param.d_data_, h_data.data(), sizeof(double) * param.num_elements_, cudaMemcpyHostToDevice);
+      cudaDeviceSynchronize();
+      return err;
     }
     int CopyToHost(CudaVectorMatrixParam& param, std::vector<double>& h_data)
     {
+      cudaDeviceSynchronize();
```
mwaxmonsky, line 26: Is there a need for explicit device synchronization on memory copies?  I'm not seeing asynchronous behavior so this may be an issue of PEBKAC on my part.  😄

test/unit/util/test_cuda_vector_matrix.cpp
```diff
@@ -446,3 +447,26 @@ TEST(CudaVectorMatrix, AssignmentFromVector)
...
+  gpu_y.Axpy(alpha, gpu_x, 1, 1);
+  gpu_y.CopyToHost();
+
+  EXPECT_EQ(gpu_y[0][0], 40.0);
+  EXPECT_EQ(gpu_y[0][1], 60.0);
+  EXPECT_EQ(gpu_y[1][1], 80.0);
+
+  double sum = std::accumulate(gpu_y.AsVector().begin(), gpu_y.AsVector().end(), 0);
+  EXPECT_EQ(sum, (10*2.0+20.0)*198 + 20.0*2.0+20.0 + 30.0*2.0+20.0);
+}
```
mwaxmonsky, line 472:
```suggestion
}

```
Missing newline at end of file.

mwaxmonsky (changes requested):
Great work @sjsprecious!  Just one small change and a clarification needed but otherwise looks great!

include/micm/util/cuda_vector_matrix.hpp
```diff
@@ -127,5 +139,24 @@ namespace micm
     {
       return CudaVectorMatrixParam { vector_matrix_param_.d_data_, vector_matrix_param_.num_elements_ };
     }
+    /// @brief Create the CUDA version of "foreach" function (i.e., for each element 
+    ///        in the VectorMatrix x and y, performing y = aphla * x + y)
```
mattldawson, line 143:
```suggestion
    /// @brief For each element in the VectorMatrix x and y, perform y = alpha * x + y,
    /// where alpha is a scalar constant.
```

mattldawson (changes requested):
Very nice! Just a minor suggestion on the function description. Also, it looks like the NVidia build is failing in the GH Actions. Is it possible we need to use a newer (or different) version of the compilers to use the `cublas` library?

src/util/cuda_vector_matrix.cu
sjsprecious: You are correct that there is no need to do device synchronization for `CopyToDevice` so far unless we do asynchronous copy. I thought we might need it in the future and just added it here. But we may never need it and it seems to confuse people here. Just remove it for clarity.

test/unit/util/test_cuda_vector_matrix.cpp
sjsprecious: How could you catch this!

sjsprecious:
> Very nice! Just a minor suggestion on the function description. Also, it looks like the NVidia build is failing in the GH Actions. Is it possible we need to use a newer (or different) version of the compilers to use the `cublas` library?

The `cublas` library comes from the CUDA library and the `cublas` interface does change in the recent version. How could I check which CUDA version is used in the GH actions?

mattldawson:
> > Very nice! Just a minor suggestion on the function description. Also, it looks like the NVidia build is failing in the GH Actions. Is it possible we need to use a newer (or different) version of the compilers to use the `cublas` library?
> 
> The `cublas` library comes from the CUDA library and the `cublas` interface does change in the recent version. How could I check which CUDA version is used in the GH actions?

it might be in the log files [here](https://github.com/NCAR/micm/actions/runs/8053171734/job/21994929398?pr=422)

I see this line:
```
#3 [internal] load metadata for nvcr.io/nvidia/nvhpc:23.5-devel-cuda12.1-ubuntu22.04
#3 DONE 1.3s
```

sjsprecious:
> > > Very nice! Just a minor suggestion on the function description. Also, it looks like the NVidia build is failing in the GH Actions. Is it possible we need to use a newer (or different) version of the compilers to use the `cublas` library?
> > 
> > 
> > The `cublas` library comes from the CUDA library and the `cublas` interface does change in the recent version. How could I check which CUDA version is used in the GH actions?
> 
> it might be in the log files [here](https://github.com/NCAR/micm/actions/runs/8053171734/job/21994929398?pr=422)
> 
> I see this line:
> 
> ```
> #3 [internal] load metadata for nvcr.io/nvidia/nvhpc:23.5-devel-cuda12.1-ubuntu22.04
> #3 DONE 1.3s
> ```

Thanks @mattldawson . I saw the error message came from:
```
#9 151.5 "/opt/nvidia/hpc_sdk/Linux_x86_64/23.5/math_libs/include/cublas_api.h", line 74: catastrophic error: cannot open source file "driver_types.h"
#9 151.5   #include "driver_types.h"
#9 151.5                            ^
#9 151.5 
#9 151.5 1 catastrophic error detected in the compilation of "/micm/test/unit/util/test_cuda_vector_matrix.cpp".
#9 151.5 Compilation terminated.
#9 151.5 make[2]: *** [test/unit/util/CMakeFiles/test_cuda_vector_matrix.dir/build.make:76: test/unit/util/CMakeFiles/test_cuda_vector_matrix.dir/test_cuda_vector_matrix.cpp.o] Error 2
#9 151.5 make[1]: *** [CMakeFiles/Makefile2:1693: test/unit/util/CMakeFiles/test_cuda_vector_matrix.dir/all] Error 2
#9 151.5 make[1]: *** Waiting for unfinished jobs....
```
and it could be caused by the nvhpc compiler version used in the GH action, which is `23.5`. Can we use `nvhpc/23.7` and `cuda/12.2.1` instead, which are the versions used on Derecho?

test/unit/util/test_cuda_vector_matrix.cpp
mwaxmonsky: 😃 GitHub has an annoying red X on the end of files in the Files Changed view.  I've caught myself so many times post commit and having to push up a bunch of newline fixes.

Thanks for fixing!

sjsprecious:
Hmm, I updated the nvhpc container with `nvhpc/23.7` and `cuda/12.2`, but the GH action still failed with the same complain about the `driver_types.h` file.

I could confirm that the `driver_types.h` file is available under the path `NCAR_INC_CUDA` on Derecho. Thus I guess this file is somehow missing from the nvhpc container image? Any thoughts here? @mattldawson @K20shores

mattldawson:
> Hmm, I updated the nvhpc container with `nvhpc/23.7` and `cuda/12.2`, but the GH action still failed with the same complain about the `driver_types.h` file.
> 
> I could confirm that the `driver_types.h` file is available under the path `NCAR_INC_CUDA` on Derecho. Thus I guess this file is somehow missing from the nvhpc container image? Any thoughts here? @mattldawson @K20shores

It could be missing, or it could be not correctly included in the include path. I'm happy to help track this down if you'd like.

sjsprecious:
> > Hmm, I updated the nvhpc container with `nvhpc/23.7` and `cuda/12.2`, but the GH action still failed with the same complain about the `driver_types.h` file.
> > I could confirm that the `driver_types.h` file is available under the path `NCAR_INC_CUDA` on Derecho. Thus I guess this file is somehow missing from the nvhpc container image? Any thoughts here? @mattldawson @K20shores
> 
> It could be missing, or it could be not correctly included in the include path. I'm happy to help track this down if you'd like.

Thanks @mattldawson . That will be great!

mwaxmonsky (approved):
Looks great @sjsprecious and @mattldawson!

mattldawson (approved):

PR #424, opened 2024-03-07
Override NormalizedError function with CUDA

### LINKED ISSUE #395: Override `NoramlizeError()` function in `CudaRosenbrockSolver`

sjsprecious:
This PR introduces a CUDA version of `NormalizedError` function, which is borrowed from NVIDIA's official documentation (https://developer.download.nvidia.com/assets/cuda/files/reduction.pdf) but modified for efficiency and robustness.

Suggested by Michael, I also tried the `reduce` function from the `Thrust` library but it is not as performant as the CUDA implementation (~4-5x slower).

The unit tests include two main parts:

- Test arrays with different lengths but the same element values. In this case, the RMSE should not change no matter what length of arrays is.
- Test arrays with different lengths and different element values. In this case, the RMSE will change depending on the length of arrays.

Fix #395


src/solver/rosenbrock.cu
```diff
@@ -61,5 +140,51 @@ namespace micm
       return kernel_duration;
     }
 
+    double NormalizedErrorDriver(double* d_y_old, double* d_y_new, 
+                                 double* d_errors, const size_t num_elements,
```
mattldawson, line 144: For consistency with `AlphaMinusJacobianDriver()`, should you pass the `CudaMatrixParam` references to this driver function, instead of the raw pointers?

mattldawson (approved):
Looks great! Just one minor question that may not need any changes

src/solver/rosenbrock.cu
sjsprecious: Thanks Matt. Based on my understanding, once the `CudaVectorMatrix` and `CudaSparseVectorMatrix` class is introduced into MICM, we should not need the `CudaMatrixParam` and `CudaSparseMatrixParam` struct any more (I will revise all those functions including `AlphaMinusJacobianDriver()` later). I do not mind passing raw pointers to the CUDA kernels. We could surely group them into a struct and make the interface clean. But I think we then need to do `cudamalloc` again for that struct (or its members) and that is something we don't want (we want all the `cudamalloc` happens in the constructors if possible). Let me know if I have any misunderstandings here.

sjsprecious:
Huge thanks to @mwaxmonsky for directing me to the cublas APIs. It turns out when `num_elements` > 1M, cublas is more efficient than the cuda implementation. I have hard-coded a number with the `if` statement in the cuda code ,but I can also make it a tunable parameter somewhere.

src/solver/rosenbrock.cu
sjsprecious: Thanks @mattldawson for your comments. I have updated the function/kernel interface as suggested.

include/micm/util/cuda_vector_matrix.hpp
```diff
@@ -1,7 +1,7 @@
 #include <micm/util/cuda_vector_matrix.cuh>
 #include <micm/util/vector_matrix.hpp>
 #include <type_traits>
-
+#include <iostream>
```
mattldawson, line 4: is this still needed?

mattldawson (changes requested):
looks great! just a couple minor suggestions

include/micm/solver/cuda_rosenbrock.hpp
```diff
@@ -76,5 +100,23 @@ namespace micm
...
     }
+
+    /// @brief Computes the scaled norm of the vector errors on the GPU; assume all the data are GPU resident already
+    /// @param y_old the original vector
+    /// @param y_new the new vector
+    /// @param errors The computed errors
+    /// @return The scaled norm of the errors
+    double NormalizedError(MatrixPolicy<double>& y_old, 
+                           MatrixPolicy<double>& y_new,
+                           MatrixPolicy<double>& errors) const
```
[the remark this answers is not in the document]
sjsprecious: Thanks for your suggestion. Revised it as suggested.

include/micm/util/cuda_vector_matrix.hpp
sjsprecious: I thought I had removed all the debug code but I forgot the header file. Removed.

mattldawson (approved):

PR #426, opened 2024-03-07
393 create cudasparsematrix class

### LINKED ISSUE #393: Create `CudaSparseMatrix` class
Extend the `SparseMatrix<T, SparseMatrixVectorOrdering>` class for use with CUDA solver functions.

Acceptance Criteria:

- `CudaSparseMatrix` class should allocate/deallocate device memory for matrix data on construction/destruction
- Matrix data is readable/writable from host and device

**Ideas**
- It might be easiest to add something like `SyncHost` and `SyncDevice` functions that would explicitly copy matrix data from device-to-host and host-to-device

mwaxmonsky:
Creates a CUDA capable version of SparseMatrix.

Has similar caveats to VectorMatrix in that the user is responsible for managing the synchronization of memory between the host and the device.

It is also imilar to CudaVectorMatrix in that only double types are supported via GPUs.  All other data types will only be managed and modified on the host.

This also updates the matrix policy tests used in the sparse matrix tests to take in a matrix template instead of a dedicated matrix type so they can be used in both matrix tests.

Closes #393


sjsprecious:
@mattldawson Apologize that I may have asked this 100 times. Are we saving all the elements of a sparse matrix like Jacobian? Or just the non-zero elements?

mattldawson:
> @mattldawson Apologize that I may have asked this 100 times. Are we saving all the elements of a sparse matrix like Jacobian? Or just the non-zero elements?

Just the non-zero elements. It's in the CSR format.

sjsprecious:
> > @mattldawson Apologize that I may have asked this 100 times. Are we saving all the elements of a sparse matrix like Jacobian? Or just the non-zero elements?
> 
> Just the non-zero elements. It's in the CSR format.

Thanks @mattldawson . If we just save the non-zero elements, why do we still need the `SparseMatrix` class since everything should be dense now?

mattldawson:
> > > @mattldawson Apologize that I may have asked this 100 times. Are we saving all the elements of a sparse matrix like Jacobian? Or just the non-zero elements?
> > 
> > 
> > Just the non-zero elements. It's in the CSR format.
> 
> Thanks @mattldawson . If we just save the non-zero elements, why do we still need the `SparseMatrix` class since everything should be dense now?

I'm not quite sure I follow, but the SparseMatrix class lets you access elements as if it were a dense matrix. For example, for the matrix (x = zero element):
```
J = [ 2 X 6 ]
    [ 1 4 X ]
    [ X X 9 ]
```
then, `J[1][1] == 4`, but `J.AsVector()[3] == 4`. Is this what you were asking?

sjsprecious:
Thanks @mattldawson . The example helps a lot.

Yes, I am confused by `J` and `J.AsVector()` here. I think `J.AsVector()` is dense in this case and contains all the non-zero elements. But `J` is the full-matrix format and it is sparse, which means we are still storing the zero-value elements in the memory. Is this correct?

The reason why I asked this is that if we could pass `J.AsVector()` to the CUDA kernel, then we could just declare it as `CudaVectorMatrix` class. But if we are passing `J` as a parameter to the CUDA kernel, then we need the `CudaSparseMatrix` class here if I understand it correctly.

mattldawson:
> Thanks @mattldawson . The example helps a lot.
> 
> Yes, I am confused by `J` and `J.AsVector()` here. I think `J.AsVector()` is dense in this case and contains all the non-zero elements. But `J` is the full-matrix format and it is sparse, which means we are still storing the zero-value elements in the memory. Is this correct?
> 
> The reason why I asked this is that if we could pass `J.AsVector()` to the CUDA kernel, then we could just declare it as `CudaVectorMatrix` class. But if we are passing `J` as a parameter to the CUDA kernel, then we need the `CudaSparseMatrix` class here if I understand it correctly.

I wouldn't try to use `CudaVectorMatrix` here. Even though both `CudaVectorMatrix` and `CudaSparseMatrix` store their underlying data in `std::vector<double>`, they each have different functionality, including different `[]` operator functions (which allows you to access elements in the sparse matrix with the `J[1][1]` syntax even though we aren't storing non-zero elements in memory)

sjsprecious:
oh, so `J[1][1]` is just for my convenience to access a non-zero element in `J`, not saying that we created a full `J` matrix in the memory?

For example, assume `J` is an MxM matrix originally with a sparse pattern. The memory allocation of `J` in the `SparseMatrix` class is not MxM, but N, where N is the total number of non-zero elements in the `J` matrix?

mattldawson:
> oh, so `J[1][1]` is just for my convenience to access a non-zero element in `J`, not saying that we created a full `J` matrix in the memory?
> 
> For example, assume `J` is an MxM matrix originally with a sparse pattern. The memory allocation of `J` in the `SparseMatrix` class is not MxM, but N, where N is the total number of non-zero elements in the `J` matrix?

That's right - the `J[1][1]` syntax is mostly useful in tests, or maybe as an initial implementation of some function, before you write the logic to access data directly in the underlying vector.

sjsprecious:
Thanks @mattldawson a lot for your clarification. That is clear to me now.

include/micm/util/cuda_sparse_matrix.hpp
```diff
@@ -0,0 +1,91 @@
...
+    {
+      return micm::cuda::CopyToDevice(param_, this->data_);
+    }
+    int CopyToHost()
+    {
+      return micm::cuda::CopyToDevice(param_, this->data_);
+    }
+    CudaVectorMatrixParam AsDeviceParam()
+    {
+      return CudaVectorMatrixParam{ param_.d_data_, param_.num_elements_ };
```
mattldawson, line 87: why not just return a copy of `this->param_`? Just in case `CudaVectorMatrixParam` changes in the future for any reason

test/unit/util/test_cuda_sparse_matrix_vector_ordering.cpp
```diff
@@ -0,0 +1,230 @@
...
+  auto matrix = testZeroMatrix<micm::CudaSparseMatrix, micm::SparseMatrixVectorOrdering<2>>();
+
+  matrix.CopyToDevice();
+  auto param = matrix.AsDeviceParam();
+  micm::cuda::AddOneDriver(param);
+
+  for(auto& elem : matrix.AsVector())
+  {
+    EXPECT_EQ(elem, 1.0);
+  }
```
mattldawson, line 20: I'm probably just not understanding something here, but if the `AddOneDriver()` adds one on the device side, how is this checking the host data without a call to `CopyToHost()`?

test/unit/util/test_cuda_sparse_matrix_vector_ordering.cpp
```diff
@@ -0,0 +1,230 @@
...
+}
+
+TEST(CudaSparseVectorMatrix, CopyAssignmentConstZeroMatrix)
+{
+  const auto matrix = testConstZeroMatrix<micm::CudaSparseMatrix, micm::SparseMatrixVectorOrdering<3>>();
+  auto oneMatrix = matrix;
+  oneMatrix.CopyToDevice();
```
mattldawson, line 34: why is a `CopyToDevice()` needed here? Should this be called on `matrix` prior to the copy assign?

test/unit/util/test_cuda_sparse_matrix_vector_ordering.cpp
```diff
@@ -0,0 +1,230 @@
...
+  for(auto& elem : oneMatrix.AsVector())
+  {
+    EXPECT_EQ(elem, 1.0);
+  }
+}
+
+TEST(CudaSparseVectorMatrix, MoveAssignmentConstZeroMatrix)
+{
+  const auto matrix = testConstZeroMatrix<micm::CudaSparseMatrix, micm::SparseMatrixVectorOrdering<3>>();
+  auto oneMatrix = std::move(matrix);
```
mattldawson, line 108: it might be good to have some device-side data on `matrix` prior to the move, to make sure it gets moved correctly

mattldawson (changes requested):
looks great! just a couple questions about the host<->device copying

include/micm/util/cuda_sparse_matrix.hpp
```diff
@@ -0,0 +1,91 @@
...
+      return *this = CudaSparseMatrix(other);
+    }
+
+    CudaSparseMatrix& operator=(CudaSparseMatrix&& other) noexcept
+    {
+      std::swap(this->data_, other.data_);
+      std::swap(this->param_, other.param_);
+      std::swap(this->number_of_blocks_, other.number_of_blocks_);
+      std::swap(this->row_ids_, other.row_ids_);
+      std::swap(this->row_start_, other.row_start_);
```
sjsprecious, line 67: what are these three variables for? should them be part of the `param_` struct and used on GPU later? If so, I guess we should use another Param struct rather than `CudaVectorMatrixParam`?

test/unit/util/cuda_matrix_utils.cu
```diff
@@ -13,9 +13,23 @@ namespace micm
...
+          std::size_t tid = blockIdx.x * blockDim.x + threadIdx.x;
+          if (tid < num_elements)
+          {
+              d_data[tid] += 1.0;
+          }
+      }
+
+      void AddOneDriver(CudaVectorMatrixParam& param)
+      {
+        AddOne<<<param.num_elements_, 1>>>(param.d_data_, param.num_elements_);
```
sjsprecious, line 32: Why use `1` rather than the predefined `BLOCK_SIZE` here?

sjsprecious (commented):
Thanks @mwaxmonsky for working on this PR. It looks great to me and I just have two minor questions besides Matt's comments.

include/micm/util/cuda_sparse_matrix.hpp
mwaxmonsky: So I didn't create another struct because I didn't see where these were being used on the GPU but if they are needed on the device, I should make a new struct for matrix params.

test/unit/util/cuda_matrix_utils.cu
mwaxmonsky: Good catch!  Added!

include/micm/util/cuda_sparse_matrix.hpp
mwaxmonsky: Would `return this->param;` do the proper copy?  I guess I must have got worried with the pointer management side of things and figured this was the safer approach.

include/micm/util/cuda_sparse_matrix.hpp
sjsprecious: @mattldawson can you provide an example where these three variables are used on CPU? I wonder maybe I miss a CUDA implementation somewhere in this case.

sjsprecious: Aha, thanks. I actually wonder how the index is mapped between the sparse matrix and vector in your yesterday's example and now I see these variables do the trick. Thank you for the clarification.

include/micm/util/cuda_sparse_matrix.hpp
mattldawson: yes, that should work because `this->param_` itself is not a pointer (even though `this` is), and the `CudaVectorMatrixParam` pointer data member (`data_`) you don't want to clone, you actually just want to copy the reference

include/micm/util/cuda_sparse_matrix.hpp
```diff
@@ -0,0 +1,91 @@
...
+      : SparseMatrix<T, OrderingPolicy>(SparseMatrix<T, OrderingPolicy>::create(other.number_of_blocks_))
+    {
+      this->data_ = std::move(other.data_);
+      this->param_ = std::move(other.param_);
+    }
+
+    CudaSparseMatrix& operator=(const CudaSparseMatrix& other)
+    {
+      return *this = CudaSparseMatrix(other);
+    }
```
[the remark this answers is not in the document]
mwaxmonsky: I may be mistaken, bit I thought this calls the copy constructors above it which are specialized between double and non-double.

include/micm/util/cuda_sparse_matrix.hpp
mwaxmonsky: Updated!

test/unit/util/test_cuda_sparse_matrix_vector_ordering.cpp
mwaxmonsky: Great catch!  This was creating an empty matrix so we weren't actually testing anything.  Updated to actually build a matrix with elements.

test/unit/util/test_cuda_sparse_matrix_vector_ordering.cpp
mwaxmonsky: So my thoughts here were that we'd create the const matrix and copy the data to the device on the new matrix and test that the original matrix stays the same throughout the process.

test/unit/util/test_cuda_sparse_matrix_vector_ordering.cpp
mwaxmonsky: I believe we are doing that in `MoveAssignmentDeSyncedHostZeroMatrix`?

mattldawson (changes requested):
looks great! I think in the process, the original `ZeroMatrix` and `ConstZeroMatrix` may have been lost (unless I missed them), similar to these: https://github.com/NCAR/micm/blob/a2be7013588013699d1378a6a10f550ab8755b24/test/unit/util/test_sparse_matrix_vector_ordering.cpp#L8-L16

mattldawson (approved):

PR #427, opened 2024-03-11
Actions

K20shores:
Removes a large directory to allow intel dockerfile to run

also edits some tests to use `EXPECT_DOUBLE_EQ`. Intel was failing these since the expected value, which was either calculated or set directly, would always have more bits of precision than whatever was read it, for some reason


docker/Dockerfile.nvhpc
```diff
@@ -28,26 +28,17 @@ ENV FC=nvfortran
...
-      && cd /cuda_build \ 
+RUN mkdir /build \
+      && cd /build \ 
       && cmake \
         -D CMAKE_BUILD_TYPE=debug \
         -D MICM_ENABLE_CUDA=ON \
         -D MICM_GPU_TYPE="a100" \
         ../micm \
-      && make -j 8 install
+      && make
```
sjsprecious, line 42: So for the nvhpc container, we just build the cuda version and do not run it?

test/regression/RosenbrockChapman/regression_test_p_force_policy.hpp
```diff
@@ -39,7 +39,7 @@ void testRateConstants(OdeSolverPolicy& solver)
     EXPECT_EQ(state.rate_constants_[i].size(), fixed_state.rate_constants_[0].size());
     for (size_t j{}; j < state.rate_constants_[i].size(); ++j)
     {
-      EXPECT_EQ(state.rate_constants_[i][j], fixed_state.rate_constants_[0][j]);
+      EXPECT_DOUBLE_EQ(state.rate_constants_[i][j], fixed_state.rate_constants_[0][j]);
```
sjsprecious, line 42: If it fails with Intel compiler only, maybe it is related to the extend 80-bit precision used by the Intel compiler (https://en.wikipedia.org/wiki/Extended_precision#:~:text=The%20x86%20extended%20precision%20format,floating%2Dpoint%20arithmetic%20in%20hardware.)? No change is needed and I just want to provide additional information.

docker/Dockerfile.nvhpc
K20shores: That's correct; in the container it will run the the cuda code, but since there's no GPU all of the initial values will be returned and we receive failures. The best we can do, as before, is build the GPU code. I removed the code that will build and test the CPU version because I thought it was redundant. I will add it back if you prefer

sjsprecious (approved):
LGTM. Thanks @K20shores for working on this.

boulderdaze (approved):

mwaxmonsky (approved):
Looks good to me.  Great work @K20shores!

src/CMakeLists.txt
```diff
@@ -17,6 +17,19 @@ target_include_directories(micm
...
+if (CMAKE_CXX_COMPILER_ID STREQUAL "Intel" OR CMAKE_CXX_COMPILER_ID STREQUAL "IntelLLVM")
+  # Set compiler flags for Intel C++
+  if (WIN32)
+    # For Windows
+    target_compile_options(micm INTERFACE "/fp:precise")
+  else ()
+    # For other operating systems (e.g., Linux, macOS)
+    target_compile_options(micm INTERFACE "-ffp-model=precise")
+  endif ()
+endif ()
```
K20shores, line 30: @sjsprecious I added these compiler flags and I was able to go back to exact equality

K20shores: I tried the `double` option, but icpx didn't recognize that as a keyword. However, `precise` worked

test/regression/RosenbrockChapman/regression_test_p_force_policy.hpp
K20shores: fixed by setting compiler flags

mattldawson (approved):
nice!

PR #429, opened 2024-03-14
Profiling and Optimization for VectorMatrix

boulderdaze:
Related issue: [Collect timing of individual solver functions in MICM and CAM-Chem #35](https://github.com/NCAR/MUSICA-Performance-Comparison/issues/35)
- Added MICM Instrumentation tool
- Reduce copy operations
- Reduce dereference operations
- Convert lambda operation to regular `for` loop operation  for `ForEach`

#### Results
1. Performance Comparison - `main` vs `profiling` vs `remove_for_each`.  Benchmark  is CAM-CHEM, TS1 configuration used. (as of now 04/18/24)
```
-------------------------------------------------------------------------------
Benchmark                          main         profiling      remove_for_each
-------------------------------------------------------------------------------
CAM-CHEM/1                       0.783 ms        0.777 ms         0.778 ms 
CAM-CHEM/10                       1.29 ms         1.31 ms          1.28 ms
CAM-CHEM/30                       2.44 ms         2.41 ms          2.43 ms
CAM-CHEM/90                       5.90 ms         5.86 ms          5.81 ms
CAM-CHEM/100                      6.51 ms         6.38 ms          6.42 ms
CAM-CHEM/480                      28.5 ms         28.1 ms          28.4 ms
CAM-CHEM/1000                     59.0 ms         59.7 ms          58.2 ms
CAM-CHEM/1440                     82.9 ms         86.8 ms          89.3 ms
Rosenbrock-DenseMatrix/1         0.296 ms        0.284 ms         0.265 ms
Rosenbrock-DenseMatrix/10         2.62 ms         2.45 ms          2.41 ms
Rosenbrock-DenseMatrix/30         7.80 ms         7.14 ms          7.09 ms
Rosenbrock-DenseMatrix/90         23.1 ms         21.6 ms          21.3 ms
Rosenbrock-DenseMatrix/100        25.5 ms         23.8 ms          24.4 ms
Rosenbrock-DenseMatrix/480         125 ms          121 ms           114 ms
Rosenbrock-DenseMatrix/1000        267 ms          264 ms           249 ms
Rosenbrock-DenseMatrix/1440        398 ms          375 ms           364 ms
Rosenbrock-VectorMatrix/1        0.426 ms        0.323 ms         0.328 ms
Rosenbrock-VectorMatrix/10        1.80 ms         1.62 ms          1.55 ms
Rosenbrock-VectorMatrix/30        3.85 ms         3.29 ms          3.23 ms
Rosenbrock-VectorMatrix/90        11.4 ms         9.56 ms          9.48 ms
Rosenbrock-VectorMatrix/100       12.7 ms         10.7 ms          10.7 ms
Rosenbrock-VectorMatrix/480       65.2 ms         58.9 ms          55.8 ms
Rosenbrock-VectorMatrix/1000       146 ms          125 ms           122 ms
Rosenbrock-VectorMatrix/1440       211 ms          185 ms           178 ms
```
2. Profile of the Individual functions in `Rosenbrock::Solve()` - VectorMatrix, 1000 grid cells
```
----------------------------------------------------------------------------------------------
                              Average Wall Duration                       Wall Duration
                              main        profiling   Occurrences      main        profiling
----------------------------------------------------------------------------------------------
AddForcingTerms              0.703 ms       0.686 ms      22        15.458 ms       15.097 ms
AlphaMinusJacobian           0.071 ms       0.077 ms      11         0.778 ms        0.849 ms
CalculateForcing             0.729 ms       0.710 ms      22        16.040 ms       15.619 ms
CalculateNegativeJacobian    1.537 ms       1.521 ms      11        16.906 ms       16.730 ms
Decompose                    3.403 ms       3.393 ms      11        37.435 ms       37.320 ms
Factor                       3.423 ms       3.414 ms      11        37.652 ms       37.549 ms
LinearFactor                 4.627 ms       4.754 ms      11        50.901 ms       52.297 ms
NormalizedError              0.490 ms       0.189 ms      11         5.395 ms        2.081 ms
Solve                        0.693 ms       0.686 ms      33        22.854 ms       22.632 ms
SubtractJacobianTerms        1.399 ms       1.376 ms      11        15.387 ms       15.138 ms
UpdateState                  4.199 ms       4.069 ms      1          4.199 ms        4.069 ms
ForEach                      0.220 ms          -          110       24.153 ms           -
Axpy                            -           0.045 ms      110            -           4.986 ms
```


include/micm/profiler/instrumentation.hpp
```diff
@@ -0,0 +1,225 @@
...
+      }
+
+      return result;
+    }
+  }
+
+}  // namespace micm
+
+
+#define MICM_PROFILE 1
```
K20shores, line 202:
```suggestion
```

in `src/CMakeLists.txt`, would you please add something like

```
if (MICM_ENABLE_PROFILE)
    target_compile_definitions(micm PUBLIC MICM_PROFILE)
endif()
```

then the above line wouldn't be needed

CMakeLists.txt
```diff
@@ -39,6 +39,7 @@ option(MICM_ENABLE_OPENACC "Build with OpenACC Support" OFF)
 option(MICM_ENABLE_LLVM "Build with LLVM support for JIT-compiling" OFF)
 option(MICM_ENABLE_TESTS "Build the tests" ON)
 option(MICM_ENABLE_EXAMPLES "Build the examples" ON)
+option(MICM_ENABLE_PROFILE "Build the instrumentation" ON)
```
K20shores, line 42:
```suggestion
option(MICM_ENABLE_PROFILE "Build the instrumentation. MICM_ENABLE_JSON must also be enabled " ON)
```

include/micm/solver/rosenbrock.inl
```diff
@@ -31,11 +14,6 @@ namespace micm
     decompositions = 0;
     solves = 0;
     singular = 0;
-    total_update_state_time = std::chrono::nanoseconds::zero();
-    total_forcing_time = std::chrono::nanoseconds::zero();
-    total_jacobian_time = std::chrono::nanoseconds::zero();
-    total_linear_factor_time = std::chrono::nanoseconds::zero();
-    total_linear_solve_time = std::chrono::nanoseconds::zero();
```
K20shores, line 38: With these removed, the timing example will need to be updated. Also, the cuda interface has some functionality to collect the time. I'm not sure how that could fit in with your profiling class. Could you look into that?

K20shores (changes requested):
Overall I like this approach much better. Just need to make sure it also applies to the cuda code, also probably the JIT code if we can

CMakeLists.txt
boulderdaze: The function `WriteProfile(const ProfileResult& result)` in `micm/profiler/instrumentation.hpp` handles the JSON format so I don't think that JSON is required for profiling by itself. What do you think?

include/micm/profiler/instrumentation.hpp
boulderdaze: Good suggestion, thanks!

include/micm/solver/rosenbrock.inl
boulderdaze: Yeah, I removed them from cuda files and will request reviews from Jian and Michael

K20shores (changes requested):
Everything looks good! Could you update this [tutorial](https://ncar.github.io/micm/user_guide/but_how_fast_is_it.html) to reflect these changes?

mattldawson (changes requested):
This is really nice! And the performance results look really good!

The only thing I would suggest is to leave the `ForEach()` function  and create a separate function on the matrix class for the specific function in the solver algorithm. Jian just added something similar to the CUDA matrix class, so if you use the same function signature:
https://github.com/NCAR/micm/blob/3be56e2b2d287d0d2a018d80c89f82c850b2605c/include/micm/util/cuda_dense_matrix.hpp#L178-L193
... and update the `Rosenbrock::Solve()` function to call that instead of `ForEach()` then it should work with the CPU and GPU code

include/micm/solver/rosenbrock.inl
sjsprecious: I removed the timing interface from the CUDA kernels so it should be fine to clean up these timers.

sjsprecious:
> This is really nice! And the performance results look really good!
> 
> The only thing I would suggest is to leave the `ForEach()` function and create a separate function on the matrix class for the specific function in the solver algorithm. Jian just added something similar to the CUDA matrix class, so if you use the same function signature:
> 
> https://github.com/NCAR/micm/blob/3be56e2b2d287d0d2a018d80c89f82c850b2605c/include/micm/util/cuda_dense_matrix.hpp#L178-L193
> 
> 
> ... and update the `Rosenbrock::Solve()` function to call that instead of `ForEach()` then it should work with the CPU and GPU code

Thanks @mattldawson for this suggestion. It will be great if there is a corresponding `axpy` implementation in the CPU code so I won't need to worry about changing the CUDA interface later.

sjsprecious (commented):
Thanks for working on this giant PR and providing the performance data. It looks nice! Just want to clarify: CAM-Chem does not use rosenbrock solver for chemistry so what does ros2 mean in the benchmark?

boulderdaze:
> This is really nice! And the performance results look really good!
> 
> The only thing I would suggest is to leave the `ForEach()` function and create a separate function on the matrix class for the specific function in the solver algorithm. Jian just added something similar to the CUDA matrix class, so if you use the same function signature:
> 
> https://github.com/NCAR/micm/blob/3be56e2b2d287d0d2a018d80c89f82c850b2605c/include/micm/util/cuda_dense_matrix.hpp#L178-L193
> 
> 
> ... and update the `Rosenbrock::Solve()` function to call that instead of `ForEach()` then it should work with the CPU and GPU code

No problem, do you have any suggestion for the function name? It's hard to come up with  a name for this set of calculation. Also @K20shores mentioned we could remove this function to speed up. What do you think?

boulderdaze:
I recollected the CAM-CHEM profile data. I learned that the previous data was collected with timing on and also writing output results which seem to take significant amount of time.

mattldawson:
> > This is really nice! And the performance results look really good!
> > The only thing I would suggest is to leave the `ForEach()` function and create a separate function on the matrix class for the specific function in the solver algorithm. Jian just added something similar to the CUDA matrix class, so if you use the same function signature:
> > https://github.com/NCAR/micm/blob/3be56e2b2d287d0d2a018d80c89f82c850b2605c/include/micm/util/cuda_dense_matrix.hpp#L178-L193
> > 
> > ... and update the `Rosenbrock::Solve()` function to call that instead of `ForEach()` then it should work with the CPU and GPU code
> 
> No problem, do you have any suggestion for the function name? It's hard to come up with a name for this set of calculation. Also @K20shores mentioned we could remove this function to speed up. What do you think?

I agree, these math functions are difficult to find good names for. I would suggest using the exact same as Jian did: `Axpy(const double alpha, const Matrix<T, L>& x, const int incx, const int incy)`

This way his function will override yours for the `CudaDenseMatrix` class. I don't think we should remove the function, because it's different for `Matrix`, `VectorMatrix` and `CudaDenseMatrix`.

boulderdaze:
> > > This is really nice! And the performance results look really good!
> > > The only thing I would suggest is to leave the `ForEach()` function and create a separate function on the matrix class for the specific function in the solver algorithm. Jian just added something similar to the CUDA matrix class, so if you use the same function signature:
> > > https://github.com/NCAR/micm/blob/3be56e2b2d287d0d2a018d80c89f82c850b2605c/include/micm/util/cuda_dense_matrix.hpp#L178-L193
> > > 
> > > ... and update the `Rosenbrock::Solve()` function to call that instead of `ForEach()` then it should work with the CPU and GPU code
> > 
> > 
> > No problem, do you have any suggestion for the function name? It's hard to come up with a name for this set of calculation. Also @K20shores mentioned we could remove this function to speed up. What do you think?
> 
> I agree, these math functions are difficult to find good names for. I would suggest using the exact same as Jian did: `Axpy(const double alpha, const Matrix<T, L>& x, const int incx, const int incy)`
> 
> This way his function will override yours for the `CudaDenseMatrix` class. I don't think we should remove the function, because it's different for `Matrix`, `VectorMatrix` and `CudaDenseMatrix`

Sorry I accidentally edited your comment. It's weird I can edit somebody's comment! 
There are only two arguments in our case, but I will follow the naming.

boulderdaze:
> Thanks for working on this giant PR and providing the performance data. It looks nice! Just want to clarify: CAM-Chem does not use rosenbrock solver for chemistry so what does ros2 mean in the benchmark?

Thank you for the question. I updated it!

boulderdaze:
Updates:
- Recollected profile data after a couple of new branches merged into `main`
- Added Kyle's branch (`remove ForEach`) for the comparison
- The presented data is the mid value in solving time among 6 attempts of solver run.

boulderdaze:
> Everything looks good! Could you update this [tutorial](https://ncar.github.io/micm/user_guide/but_how_fast_is_it.html) to reflect these changes?

Yes, I updated it. Could you review it?

include/micm/process/process_set.hpp
```diff
@@ -315,15 +333,18 @@ namespace micm
       {
         for (std::size_t i_ind = 0; i_ind < number_of_reactants_[i_rxn]; ++i_ind)
         {
-          std::vector<double> d_rate_d_ind(L, 0);
-          for (std::size_t i_cell = 0; i_cell < L; ++i_cell)
-            d_rate_d_ind[i_cell] = v_rate_constants[offset_rc + i_rxn * L + i_cell];
+          std::fill(d_rate_d_ind.begin(), d_rate_d_ind.end(), 0);
```
K20shores, line 336: is this necessary?

K20shores (approved):

include/micm/solver/lu_decomposition.inl
```diff
@@ -264,52 +278,58 @@ namespace micm
       auto lkj_uji = lkj_uji_.begin();
       auto uii = uii_.begin();
       is_singular = false;
-      const std::size_t n_cells = std::min(A.GroupVectorSize(), A.size() - i_group * A.GroupVectorSize());
+      std::size_t n_cells = std::min(A_GroupVectorSize, A_Size - i_group * A_GroupVectorSize);
```
mwaxmonsky, line 281: Is there a reason this has to be non-`const` now?  I might be missing a detail below but nothing sticks out as modifying it.

include/micm/process/process_set.hpp
boulderdaze: Originally the vector `d_rate_d_ind` was created in the exact location (inside double nested `for` loop). I got it out of the loops to avoid keeping hitting the heap so It has to get refreshed to zero for every iteration. Maybe it makes more sense to put it at the end of the loop. If my intention doesn't seem straightforward, I am happy to get the original version back. I'd prefer readability for this case. What do you think?

mattldawson (approved):
looks great!

test/unit/util/test_matrix_policy.hpp
```diff
@@ -204,6 +205,32 @@ MatrixPolicy<double> testAssignmentFromVector()
   return matrix;
 }
 
+template<template<class> class MatrixPolicy>
+MatrixPolicy<double> testAxpy()
+{
+  MatrixPolicy<double> matrix{ 4, 3, 100.0 };
+  MatrixPolicy<double> other{ 4, 3, 200.0 };
```
mwaxmonsky, line 212:
```suggestion
  std::size_t num_rows = 4;
  std::size_t num_columns = 3;
  MatrixPolicy<double> matrix{ 4, 3, 100.0 };
  MatrixPolicy<double> other{ 4, 3, 200.0 };
```
That way we can reuse the dimensions below.

test/unit/util/test_matrix_policy.hpp
```diff
@@ -204,6 +205,32 @@ MatrixPolicy<double> testAssignmentFromVector()
...
+  double alpha = 1.39;
+  double sum = 0.0;
+  double result = 0.0;
+
+  for (int i = 0; i < 4; ++i)
+    for (int j = 0; j < 3; ++j)
+    {
+      matrix[i][j] = i * 10.3 + j * 100.5;
+      other[i][j] = i * 1.7 + j * 10.2;
+      sum += i * 10.3 + j * 100.5 + alpha * (i * 1.7 + j * 10.2);
```
mwaxmonsky, line 222:
```suggestion
      auto y = i * 10.3 + j * 100.5;
      auto x = i * 1.7 + j * 10.2;
      matrix[i][j] = y
      other[i][j] = x;
      sum += y + alpha * x;
```
This might be a little clearer and saves us from double computing x and y without needing to rely on the accessors for `matrix` and `other`.

mwaxmonsky (commented):
Looks great @boulderdaze!  Just some clarification requests and possibly some modifications that might be worth looking at.

examples/util_example.cpp
```diff
@@ -0,0 +1,194 @@
...
+  std::unordered_map<std::string, double> environments;
+  std::unordered_map<std::string, std::vector<double>> concentrations;
+  std::unordered_map<std::string, std::vector<double>> custom_rate_params;
+
+  bool empty()
+  {
+    if (!environments.empty())
+      return false;
+    else
+      return true;
```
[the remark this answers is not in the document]
boulderdaze: Good suggestion, thanks!

include/micm/process/process_set.hpp
boulderdaze: I think it actually doesn't need this. Removed, thanks for catching it!

include/micm/solver/lu_decomposition.inl
boulderdaze: No, It doesn't. I put `const` back in, thanks for noticing!

examples/util_example.cpp
```diff
@@ -0,0 +1,194 @@
...
+        return dataMap;
+      }
+      key = line.substr(0, second_last_delimiter_pos);
+      value_string = line.substr(second_last_delimiter_pos + 1, last_delimiter_pos - second_last_delimiter_pos);
+      try
+      {
+        value = std::stod(value_string);
+        dataMap.custom_rate_params[key + ".effective radius [m]"].emplace_back(value);
+      }
+      catch(const std::exception& e)
```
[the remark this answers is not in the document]
boulderdaze: No, I naively copied and pasted this portion of the code. Thanks for noticing it!! Fixed.

examples/util_example.cpp
```diff
@@ -0,0 +1,194 @@
...
+        dataMap.incomplete_parsing();
+        return dataMap;
+      }
+      value_string = line.substr(last_delimiter_pos + 1);
+      try
+      {
+        value = std::stod(value_string);
+        dataMap.custom_rate_params[key + ".particle number concentration [# m-3]"].emplace_back(value);
+      }
+      catch(const std::exception& e)
```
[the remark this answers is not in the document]
boulderdaze: Thanks!

test/unit/util/test_matrix_policy.hpp
boulderdaze: I like your suggestion. Done

test/unit/util/test_matrix_policy.hpp
boulderdaze: I like this, thanks!

include/micm/process/process_set.hpp
```diff
@@ -222,8 +233,10 @@ namespace micm
       std::vector<double> rate(L, 0);
       for (std::size_t i_rxn = 0; i_rxn < number_of_reactants_.size(); ++i_rxn)
       {
-        for (std::size_t i_cell = 0; i_cell < L; ++i_cell)
-          rate[i_cell] = v_rate_constants[offset_rc + i_rxn * L + i_cell];
+        std::copy(
+            v_rate_constants.begin() + offset_rc + i_rxn * L,
+            v_rate_constants.begin() + offset_rc + i_rxn * L + L,
+            rate.begin());
```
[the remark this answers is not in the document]
boulderdaze: I converted `copy` to `assign` for the lines you mentioned in `process_set.hpp`. In `linear_solver.inl`, they are all iterator unlike this case, so using `copy` could remain. Thank you for the suggestion!

include/micm/process/process_set.hpp
```diff
@@ -315,15 +333,18 @@ namespace micm
...
         for (std::size_t i_ind = 0; i_ind < number_of_reactants_[i_rxn]; ++i_ind)
         {
-          std::vector<double> d_rate_d_ind(L, 0);
-          for (std::size_t i_cell = 0; i_cell < L; ++i_cell)
-            d_rate_d_ind[i_cell] = v_rate_constants[offset_rc + i_rxn * L + i_cell];
+          std::fill(d_rate_d_ind.begin(), d_rate_d_ind.end(), 0);
+          std::copy(
+              v_rate_constants.begin() + offset_rc + i_rxn * L,
+              v_rate_constants.begin() + offset_rc + i_rxn * L + L,
+              d_rate_d_ind.begin());
```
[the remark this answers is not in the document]
boulderdaze: Done, thanks

mwaxmonsky (approved):
Great work @boulderdaze!  Thanks!

sjsprecious (approved):

PR #436, opened 2024-03-26
Optimize data movement for CudaRosenbrocksolver and compute minus Jacobian matrix on GPU

### LINKED ISSUE #391: Move device data management of `CudaRosenbrockSolver` data members to constructor/destructor
Move device memory allocation, deallocation, and copying to `CudaRosenbrockSolver`.

**Acceptance Criteria**
- device pointers for all index array data members of `CudaRosenbrockSolver` are created during construction
- potentially include rate constants, state variables, and jacobian in the data management here
- all device pointer data members are freed during deconstruction

**Ideas**
- allocate and copy the array of indices for the jacobian elements used [here](https://github.com/NCAR/micm/blob/9416a2f2106e9e86ff550d9d2fc0b395a49e5605/src/solver/rosenbrock.cu#L25-L26)

sjsprecious:
This PR moves the data movement of array of indices for the diagonal elements of Jacobian matrix to the constructor/destructor of `CudaRosenbrocksolver` class.

In addition:
- currently the operation of negating Jacobian matrix is done on the CPU (https://github.com/NCAR/micm/blob/main/include/micm/solver/cuda_rosenbrock.hpp#L94-L95) for the `AlphaMinusJacobian` function. This PR moves this operation to the `AddJacobianTermsKernel` function so the CUDA kernel now returns `-J` ranther than `J` matrix.
- the `AlphaMinusJacobianKernel` function is revised to achieve more parallelism on GPU. 

All the 40 tests passed on Derecho's A100 GPU with `nvhpc/23.7`.

Fix #391

### DIFF
--- a/src/process/process_set.cu
+++ b/src/process/process_set.cu
@@ -54,14 +54,15 @@ namespace micm
       }    // end of checking a valid CUDA thread id
     }      // end of AddForcingTerms_kernel
 
-    /// This is the CUDA kernel that forms the Jacobian matrix on the device
+    /// This is the CUDA kernel that forms the minus Jacobian matrix (-J) on the device
     __global__ void AddJacobianTermsKernel(
         double* d_rate_constants,
         double* d_state_variables,
         double* d_jacobian,
         ProcessSetParam devstruct,
-        size_t n_grids,
-        size_t n_reactions)
+        const size_t num_grid_cells,
+        const size_t n_reactions,
+        const size_t num_elments_per_jacobian)
     {
       /// Local device variables
       size_t tid = blockIdx.x * blockDim.x + threadIdx.x;
@@ -73,8 +74,9 @@ namespace micm
       size_t* d_number_of_products = devstruct.number_of_products_;
       size_t* d_jacobian_flat_ids = devstruct.jacobian_flat_ids_;
       double* d_yields = devstruct.yields_;
+      const size_t jacobian_flat_ids_size = devstruct.jacobian_flat_ids_size_;
 
-      if (tid < n_grids)
+      if (tid < num_grid_cells)
       {
         // loop over reactions in a grid
         for (size_t i_rxn = 0; i_rxn < n_reactions; ++i_rxn)
@@ -82,12 +84,12 @@ namespace micm
           // loop over reactants in a reaction
           for (size_t i_ind = 0; i_ind < d_number_of_reactants[i_rxn]; ++i_ind)
           {
-            double d_rate_d_ind = d_rate_constants[i_rxn * n_grids + tid];
+            double d_rate_d_ind = d_rate_constants[i_rxn * num_grid_cells + tid];
             for (size_t i_react = 0; i_react < d_number_of_reactants[i_rxn]; ++i_react)
             {
               if (i_react != i_ind)
               {
-                d_rate_d_ind *= d_state_variables[d_reactant_ids[react_ids_offset + i_react] * n_grids + tid];
+                d_rate_d_ind *= d_state_variables[d_reactant_ids[react_ids_offset + i_react] * num_grid_cells + tid];
               }
             }
             for (size_t i_dep = 0; i_dep < d_number_of_reactants[i_rxn]; ++i_dep)
@@ -106,6 +108,11 @@ namespace micm
           react_ids_offset += d_number_of_reactants[i_rxn];
           yields_offset += d_number_of_products[i_rxn];
         }  // end of loop over reactions in a grid
+        // Form the minus Jaocbian matrix (-J)
+        for (size_t idx = 0; idx < num_elments_per_jacobian; ++idx)
+        {
+            d_jacobian[idx*num_grid_cells + tid] = -d_jacobian[idx*num_grid_cells + tid]; 
+        }
       }    // end of checking a CUDA thread id
     }      // end of AddJacobianTermsKernel
 
@@ -193,11 +200,13 @@ namespace micm
       double* d_state_variables;
       double* d_jacobian;
 
+      size_t num_elments_per_jacobian = sparseMatrix.jacobian_size_ / matrixParam.n_grids_;
+
       // allocate device memory
       cudaMalloc(&d_rate_constants, sizeof(double) * matrixParam.n_grids_ * matrixParam.n_reactions_);
       cudaMalloc(&d_state_variables, sizeof(double) * matrixParam.n_grids_ * matrixParam.n_species_);
       cudaMalloc(&d_jacobian, sizeof(double) * sparseMatrix.jacobian_size_);
-
+      
       // transfer data from host to device
       cudaMemcpy(
           d_rate_constants,
@@ -219,7 +228,7 @@ namespace micm
       // launch kernel and measure time performance
       auto startTime = std::chrono::high_resolution_clock::now();
       AddJacobianTermsKernel<<<num_blocks, BLOCK_SIZE>>>(
-          d_rate_constants, d_state_variables, d_jacobian, devstruct, n_grids, n_reactions);
+          d_rate_constants, d_state_variables, d_jacobian, devstruct, n_grids, n_reactions, num_elments_per_jacobian);
       cudaDeviceSynchronize();
       auto endTime = std::chrono::high_resolution_clock::now();
       auto kernel_duration = std::chrono::duration_cast<std::chrono::nanoseconds>(endTime - startTime);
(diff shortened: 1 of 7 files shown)


src/process/process_set.cu
```diff
@@ -106,6 +108,11 @@ namespace micm
           react_ids_offset += d_number_of_reactants[i_rxn];
           yields_offset += d_number_of_products[i_rxn];
         }  // end of loop over reactions in a grid
+        // Form the minus Jaocbian matrix (-J)
+        for (size_t idx = 0; idx < num_elments_per_jacobian; ++idx)
+        {
+            d_jacobian[idx*num_grid_cells + tid] = -d_jacobian[idx*num_grid_cells + tid]; 
+        }
```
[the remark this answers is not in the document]
sjsprecious: Thanks @mwaxmonsky . I agree that moving this section to `AlphaMinusJacobianKernel` function will avoid changing the test for the `AddJacobianTerms` function. The reason I did not do it is that negating Jacobian in `AlphaMinusJacobianKernel` function will increase the resource request (i.e., we need to go through the whole Jacobian matrix rather than the diagonal elements only) and reduce the parallelism (i.e., each CUDA thread needs to loop through the `jacobian_diagonal_elements` in serial and checks if it is working on a diagonal element), compared to current implementation. In contrast, the `AddJacobianTerms` just parallelizes at the `num_grid_cells` level and all the other loops are done in serial. Thus negating the Jacobian matrix here does not request more resources and reduce the parallelism. I am happy to make the changes you suggested and we can revisit this implementation if they become a hotspot in the future performance tests.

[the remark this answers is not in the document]
mwaxmonsky: Hmmmmm, @mattldawson, is this a reasonable tradeoff?  We would bifurcate the meaning of the API but gain the efficiency we expect for GPUs.

include/micm/solver/rosenbrock.hpp
```diff
@@ -188,7 +188,7 @@ namespace micm
     /// @param rate_constants List of rate constants for each needed species
     /// @param number_densities The number density of each species
     /// @param jacobian The matrix of partial derivatives
```
mattldawson, line 190:
```suggestion
    /// @param jacobian The matrix of negative partial derivatives
```
Maybe also update the brief description of this function above to indicate that it returns the negative of the Jacobian

test/integration/e5.hpp
```diff
@@ -104,7 +104,7 @@ class E5 : public micm::RosenbrockSolver<MatrixPolicy, SparseMatrixPolicy, Linea
   /// @param rate_constants List of rate constants for each needed species
   /// @param number_densities The number density of each species
   /// @param jacobian The matrix of partial derivatives
```
mattldawson, line 106:
```suggestion
  /// @param jacobian The matrix of negative partial derivatives
```

test/integration/hires.hpp
```diff
@@ -105,7 +105,7 @@ class HIRES : public micm::RosenbrockSolver<MatrixPolicy, SparseMatrixPolicy, Li
   /// @param rate_constants List of rate constants for each needed species
   /// @param number_densities The number density of each species
   /// @param jacobian The matrix of partial derivatives
```
mattldawson, line 107:
```suggestion
  /// @param jacobian The matrix of negative partial derivatives
```

test/integration/oregonator.hpp
```diff
@@ -100,7 +100,7 @@ class Oregonator : public micm::RosenbrockSolver<MatrixPolicy, SparseMatrixPolic
   /// @param rate_constants List of rate constants for each needed species
   /// @param number_densities The number density of each species
   /// @param jacobian The matrix of partial derivatives
```
mattldawson, line 102:
```suggestion
  /// @param jacobian The matrix of negative partial derivatives
```

test/unit/process/test_cuda_process_set.cpp
```diff
@@ -147,7 +147,7 @@ void testRandomSystemAddJacobianTerms(std::size_t n_cells, std::size_t n_reactio
     elem = get_double();
 
   auto non_zero_elements = cpu_set.NonZeroJacobianElements();
-  auto builder = SparseMatrixPolicy<double>::create(n_species).number_of_blocks(n_cells).initial_value(100.0);
+  auto builder = SparseMatrixPolicy<double>::create(n_species).number_of_blocks(n_cells).initial_value(0.0);
```
mattldawson, line 150: Is this change still necessary, now that the CPU and GPU functions do the same thing?

test/unit/process/test_process_set_policy.hpp
```diff
@@ -98,36 +98,36 @@ void testProcessSet(const std::function<ProcessSetPolicy(
   compare_pair(*(++elem), index_pair(4, 0));
   compare_pair(*(++elem), index_pair(4, 2));
 
-  auto builder = SparseMatrixPolicy<double>::create(5).number_of_blocks(2).initial_value(100.0);
+  auto builder = SparseMatrixPolicy<double>::create(5).number_of_blocks(2).initial_value(0.0);
```
mattldawson, line 101: I would suggest leaving the initial value 100.0 and modifying the expectations to be `100 - ...` instead of `100 + ...`, just because the functions should preserve any existing values in the Jacobian (this may be important when we add aerosol processes that may have separate functions that modify the Jacobian)

mattldawson (changes requested):
looks great! just a couple minor suggestions. In addition to the inline comments, I would suggest modifying the description of the `AlphaMinusJacobian` function in `Rosenrock.hpp` to reflect the fact that it expects the negative jacobian as an input argument.

include/micm/solver/rosenbrock.hpp
sjsprecious: Done.

test/integration/e5.hpp
sjsprecious: Done.

test/integration/hires.hpp
sjsprecious: Done.

test/integration/oregonator.hpp
sjsprecious: Done.

test/unit/process/test_cuda_process_set.cpp
sjsprecious: I think so. The current implementation of CPU and GPU works only when the initial values of Jacobian matrix are zero. Otherwise, we have to write a separate loop to negate the matrix, like my original GPU implementation.

test/unit/process/test_process_set_policy.hpp
sjsprecious: Good point. But ins this case, we can not negate the matrix like the current implementation. If there is an initial value of Jacobian matrix such as `100`, then the calculated Jacobian will be `100+a-b` and the negative Jacobian will be `-100-a+b`. To compute this negative Jacobian matrix correctly, we will then have to write a separate loop to negate it.

mattldawson: I'm not sure that applies to this test - this just tests the function that subtracts jacobian elements from a matrix. It shouldn't care what values are there to start with. If all the matrix elements start at `100`, and jacobian terms being subtracted from a particular element are `a` and `-b`, then the final value for that element should be `100-a+b`

test/unit/process/test_cuda_process_set.cpp
mattldawson: But, isn't this just testing the function that subtracts values from a matrix? I don't think it should care if it starts at 0, or 100, or -100. If the CPU and GPU functions are doing the same thing, they should get the same answer.

sjsprecious: After a second thought, I agree with you that for this test, it should not matter what the initial value is for the Jacobian matrix. We just need to make sure that when running the whole solver, the initial value of Jacobian matrix must be zero to function properly.

test/unit/process/test_process_set_policy.hpp
sjsprecious: Agreed. Thanks for your suggestions.

sjsprecious: Done.

test/unit/process/test_cuda_process_set.cpp
sjsprecious: Done.

mattldawson (approved):
very nice!

mwaxmonsky (approved):
Looks great!  One little formatting change might be needed but nothing to hold it up!

PR #439, opened 2024-03-29
Update the CUDA implementation of AddforcingTerms with the CudaVectorMatrix class

sjsprecious:
This PR updates the input type for the CUDA implementation of `AddforcingTerms` function with the `CudaVectorMatrix` class.

I have to comment out the `test_cuda_rosenbrock` test for now because it requests me to update all the other CUDA kernels to pass. The other 40 tests all passed on Derecho's A100 GPU by using `nvhpc/23.7`

### DIFF
--- a/include/micm/process/cuda_process_set.hpp
+++ b/include/micm/process/cuda_process_set.hpp
@@ -31,11 +31,10 @@ namespace micm
     void SetJacobianFlatIds(const SparseMatrix<double, OrderingPolicy>& matrix);
 
     template<template<class> typename MatrixPolicy>
-    requires VectorizableDense<MatrixPolicy<double>> std::chrono::nanoseconds AddForcingTerms(
+    requires VectorizableDense<MatrixPolicy<double>> void AddForcingTerms(
         const MatrixPolicy<double>& rate_constants,
         const MatrixPolicy<double>& state_variables,
-        MatrixPolicy<double>& forcing)
-    const;
+        MatrixPolicy<double>& forcing);
 
     template<template<class> class MatrixPolicy, template<class> class SparseMatrixPolicy>
     requires VectorizableDense<MatrixPolicy<double>> && VectorizableSparse<SparseMatrixPolicy<double>>
@@ -91,21 +90,20 @@ namespace micm
 
   template<template<class> class MatrixPolicy>
   requires VectorizableDense<MatrixPolicy<double>>
-  inline std::chrono::nanoseconds CudaProcessSet::AddForcingTerms(
+  inline void CudaProcessSet::AddForcingTerms(
       const MatrixPolicy<double>& rate_constants,
       const MatrixPolicy<double>& state_variables,
-      MatrixPolicy<double>& forcing) const
+      MatrixPolicy<double>& forcing)
   {
-    CudaMatrixParam matrix;
-    matrix.rate_constants_ = rate_constants.AsVector().data();
-    matrix.state_variables_ = state_variables.AsVector().data();
-    matrix.forcing_ = forcing.AsVector().data();
-    matrix.n_grids_ = rate_constants.size();
-    matrix.n_reactions_ = rate_constants[0].size();
-    matrix.n_species_ = state_variables[0].size();
+    this->devstruct_.number_of_reactions_ = rate_constants[0].size();
+    this->devstruct_.number_of_species_ = state_variables[0].size();
+    this->devstruct_.number_of_grid_cells_ = rate_constants.size();
 
-    std::chrono::nanoseconds kernel_duration = micm::cuda::AddForcingTermsKernelDriver(matrix, this->devstruct_);
-    return kernel_duration;  // time performance of kernel function
+    // Local pointer to the input matrix, whose data are already on the device
+    const double* d_rate_constants = rate_constants.AsDeviceParam().d_data_;
+    const double* d_state_variables = state_variables.AsDeviceParam().d_data_;
+    double* d_forcing = forcing.AsDeviceParam().d_data_;
+    micm::cuda::AddForcingTermsKernelDriver(d_rate_constants, d_state_variables, d_forcing, this->devstruct_);
   }
 
   template<template<class> class MatrixPolicy, template<class> class SparseMatrixPolicy>
--- a/include/micm/process/cuda_process_set.hpp
+++ b/include/micm/process/cuda_process_set.hpp
@@ -34,7 +34,8 @@ namespace micm
     requires VectorizableDense<MatrixPolicy<double>> void AddForcingTerms(
         const MatrixPolicy<double>& rate_constants,
         const MatrixPolicy<double>& state_variables,
-        MatrixPolicy<double>& forcing);
+        MatrixPolicy<double>& forcing)
+    const;
 
     template<template<class> class MatrixPolicy, template<class> class SparseMatrixPolicy>
     requires VectorizableDense<MatrixPolicy<double>> && VectorizableSparse<SparseMatrixPolicy<double>>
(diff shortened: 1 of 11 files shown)


include/micm/process/cuda_process_set.hpp
```diff
@@ -31,11 +31,10 @@ namespace micm
     void SetJacobianFlatIds(const SparseMatrix<double, OrderingPolicy>& matrix);
 
     template<template<class> typename MatrixPolicy>
-    requires VectorizableDense<MatrixPolicy<double>> std::chrono::nanoseconds AddForcingTerms(
+    requires VectorizableDense<MatrixPolicy<double>> void AddForcingTerms(
         const MatrixPolicy<double>& rate_constants,
         const MatrixPolicy<double>& state_variables,
-        MatrixPolicy<double>& forcing)
-    const;
+        MatrixPolicy<double>& forcing);
```
mattldawson, line 37: Why can this no longer be a `const` function?

sjsprecious: Thanks @mattldawson . I modified `this->devstruct` inside the `AddForcingTerms` function now (https://github.com/NCAR/micm/blob/update_cuda_class_for_addforcing/include/micm/process/cuda_process_set.hpp#L98-L100) and I thought I could not make it `const` anymore. If it is not the case or we can do it in the constructor instead, I am happy to put `const` back.

mattldawson: Ah, ok - I would probably set these values in the constructor. If you want, you can do an assert in this function that they are the correct values, but having these functions be `const` is how we ensure thread-safety for the `Rosenbrock::Solve()` function

mattldawson: Or you could also create a copy of `this->devstruct_` and modify that, but since `ProcessSet` is created for and only works with matrices of specific sizes, my vote would be to do it in the constructor.

test/unit/process/test_cuda_process_set.cpp
```diff
@@ -63,25 +70,37 @@ void testRandomSystemAddForcingTerms(std::size_t n_cells, std::size_t n_reaction
...
+    gpu_state.variables_.AsVector()[i] = cpu_state.variables_.AsVector()[i];
+  }
+  gpu_state.variables_.CopyToDevice();
 
-  MatrixPolicy<double> rate_constants{ n_cells, n_reactions };
-  for (auto& elem : rate_constants.AsVector())
+  CPUMatrixPolicy<double> cpu_rate_constants{ n_cells, n_reactions };
+  GPUMatrixPolicy<double> gpu_rate_constants{ n_cells, n_reactions };
+  for (auto& elem : cpu_rate_constants.AsVector())
     elem = get_double();
```
mwaxmonsky, line 87:
```suggestion
  std::ranges::generate(cpu_rate_constants.AsVector(), get_double);
```
Again, the for loop is perfectly fine but this might have other benefits behind the scenes using the generator.

test/unit/process/test_cuda_process_set.cpp
```diff
@@ -63,25 +70,37 @@ void testRandomSystemAddForcingTerms(std::size_t n_cells, std::size_t n_reaction
...
+  for (int i = 0; i < cpu_rate_constants.AsVector().size(); i++)
+  {
+    gpu_rate_constants.AsVector()[i] = cpu_rate_constants.AsVector()[i];
+  }
```
mwaxmonsky, line 91: Similar assignment comment from above.

test/unit/process/test_cuda_process_set.cpp
```diff
@@ -63,25 +70,37 @@ void testRandomSystemAddForcingTerms(std::size_t n_cells, std::size_t n_reaction
...
   }
 
-  micm::ProcessSet cpu_set{ processes, state.variable_map_ };
-  micm::CudaProcessSet gpu_set{ processes, state.variable_map_ };
+  micm::ProcessSet cpu_set{ processes, cpu_state.variable_map_ };
+  micm::CudaProcessSet gpu_set{ processes, gpu_state.variable_map_ };
 
-  for (auto& elem : state.variables_.AsVector())
+  for (auto& elem : cpu_state.variables_.AsVector())
     elem = get_double();
```
mwaxmonsky, line 77: Similar generate comment from below.

mwaxmonsky (commented):
Looks great Jian!  Just a few more comments that might be worth using but nothing blocking approval.  Just had the same comments Matt had about the const-ness of `AddForcingTerms` and once that's resolved, I'm happy to approve!

test/unit/process/test_cuda_process_set.cpp
sjsprecious: Done.

test/unit/process/test_cuda_process_set.cpp
sjsprecious: Done.

test/unit/process/test_cuda_process_set.cpp
sjsprecious: Done.

test/unit/process/test_cuda_process_set.cpp
```diff
@@ -63,25 +70,37 @@ void testRandomSystemAddForcingTerms(std::size_t n_cells, std::size_t n_reaction
...
+  for (int i = 0; i < cpu_state.variables_.AsVector().size(); i++)
+  {
+    gpu_state.variables_.AsVector()[i] = cpu_state.variables_.AsVector()[i];
+  }
```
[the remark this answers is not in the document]
sjsprecious: Thanks @mwaxmonsky for your suggestions. I wonder how you find out these cleaner implementations?

include/micm/process/cuda_process_set.hpp
sjsprecious: Thanks @mattldawson and @mwaxmonsky for your comments. I have updated the implementations to use the `CudaVectorMatrixParams` struct to convey those integer variables instead. Let me know if the new implementation looks better.

test/unit/process/test_cuda_process_set.cpp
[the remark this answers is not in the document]
mwaxmonsky: I have way too much free time and watch the `algorithm` header like a hawk.  :D
https://en.cppreference.com/w/cpp/algorithm
Thanks for taking a look at these Jian!

include/micm/process/cuda_process_set.hpp
```diff
@@ -31,7 +31,7 @@ namespace micm
     void SetJacobianFlatIds(const SparseMatrix<double, OrderingPolicy>& matrix);
 
     template<template<class> typename MatrixPolicy>
-    requires VectorizableDense<MatrixPolicy<double>> std::chrono::nanoseconds AddForcingTerms(
+    requires VectorizableDense<MatrixPolicy<double>> void AddForcingTerms(
         const MatrixPolicy<double>& rate_constants,
         const MatrixPolicy<double>& state_variables,
         MatrixPolicy<double>& forcing)
```
mattldawson, line 37:
```suggestion
    requires VectorizableDense<MatrixPolicy<double>> void AddForcingTerms(
        const MatrixPolicy<double>& rate_constants,
        const MatrixPolicy<double>& state_variables,
        MatrixPolicy<double>& forcing) const
```

include/micm/util/cuda_vector_matrix.hpp
```diff
@@ -151,6 +151,10 @@ namespace micm
     {
       return this->handle_;
     }
+    void SetNumberofGridCells(const std::size_t n)
+    {
+      this->vector_matrix_param_.number_of_grid_cells_ = n;
+    }
```
mattldawson, line 157: I would recommend having the `number_of_grid_cells_` be set during construction, similar to how `number_of_elements_` is set. Having a separate function that needs to be called to set the number of grid cells would let instances of this class be an invalid state until this function is called, which can lead to problems.

include/micm/util/cuda_param.hpp
```diff
@@ -109,7 +109,8 @@ struct LinearSolverParam
 struct CudaVectorMatrixParam
 {
   double* d_data_;
-  size_t num_elements_;
+  size_t number_of_elements_;
```
mattldawson, line 112: Yes!

mattldawson (changes requested):
looking great! just a couple minor suggestions

include/micm/process/cuda_process_set.hpp
sjsprecious: I think this is a Github issue. I have added the `const` back already (line 38).

include/micm/util/cuda_vector_matrix.hpp
sjsprecious: Thanks for the suggestion. Done.

include/micm/process/cuda_process_set.hpp
mattldawson: ah, sorry - I missed that

mattldawson (approved):
very nice!

PR #443, opened 2024-04-01
Rename CUDA matrix class to be Dense or Sparse

### LINKED ISSUE #442: The name of CUDA matrix class may be confusing
Currently we have `CudaVectorMatrix` and `CudaSparseMatrix` class defined for the GPU code. However, an object of type `CudaSparseMatrix` is also a vector matrix. Thus what we really mean for the GPU code is actually `CudaDenseVectorMatrix` and `CudaSparseVectorMatrix`.

After discussion with @mattldawson and @mwaxmonsky , we decide to rename all the `CudaVectorMatrix` to `CudaDenseMatrix` for clarity. We will also drop the term `Vector` for all the GPU code, assuming that we will always use a vectorized matrix for GPU computing.

sjsprecious:
This PR renames the previous `CudaVectorMatrix` class to `CudaDenseMatrix` class for clarity. Now we only have `CudaDenseMatrix` and `CudaSparseMatrix`, and they are only for the vector matrix on GPU.

Maybe we should highlight somewhere that the GPU code only works with the vector matrix?

All the 40 tests passed on Derecho's A100 GPU by using `nvhpc/23.7`.

Fixed #442

### DIFF
--- a/include/micm/process/cuda_process_set.cuh
+++ b/include/micm/process/cuda_process_set.cuh
@@ -21,15 +21,15 @@ namespace micm
     /// This is the host function that will call the CUDA kernel
     ///   to calculate the forcing terms
     void AddForcingTermsKernelDriver(
-        const CudaVectorMatrixParam& rate_constants_param,
-        const CudaVectorMatrixParam& state_variables_param,
-        CudaVectorMatrixParam& forcing_param,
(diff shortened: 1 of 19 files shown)


mattldawson (approved):
I really like the new names!

mwaxmonsky (approved):
Great work Jian!

PR #448, opened 2024-04-04
add support for non-double properties

### LINKED ISSUE #407: Read species properties from configuration data
Add the ability to read standard MICM properties for chemical species:
```
"molecular weight [kg mol-1]"
```
and allow users to decorate `Species` objects with custom information, using keys that start with `__`

**Ideas**
- standard properties (e.g., molecular weight) can be stored in `Species::properties_` map.
- for non-standard properties (which can be of any valid type in JSON/YAML), the parser can be extended to include a function that accepts a query parameter and returns a set of results with the standard or user-defined properties. Something like:
```
  micm::SolverParameters solver_params = solverConfig.GetSolverParams();

  auto chemical_system = solver_params.system_;
  auto reactions = solver_params.processes_;

  std::map<std::string, double> mol_wt = solverConfig.GetProperties("System.Phases.Species.molecular weight [kg mol-1]");
  std::map<std::string, bool> is_advected = solverConfig.GetProperties("System.Phases.Species.__is_advected");
```
with a `GetProperties` function something like:
```
template<class T>
std::map<std::string, T> SolverConfiguration::GetProperties(std::string query) const;
```

mattldawson:
Adds support for non-double species properties. Closes #407

### DIFF
--- a/include/micm/configure/solver_config.hpp
+++ b/include/micm/configure/solver_config.hpp
@@ -372,10 +372,11 @@ namespace micm
     {
       // required keys
       const std::string NAME = "name";
+      const std::string TYPE = "type";
 
       auto status = ValidateSchema(
           object,
-          { NAME, "type" },
+          { NAME, TYPE },
           { "tracer type", "absolute tolerance", "diffusion coefficient [m2 s-1]", "molecular weight [kg mol-1]" });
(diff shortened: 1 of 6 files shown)


K20shores (approved):

boulderdaze (approved):
Looks good!

PR #449, opened 2024-04-05
Initial attempt to format cuda code.

mwaxmonsky:
Adds version output and verbose file lists to determine which specific files are being found and forwarded to `clang-format`.

Added a job that will always succeed to show the formats for debugging purposes on each push.


mattldawson (approved):
nice!

K20shores (approved):

PR #457, opened 2024-04-10
456 read and use relative tolerance absolute tolerance

### LINKED ISSUE #456: Read and use relative tolerance, absolute tolerance
The configuration reader does not read the relative tolerance for the solver in a configuration. It reads the absolute tolerance for a species, but does not use it.

- read the relative tolerance and set it on the RosenbrockSolverParameters
- create an array of absolute tolerances that is the same size as the number of species., initialize it to 1e-3 (what it is now)
- for any species that has the absolute tolerance, set it in the array
- use this array when calculating the normalized error

K20shores:
Closes #456 

Reads the relative tolerance on the solver config and sets it on a set of parameters.
Uses the absolute tolerance for each species and sets it individually if it's present


include/micm/solver/rosenbrock_solver_parameters.hpp
```diff
@@ -51,7 +52,7 @@ namespace micm
     // Gamma_i = \sum_j  gamma_{i,j}
     std::array<double, 6> gamma_{};
 
-    double absolute_tolerance_{ 1e-3 };
+    std::vector<double> absolute_tolerance_;
```
sjsprecious, line 55: Is there a reason to set `absolute_tolerance` individually for each species but an overall `relative_tolerance` for everything?

test/integration/analytical_rosenbrock.cpp
```diff
@@ -222,7 +222,7 @@ TEST(AnalyticalExamples, Oregonator)
 
   auto params = micm::RosenbrockSolverParameters::six_stage_differential_algebraic_rosenbrock_parameters();
   params.relative_tolerance_ = 1e-4;
-  params.absolute_tolerance_ = 1e-6 * params.relative_tolerance_;
+  params.absolute_tolerance_ = std::vector<double>(3, 1e-6 * params.relative_tolerance_);
```
sjsprecious, line 225: I assume `3` here refers to three-stage Rosenbrock solver specifically?

include/micm/solver/rosenbrock_solver_parameters.hpp
K20shores: Only that the camp config doesn't yet support that.

src/solver/rosenbrock.cu
```diff
@@ -125,7 +125,7 @@ namespace micm
           if (g_tid < n)
           {
             d_ymax = max(fabs(d_y_old[g_tid]), fabs(d_y_new[g_tid]));
-            d_scale = atol + rtol * d_ymax;
+            d_scale = atol[g_tid] + rtol * d_ymax;
```
sjsprecious, line 128: I think this won't work because `g_tid` will loop through the whole vector length, which will be `number_of_grid_cells * number_of_species` here. My understanding is that `atol` only has the size of `number_of_species`?

src/solver/rosenbrock.cu
```diff
@@ -208,7 +208,7 @@ namespace micm
       if (tid < num_elements)
       {
         d_ymax = max(fabs(d_y_old[tid]), fabs(d_y_new[tid]));
-        d_scale = atol + rtol * d_ymax;
+        d_scale = atol[tid] + rtol * d_ymax;
```
sjsprecious, line 211: Same question above.

src/solver/rosenbrock.cu
K20shores: Well. The number of tolerances is the number of species, but then I duplicated those across all grid cells. So if we have three species with tolerances [(a, 0.1), (b, 0.2), (c, 0.3)] and are solving 3 grid cells, then the actual atol array that is passed in (or will be once I add it) will be [0.1, 0.2, 0.3, 0.1, 0.2, 0.3, 0.1, 0.2, 0.3]

So. I'm not sure what that means. This also means that we are ballooning memory on the CPU when we don't need to, but it seems like this thing could be beneficial for the GPU to reduce cache misses.

Either way, I need to get a release of the python code using these tolerances out to A&M before tomorrow. Should I just make an issue for this specific thing?

sjsprecious: Thanks Kyle. If `atol` is duplicated for all the grid cells, then there is no problem with this modification. But if we want to save memory for `atol`, we will need `g_tid % number_of_species` instead here.

include/micm/solver/rosenbrock_solver_parameters.hpp
sjsprecious: Thanks for the information.

include/micm/configure/solver_config.hpp
```diff
@@ -115,6 +120,13 @@ namespace micm
     // Error string
     std::stringstream last_json_object_;
 
+    // Constructor
+
+    JsonReaderPolicy(RosenbrockSolverParameters parameters)
+        : parameters_(parameters)
+    {
+    }
+
```
boulderdaze, line 129: Does it makes sense to make it `const ref` to avoid copy?
```suggestion
    JsonReaderPolicy(const RosenbrockSolverParameters& parameters)
        : parameters_(parameters)
    {
    }

```

include/micm/configure/solver_config.hpp
```diff
@@ -1106,6 +1138,10 @@ namespace micm
     ConfigParseStatus last_parse_status_ = ConfigParseStatus::None;
 
    public:
+
+    SolverConfig() : ConfigTypePolicy(RosenbrockSolverParameters::three_stage_rosenbrock_parameters()) {}
+    SolverConfig(RosenbrockSolverParameters parameters) : ConfigTypePolicy(parameters) {}
```
boulderdaze, line 1143: Same here
```suggestion
    SolverConfig(const RosenbrockSolverParameters& parameters) : ConfigTypePolicy(parameters) {}
```

boulderdaze (approved):

include/micm/solver/rosenbrock.inl
```diff
@@ -132,7 +157,35 @@ namespace micm
       for (auto& name : system.UniqueNames(state_reordering))
         variable_map[name] = index++;
     }
-    
+
+    MatrixPolicy<double> absolute_tolerances(parameters_.number_of_grid_cells_, variable_map.size(), 1e-3);
+    if (parameters_.absolute_tolerance_.size() != absolute_tolerances.AsVector().size())
```
mattldawson, line 162: Why are the sizes of these vectors being compared?

mattldawson (approved):
looks good! just one minor question

include/micm/solver/rosenbrock.inl
K20shores: in case someone set them on the parameters before the simulation because they want to override the values

src/solver/rosenbrock.cu
```diff
@@ -99,8 +99,9 @@ namespace micm
       double* d_y_new = y_new_param.d_data_;
       double* d_errors_input = devstruct.errors_input_;
       double* d_errors_output = devstruct.errors_output_;
-      const double atol = ros_param.absolute_tolerance_;
+      const double* atol = devstruct.absolute_tolerance_;
       const double rtol = ros_param.relative_tolerance_;
+      const size_t n_species = y_old_param.number_of_elements_;
```
sjsprecious, line 104:
```suggestion
      const size_t number_of_species = y_old_param.number_of_elements_;
```

src/solver/rosenbrock.cu
```diff
@@ -125,7 +126,7 @@ namespace micm
           if (g_tid < n)
           {
             d_ymax = max(fabs(d_y_old[g_tid]), fabs(d_y_new[g_tid]));
-            d_scale = atol + rtol * d_ymax;
+            d_scale = atol[g_tid % n_species] + rtol * d_ymax;
```
sjsprecious, line 129:
```suggestion
            d_scale = atol[g_tid % number_of_species] + rtol * d_ymax;
```

src/solver/rosenbrock.cu
```diff
@@ -199,16 +200,17 @@ namespace micm
       double* d_y_old = y_old_param.d_data_;
       double* d_y_new = y_new_param.d_data_;
       double* d_errors = devstruct.errors_input_;
-      double atol = ros_param.absolute_tolerance_;
+      double* atol = devstruct.absolute_tolerance_;
       double rtol = ros_param.relative_tolerance_;
       const size_t num_elements = devstruct.errors_size_;
+      const size_t n_species = y_old_param.number_of_elements_;
```
sjsprecious, line 206:
```suggestion
      const size_t number_of_species = y_old_param.number_of_elements_;
```

src/solver/rosenbrock.cu
```diff
@@ -199,16 +200,17 @@ namespace micm
...
 
       // Global thread ID
       size_t tid = blockIdx.x * BLOCK_SIZE + threadIdx.x;
       if (tid < num_elements)
       {
         d_ymax = max(fabs(d_y_old[tid]), fabs(d_y_new[tid]));
-        d_scale = atol + rtol * d_ymax;
+        d_scale = atol[tid % n_species] + rtol * d_ymax;
```
sjsprecious, line 213:
```suggestion
        d_scale = atol[tid % number_of_species] + rtol * d_ymax;
```

sjsprecious (approved):
Thanks @K20shores for working on this. Looks good to me and just a few minor consistency suggestions.

include/micm/configure/solver_config.hpp
```diff
@@ -72,16 +73,19 @@ namespace micm
...
     System system_;
     std::vector<Process> processes_;
+    RosenbrockSolverParameters parameters_;
 
-    SolverParameters(const System& system, std::vector<Process>&& processes)
+    SolverParameters(const System& system, std::vector<Process>&& processes, const RosenbrockSolverParameters&& parameters)
         : system_(system),
-          processes_(std::move(processes))
+          processes_(processes),
+          parameters_(parameters)
```
boulderdaze, line 81: My understanding is that `std::move` is necessary to actually move what `processes` owns to `processes_`. Could you confirm that it's not copy?

include/micm/configure/solver_config.hpp
```diff
@@ -72,16 +73,19 @@ namespace micm
...
     {
     }
 
-    SolverParameters(System&& system, std::vector<Process>&& processes)
-        : system_(std::move(system)),
-          processes_(std::move(processes))
+    SolverParameters(System&& system, std::vector<Process>&& processes, RosenbrockSolverParameters&& parameters)
+        : system_(system),
+          processes_(processes),
+          parameters_(parameters)
```
boulderdaze, line 88: Same question as the above

boulderdaze (commented):
Looks good! Interesting to see `three_stage_rosenbrock_param()` set as the default. Liked that it looks cleaner and I am curious if this is the most commonly used.

boulderdaze (changes requested):
Oops I just want to see my question for move operation confirmed

include/micm/configure/solver_config.hpp
K20shores: It isn't necessary here. `std::move` only attempts to create an rvalue reference, but it doesn't actually move anything. If we are in this constructor, then `parameters` is already an rvalue reference (`&&`).

mattldawson (approved):
looks good!

sjsprecious:
@K20shores can you merge in the latest main branch so that I could check whether all the GPU tests pass on Derecho? Thanks.

boulderdaze (approved):

mwaxmonsky (approved):
Looks great!  Good work @K20shores!

include/micm/solver/cuda_rosenbrock.hpp
```diff
@@ -66,6 +66,7 @@ namespace micm
       hoststruct.errors_size_ = parameters.number_of_grid_cells_ * system.StateSize();
       hoststruct.jacobian_diagonal_elements_ = this->state_parameters_.jacobian_diagonal_elements_.data();
       hoststruct.jacobian_diagonal_elements_size_ = this->state_parameters_.jacobian_diagonal_elements_.size();
+      hoststruct.absoluate_tolerance_ = parameters.absolute_tolerance_.data();
```
sjsprecious, line 69:
```suggestion
      hoststruct.absolute_tolerance_ = parameters.absolute_tolerance_.data();
```

include/micm/util/cuda_param.hpp
```diff
@@ -121,6 +121,7 @@ struct CudaRosenbrockSolverParam
   // for NormalizedError function
   double* errors_input_;
   double* errors_output_;
+  double* absolute_tolerance_;
```
sjsprecious, line 124:
```suggestion
  const double* absolute_tolerance_;
```

src/solver/rosenbrock.cu
```diff
@@ -199,16 +200,17 @@ namespace micm
       double* d_y_old = y_old_param.d_data_;
       double* d_y_new = y_new_param.d_data_;
       double* d_errors = devstruct.errors_input_;
-      double atol = ros_param.absolute_tolerance_;
+      double* atol = devstruct.absolute_tolerance_;
```
sjsprecious, line 203:
```suggestion
      const double* atol = devstruct.absolute_tolerance_;
```

sjsprecious (changes requested):
The `CudaNormalizedError` test failed on Derecho's A100 GPU. I guess the new absolute tolerance vector is not used correctly in the CUDA kernel.

sjsprecious (approved):
All the 41 tests pass on Derecho's A100 GPU.

PR #459, opened 2024-04-15
adding a test to show that integrated reaction rates can be collected

K20shores:


test/integration/integrated_reaction_rates.cpp
```diff
@@ -0,0 +1,66 @@
...
+  state.variables_[0] = concentrations;
+  state.conditions_[0].temperature_ = 273;
+  state.conditions_[0].pressure_ = 1000;
+
+  for (double t{}; t < 100; ++t)
+  {
+    std::cout << state.variables_[0][state.variable_map_.at(irr_1.name_)] << " " << state.variables_[0][state.variable_map_.at(irr_2.name_)] << std::endl;
+    auto result = solver.Solve(30.0, state);
+    EXPECT_GE(result.result_[0][state.variable_map_.at(irr_1.name_)], state.variables_[0][state.variable_map_.at(irr_1.name_)]);
+    EXPECT_GE(result.result_[0][state.variable_map_.at(irr_2.name_)], state.variables_[0][state.variable_map_.at(irr_2.name_)]);
```
sjsprecious, line 62: Could you explain what the `EXPECT_GE` here is checking?

K20shores: Integrated reaction rates are something that we report in music box interactive. It measures the amount of mass accumulated (produced) over time from a single equation, which is why these species only appear on the right side of the equation. It's greater than or equal specifically because it should be a nondecreasing quantity. To illustrate that, I can add a reaction like photolysis which would only produce mass during the day, but none at night

K20shores: this is testing if the new concentration (`result.result_`) is greater than or equal to the old concentration (`state.variables_`) specifically for the integrated reaction rates.

I added this test because the A&M students were seeing issues in their box model with integrated reaction rates and I wanted to be sure we were never somehow losing mass. I guess this isn't foolproof, but it's one way to show that we aren't losing mass (though we are creating it)

sjsprecious: Ah I see. Thanks for the clarification and it is clear to me now.

mattldawson (approved):
nice!

test/integration/integrated_reaction_rates.cpp
mwaxmonsky, line 62: Would it make sense to pull out the `state.variable_map_.at(irr_1.name_)` and `state.variable_map_.at(irr_2.name_)` into their own named variables?  It's not quite clear to me what they represent.

test/integration/integrated_reaction_rates.cpp
```diff
@@ -0,0 +1,66 @@
...
+    state.variables_ = result.result_;
+  }
+  std::cout << state.variables_[0][state.variable_map_.at(irr_1.name_)] << " " << state.variables_[0][state.variable_map_.at(irr_2.name_)] << std::endl;
```
mwaxmonsky, line 65: Similar comment as above.

mwaxmonsky (changes requested):
Looks great!  Just one clarification if possible.

mwaxmonsky (approved):
Looks great!  Thanks @K20shores!

test/integration/integrated_reaction_rates.cpp
```diff
@@ -0,0 +1,66 @@
...
+  auto state = solver.GetState();
+
+  std::vector<double> concentrations{ 1, 0, 0, 0, 0 };
+  state.variables_[0] = concentrations;
+  state.conditions_[0].temperature_ = 273;
+  state.conditions_[0].pressure_ = 1000;
+
+  for (double t{}; t < 100; ++t)
+  {
+    std::cout << state.variables_[0][state.variable_map_.at(irr_1.name_)] << " " << state.variables_[0][state.variable_map_.at(irr_2.name_)] << std::endl;
```
boulderdaze: Just curious, is `std::cout` meant to be in the test?

boulderdaze (approved):

sjsprecious (approved):

PR #464, opened 2024-04-18
Update the CUDA implementation of CudaLuDecomposition with the CUDA matrix class

sjsprecious:
This PR updates the input type for the CUDA implementation of `LuDecompose` function with the `CudaSparseMatrix` class.


mattldawson (approved):
very nice!

test/unit/solver/test_cuda_lu_decomposition.cpp
```diff
@@ -49,76 +50,84 @@ void check_results(
...
+  std::vector<double> cpu_L_vector = cpu_LU.first.AsVector();
+  std::vector<double> cpu_U_vector = cpu_LU.second.AsVector();
+  for (int i = 0; i < L_size; ++i)
+  {
+    EXPECT_EQ(gpu_L_vector[i], cpu_L_vector[i]);
+  };
+  for (int j = 0; j < U_size; ++j)
+  {
+    EXPECT_EQ(gpu_U_vector[j], cpu_U_vector[j]);
+  };
```
mwaxmonsky, line 106: Should these be `EXPECT_DOUBLE_EQ` or `EXPECT_NEAR` since they are technically `double`s?

mwaxmonsky (approved):
Looks great @sjsprecious!  Nothing worth holding it up, just a minor clarification if needed.

test/unit/solver/test_cuda_lu_decomposition.cpp
sjsprecious: Thanks. Revised as suggested.

PR #468, opened 2024-04-20
Updates to error handling

### LINKED ISSUE #466: Prepare error handling for MUSICA library
In preparation for implementing a solution for error handling in the MUSICA library, ensure that MICM consistently throws when errors are encountered, and use custom exceptions that can be caught and sent through the MUSICA C API in some way.

**Acceptance Criteria**
- All errors encountered in MICM throw custom exceptions

**Ideas**
- Explore `std::error_code` and `std::error_condition`

mattldawson:
Adds custom error codes to MICM classes and updates existing error handling to use them. The hope is that in the MUSICA library we can catch these errors and turn them into meaningful codes/messages that can be returned through the C API.

closes #466


K20shores (commented):
I like it! I guess one concern is that we are potentially masking system errors in our tests if they were ever to be thrown and happen to have the same numerical value as the micm error values. 

And on that note, in MUSICA, I don't see how we will be able to identify exactly which error was thrown since we have overlapping codes. It seems like we would need to know exactly which part of the code threw this so that we know which `enum class` to match the error code to. What am I missing?

mattldawson:
> I like it! I guess one concern is that we are potentially masking system errors in our tests if they were ever to be thrown and happen to have the same numerical value as the micm error values.
> 
> And on that note, in MUSICA, I don't see how we will be able to identify exactly which error was thrown since we have overlapping codes. It seems like we would need to know exactly which part of the code threw this so that we know which `enum class` to match the error code to. What am I missing?

Yeah, I was starting to work through that in the MUSICA library. One option might be to use a combination of the category name and the `int` value of the enum. Something like:
```
struct Error {
  int code_;
  const char* category_;
  const char* message_;
};
```
This way on the user side (even in Fortran, and I'm guessing Python), you could use `strcmp(error.category_, "MICM Matrix") == 0 && error.code_ == 2` to identify our `MicmMatrixErrc::InvalidVector` error. And this should avoid us masking other system errors. What do you think?

WardF:
I'm taking a look at this today; getting caught up on the conversation. In C, software error codes are typically negative, and system error codes are typically positive, to avoid collisions/confusion.  I'm about to review the PR changes so perhaps it's already in place, but in the netCDF project we have an `errors.h` file which we use to define all the errors in one place and easier developer reference.

K20shores:
> I'm taking a look at this today; getting caught up on the conversation. In C, software error codes are typically negative, and system error codes are typically positive, to avoid collisions/confusion. I'm about to review the PR changes so perhaps it's already in place, but in the netCDF project we have an `errors.h` file which we use to define all the errors in one place and easier developer reference.

I am a fan of having all of the errors in one place as well. It may mean a long file but it also means we don't need to have a specific error class for each header in micm. @mattldawson would you have anything against a single error file?

sjsprecious:
For the error handling, would it be better to add the `__FILE__` and `__LINE__` information so that a user can be directed to the exact location where the error occurs?

mwaxmonsky:
> For the error handling, would it be better to add the `__FILE__` and `__LINE__` information so that a user can be directed to the exact location where the error occurs?

If we do, we might be able to leverage the C++ library: https://en.cppreference.com/w/cpp/utility/source_location

mattldawson:
> For the error handling, would it be better to add the `__FILE__` and `__LINE__` information so that a user can be directed to the exact location where the error occurs?

My thinking is that the errors we return to API users should be clear enough for them to act upon without having to look through our source code. (Problems with configuration files, passing improperly sized arrays to API functions, etc.) If we're just interested in bug checking our code at runtime, we can do this with asserts instead of throwing exceptions. What does everyone think?

WardF (approved):
The functionality of this looks good, albeit I'm taking a 10,000 foot view.  My one thought is that the code would benefit from some comments; while a lot of it is self evident now, it might be less so down the road.  "Code needing more in-line comments" is hardly novel feedback, I acknowledge XD.  Particularly since it could apply to any number of my own PRs.  But since I'm looking at this as an outsider and not a core developer, the function is less immediately obvious to me, and the lack of comments stands out more.

mattldawson:
> > I'm taking a look at this today; getting caught up on the conversation. In C, software error codes are typically negative, and system error codes are typically positive, to avoid collisions/confusion. I'm about to review the PR changes so perhaps it's already in place, but in the netCDF project we have an `errors.h` file which we use to define all the errors in one place and easier developer reference.
> 
> I am a fan of having all of the errors in one place as well. It may mean a long file but it also means we don't need to have a specific error class for each header in micm. @mattldawson would you have anything against a single error file?

I like that too. The only thing that's a little unclear to me is how the dependencies would work. If the error codes live in MUSICA, would MICM and TUV-x depend on MUSICA? Or if they live in MICM and TUV-x, would MICM and TUV-x depend on each other?

WardF:
> > > I'm taking a look at this today; getting caught up on the conversation. In C, software error codes are typically negative, and system error codes are typically positive, to avoid collisions/confusion. I'm about to review the PR changes so perhaps it's already in place, but in the netCDF project we have an `errors.h` file which we use to define all the errors in one place and easier developer reference.
> > 
> > 
> > I am a fan of having all of the errors in one place as well. It may mean a long file but it also means we don't need to have a specific error class for each header in micm. @mattldawson would you have anything against a single error file?
> 
> I like that too. The only thing that's a little unclear to me is how the dependencies would work. If the error codes live in MUSICA, would MICM and TUV-x depend on MUSICA? Or if they live in MICM and TUV-x, would MICM and TUV-x depend on each other?

If each project maintained a list of errors, their own `errors.h`, or `<project>_errors.h` which could be included as part of error handling, would that avoid explicit dependencies between the projects for error handling? I feel like that's a naive question, but it's the first that springs to mind.

WardF:
> > For the error handling, would it be better to add the `__FILE__` and `__LINE__` information so that a user can be directed to the exact location where the error occurs?
> 
> My thinking is that the errors we return to API users should be clear enough for them to act upon without having to look through our source code. (Problems with configuration files, passing improperly sized arrays to API functions, etc.) If we're just interested in bug checking our code at runtime, we can do this with asserts instead of throwing exceptions. What does everyone think?

From experience, `ASSERT` is great when it works for the intended purpose, but when it leaks out into user-space (e.g. unexpected/unanticipated behavior results in an 'unreachable' `ASSERT` being reached by a user at runtime) it can become a bit of a pain. So I guess with careful planning, using an `ASSERT` for debugging purposes instead of FILE and LINE would work.  +100 to the idea of 'clear error messaging'.  

From a user support perspective, I wonder now about a user sending you details around a problem they had.  Would the error be enough for you to debug what's going on? Or would `__FILE__` and `__LINE__` be useful information to you then?  Just thinking out loud, clearly.

mattldawson:
> > > > I'm taking a look at this today; getting caught up on the conversation. In C, software error codes are typically negative, and system error codes are typically positive, to avoid collisions/confusion. I'm about to review the PR changes so perhaps it's already in place, but in the netCDF project we have an `errors.h` file which we use to define all the errors in one place and easier developer reference.
> > > 
> > > 
> > > I am a fan of having all of the errors in one place as well. It may mean a long file but it also means we don't need to have a specific error class for each header in micm. @mattldawson would you have anything against a single error file?
> > 
> > 
> > I like that too. The only thing that's a little unclear to me is how the dependencies would work. If the error codes live in MUSICA, would MICM and TUV-x depend on MUSICA? Or if they live in MICM and TUV-x, would MICM and TUV-x depend on each other?
> 
> If each project maintained a list of errors, their own `errors.h`, or `<project>_errors.h` which could be included as part of error handling, would that avoid explicit dependencies between the projects for error handling? I feel like that's a naive question, but it's the first that springs to mind.

I like that idea. Would you expect the error header to just define the `int` codes and the category strings? Something like:
```
#define ERR_CAT_MICM_MATRIX "MICM Matrix"
#define ERR_CODE_MICM_ROW_SIZE_MISMATCH 1
#define ERR_CODE_MICM_INVALID_VECTOR 2
```
and then update, e.g., the Matrix class to use them:
```
enum class MicmMatrixErrc
{
  RowSizeMismatch = ERR_CODE_MICM_ROW_SIZE_MISMATCH,
  InvalidVector = ERR_CODE_MICM_INVALID_VECTOR,
};

namespace
{
  class MicmMatrixErrorCategory : public std::error_category
  {
   public:
    const char *name() const noexcept override
    {
      return ERR_CAT_MICM_MATRIX;
    }
    ...
```
The only thing might be that we couldn't use the same label for specific errors in different categories. (like if INVALID_VECTOR is an error in another class like SparseMatrix)

mattldawson:
> The functionality of this looks good, albeit I'm taking a 10,000 foot view. My one thought is that the code would benefit from some comments; while a lot of it is self evident now, it might be less so down the road. "Code needing more in-line comments" is hardly novel feedback, I acknowledge XD. Particularly since it could apply to any number of my own PRs. But since I'm looking at this as an outsider and not a core developer, the function is less immediately obvious to me, and the lack of comments stands out more.

I have gone back and forth over time between 'comment everything' and 'try to write self-documenting code', and I don't know where I'm at right now. I'm happy to add more comments though. Was there anything in particular that you noticed should have more documentation?

mattldawson:
Another benefit to the single file for errors might be that if we just use CPP definitions, we can include the same header in the fortran wrapper code and maybe cut down on the passing back and forth of strings between c and fortran or keeping this list in two places.

mattldawson:
> > > For the error handling, would it be better to add the `__FILE__` and `__LINE__` information so that a user can be directed to the exact location where the error occurs?
> > 
> > 
> > My thinking is that the errors we return to API users should be clear enough for them to act upon without having to look through our source code. (Problems with configuration files, passing improperly sized arrays to API functions, etc.) If we're just interested in bug checking our code at runtime, we can do this with asserts instead of throwing exceptions. What does everyone think?
> 
> From experience, `ASSERT` is great when it works for the intended purpose, but when it leaks out into user-space (e.g. unexpected/unanticipated behavior results in an 'unreachable' `ASSERT` being reached by a user at runtime) it can become a bit of a pain. So I guess with careful planning, using an `ASSERT` for debugging purposes instead of FILE and LINE would work. +100 to the idea of 'clear error messaging'.
> 
> From a user support perspective, I wonder now about a user sending you details around a problem they had. Would the error be enough for you to debug what's going on? Or would `__FILE__` and `__LINE__` be useful information to you then? Just thinking out loud, clearly.

In going through the process of making these changes, I think it would be pretty clear what part of the code threw the error (the error categories are specific to one class, or in one case two classes that extend the same base class, and the individual errors are mostly only thrown in one place, or a few places at most).

To me what's important is that these errors always indicate a problem that the user should address, not one that indicates a bug in our code. If we also want to make it easier for users to report bugs in our code, what if we added something like an `InternalMicmErrc`? If we add some run-time bug check to our code that trips, it could throw this and include the file and line number of the detailed message. The user would get something like "Internal MICM Error(file, line): Please file a bug report..." What does everyone think of something like this?

sjsprecious:
> > > > For the error handling, would it be better to add the `__FILE__` and `__LINE__` information so that a user can be directed to the exact location where the error occurs?
> > > 
> > > 
> > > My thinking is that the errors we return to API users should be clear enough for them to act upon without having to look through our source code. (Problems with configuration files, passing improperly sized arrays to API functions, etc.) If we're just interested in bug checking our code at runtime, we can do this with asserts instead of throwing exceptions. What does everyone think?
> > 
> > 
> > From experience, `ASSERT` is great when it works for the intended purpose, but when it leaks out into user-space (e.g. unexpected/unanticipated behavior results in an 'unreachable' `ASSERT` being reached by a user at runtime) it can become a bit of a pain. So I guess with careful planning, using an `ASSERT` for debugging purposes instead of FILE and LINE would work. +100 to the idea of 'clear error messaging'.
> > From a user support perspective, I wonder now about a user sending you details around a problem they had. Would the error be enough for you to debug what's going on? Or would `__FILE__` and `__LINE__` be useful information to you then? Just thinking out loud, clearly.
> 
> In going through the process of making these changes, I think it would be pretty clear what part of the code threw the error (the error categories are specific to one class, or in one case two classes that extend the same base class, and the individual errors are mostly only thrown in one place, or a few places at most).
> 
> To me what's important is that these errors always indicate a problem that the user should address, not one that indicates a bug in our code. If we also want to make it easier for users to report bugs in our code, what if we added something like an `InternalMicmErrc`? If we add some run-time bug check to our code that trips, it could throw this and include the file and line number of the detailed message. The user would get something like "Internal MICM Error(file, line): Please file a bug report..." What does everyone think of something like this?

Ah, I see. I am thinking about the code bug and I want the error message to be direct and pointed to the specific source code for faster debugging later. But regarding a user error, what you and others propose here makes sense to me, and using an error header file with enumed error code sounds a good idea. Thanks for your clarification.

boulderdaze:
> > I like it! I guess one concern is that we are potentially masking system errors in our tests if they were ever to be thrown and happen to have the same numerical value as the micm error values.
> > And on that note, in MUSICA, I don't see how we will be able to identify exactly which error was thrown since we have overlapping codes. It seems like we would need to know exactly which part of the code threw this so that we know which `enum class` to match the error code to. What am I missing?
> 
> Yeah, I was starting to work through that in the MUSICA library. One option might be to use a combination of the category name and the `int` value of the enum. Something like:
> 
> ```
> struct Error {
>   int code_;
>   const char* category_;
>   const char* message_;
> };
> ```
> 
> With this to turn `std::system_error` into an `Error`:
> 
> ```
> Error ToError(const std::system_error& e)
> {
>     Error error;
>     error.code_ = e.code().value();
>     error.category_ = e.code().category().name();
>     error.message_ = e.what();
>     return error;
> }
> ```
> 
> This way on the user side (even in Fortran, and I'm guessing Python), you could use `strcmp(error.category_, "MICM Matrix") == 0 && error.code_ == 2` to identify our `MicmMatrixErrc::InvalidVector` error. And this should avoid us masking other system errors. What do you think?

I like this idea

boulderdaze (approved):
Looks good! Many interesting discussions above. I am not very familiar with the `system_error` header but we might be able to create a macro to avoid having to create each error category class to the applicable classes? something like, 
`MICM_ERROR(category, error code, message)`
Just a thought, thanks for the work! I think it's quite tricky to design the error handler since we also have to consider MUSICA.

mattldawson:
Based on the discussion, I'll go ahead and create a micm error header and add an internal error category that includes the filename and line number in the message. (I'm not sure if we actually have any run-time checks that I can apply this to, but if I can't find one I'll just add a test that throws this error.)

I'll request a re-review once that's ready, but let me know if you would prefer not to review this again and I won't include you in the re-review.

Thanks for the discussion everyone!

include/micm/configure/solver_config.hpp
```diff
@@ -21,52 +22,98 @@
...
 #include <micm/system/system.hpp>
 #include <micm/util/constants.hpp>
+#include <micm/util/error.hpp>
 #include <nlohmann/json.hpp>
 #include <sstream>
 
-namespace micm
+enum class MicmConfigErrc
+{
+  Success = MICM_CONFIGURATION_ERROR_CODE_SUCCESS,
```
K20shores, line 31: Now that we have a better error system, should we think about removing the success status since it's not an error? Rather than checking the return value throughout the configuration, we can throw an error when we arrive at one

include/micm/util/error.hpp
```diff
@@ -0,0 +1,54 @@
...
+#define MICM_MATRIX_ERROR_CODE_ROW_SIZE_MISMATCH 1
+#define MICM_MATRIX_ERROR_CODE_INVALID_VECTOR 2
+#define MICM_MATRIX_ERROR_CODE_ELEMENT_OUT_OF_RANGE 3
+#define MICM_MATRIX_ERROR_CODE_MISSING_BLOCK_INDEX 4
+#define MICM_MATRIX_ERROR_CODE_ZERO_ELEMENT_ACCESS 5
+
+#define MICM_ERROR_CATEGORY_INTERNAL "MICM Internal Error"
+#define MICM_INTERNAL_ERROR_CODE_GENERAL 1
+#define MICM_INTERNAL_ERROR_CODE_CUDA 2
+#define MICM_INTERNAL_ERROR_CODE_CUBLAS 3
```
K20shores, line 54: I like the categories, but I still worry that we could end up in a situation (not sure how), that we only have the integer value. What do #defines gain us over `static const strings` and one large `enum class` (for which we don't manually define the values)?

include/micm/util/error.hpp
```diff
@@ -0,0 +1,54 @@
+#if 0
+// Copyright (C) 2023-2024 National Center for Atmospheric Research,
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// This file defines error categories and codes for MICM
+#endif
```
K20shores, line 7: why this instead of #pragma once?

K20shores (commented):
Really like this. I left other comments but none of them would prevent this from being merged

K20shores (approved):

include/micm/configure/solver_config.hpp
mattldawson: Good point. I'll do that now.

include/micm/util/error.hpp
mattldawson: If we include this in a fortran file, any comments would have to be flagged with `!` instead of `//`. This was just to get the CPP to remove these lines entirely

include/micm/solver/jit_linear_solver.inl
```diff
@@ -41,7 +41,7 @@ namespace micm
     solve_function_ = NULL;
     if (matrix.Size() != L || matrix.GroupVectorSize() != L)
     {
-      throw std::runtime_error("Invalid matrix for JitLinearSolver. Check the the VectorMatrix template parameters.");
+      throw std::system_error(make_error_code(MicmJitErrc::InvalidMatrix), "");
```
mwaxmonsky, line 44: Might be worth having a message why it's invalid?

include/micm/solver/jit_lu_decomposition.inl
```diff
@@ -35,7 +35,7 @@ namespace micm
     decompose_function_ = NULL;
     if (matrix.Size() > L)
     {
-      throw std::runtime_error("Invalid matrix for JitLuDecomposition. Check the the VectorMatrix template parameters.");
+      throw std::system_error(make_error_code(MicmJitErrc::InvalidMatrix), "");
```
mwaxmonsky, line 38: Might be worth having a message why it's invalid?

include/micm/util/error.hpp
mattldawson: I could go either way. There is something about having one big bucket for things that makes me nervous, but I can't actually think of a real example of how this would be a problem in this case.

The one argument for `#defines` is that  it allows this file to be included in fortran files as well, and might avoid some duplication or difficult passing of strings, but this isn't a super big deal.

I guess, my suggestion would be to keep the categories until we find ourselves in a situation where we only have the integer code for some reason and decide then if we should reorganize. What do you think?

include/micm/solver/jit_rosenbrock.hpp
```diff
@@ -87,7 +87,7 @@ namespace micm
       MatrixPolicy<double> temp{};
       if (temp.GroupVectorSize() != parameters.number_of_grid_cells_)
       {
-        throw std::runtime_error("Number of grid cells for JitRosenbrockSolver must match template parameter.");
+        throw std::system_error(make_error_code(MicmJitErrc::InvalidMatrix), "");
```
mwaxmonsky, line 90: Might be worth having a message why it's invalid?

include/micm/solver/jit_rosenbrock.hpp
```diff
@@ -113,7 +113,7 @@ namespace micm
       }
       else
       {
-        throw std::runtime_error("Failed to generate the alpha minus jacobia JIT function.");
+        throw std::system_error(make_error_code(MicmJitErrc::MissingJitFunction), "AlphaMinusJacobian");
```
mwaxmonsky, line 116:
```suggestion
        throw std::system_error(make_error_code(MicmJitErrc::MissingJitFunction), std::source_location::current().function_name());
```

include/micm/util/matrix.hpp
```diff
@@ -50,7 +52,7 @@ namespace micm
         // check that this row matches the expected rectangular matrix dimensions
         if (other.size() < y_dim_)
         {
-          throw std::runtime_error("Matrix row size mismatch in assignment from vector");
+          throw std::system_error(make_error_code(MicmMatrixErrc::RowSizeMismatch), "");
```
mwaxmonsky, line 55: Might be worth having a message why it's mismatched for easier debugging?

include/micm/util/matrix.hpp
```diff
@@ -163,7 +165,7 @@ namespace micm
                   // check that this row matches the expected rectangular matrix dimensions
                   if (other[x].size() != y_dim)
                   {
-                    throw std::runtime_error("Invalid vector for matrix assignment");
+                    throw std::system_error(make_error_code(MicmMatrixErrc::InvalidVector), "");
```
mwaxmonsky, line 168: Might be worth having a message why it's invalid?

include/micm/util/vector_matrix.hpp
```diff
@@ -55,7 +100,7 @@ namespace micm
       {
         if (other.size() < y_dim_)
         {
-          throw std::runtime_error("Matrix row size mismatch in assignment from vector.");
+          throw std::system_error(make_error_code(MicmVectorMatrixErrc::RowSizeMismatch), "");
```
mwaxmonsky, line 103: Might be worth having a message why it's mismatched for easier debugging?

include/micm/util/vector_matrix.hpp
```diff
@@ -173,7 +218,7 @@ namespace micm
                 {
                   if (other_row.size() != y_dim)
                   {
-                    throw std::runtime_error("Invalid vector for matrix assignment");
+                    throw std::system_error(make_error_code(MicmVectorMatrixErrc::InvalidVector), "");
```
mwaxmonsky, line 221: Might be worth having a message why it's invalid?

src/solver/rosenbrock.cu
```diff
@@ -247,7 +288,7 @@ namespace micm
 
       if (number_of_elements != errors_param.number_of_elements_)
       {
-        throw std::runtime_error("devstruct.errors_input_ and errors_param have different sizes.");
+        throw std::system_error(make_error_code(MicmCudaRosenbrockSolverErrc::ParameterMismatch), "");
```
mwaxmonsky, line 291: Might be worth having a message why it's mismatched for easier debugging?

src/solver/rosenbrock.cu
```diff
@@ -263,7 +304,7 @@ namespace micm
         if (stat != CUBLAS_STATUS_SUCCESS)
         {
           printf(cublasGetStatusString(stat));
-          throw std::runtime_error("Error while calling cublasDnrm2.");
+          throw std::system_error(make_error_code(MicmCudaRosenbrockSolverErrc::CublasError), "");
```
mwaxmonsky, line 307:
```suggestion
          throw std::system_error(make_error_code(MicmCudaRosenbrockSolverErrc::CublasError), "cublasStatus_t came back as " + stat);
```
Just to provide the actual error code to the user.

mwaxmonsky (approved):
Looks great @mattldawson!

Nothing worth holding this up but we might want to expand on some of the error messages if it makes sense and use the std library to get the function name in one place.

include/micm/solver/jit_rosenbrock.hpp
K20shores: that's a neat trick

src/solver/rosenbrock.cu
K20shores: if that's the case should the printf the line above be removed?

mwaxmonsky: I would say yes since I don't think we need to print to stdout anymore if there's an error since we are throwing the exception now.

include/micm/solver/jit_rosenbrock.hpp
mwaxmonsky: Oh, this requires the `#include <source_location>` header.

include/micm/configure/solver_config.hpp
mattldawson: done!

include/micm/solver/jit_lu_decomposition.inl
mattldawson: done!

include/micm/solver/jit_rosenbrock.hpp
mattldawson: done!

include/micm/solver/jit_rosenbrock.hpp
mattldawson: pretty cool! done!

include/micm/util/matrix.hpp
mattldawson: done!

include/micm/util/matrix.hpp
mattldawson: done!

include/micm/util/vector_matrix.hpp
mattldawson: done!

include/micm/util/vector_matrix.hpp
mattldawson: done!

src/solver/rosenbrock.cu
mattldawson: I ended up switching this to use the new `INTERNAL_ERROR()` because it looked to me more like a check for a bug in our code. But, if that's not correct, I can change it back to throw one of our user errors

src/solver/rosenbrock.cu
mattldawson: yeah, I agree - I try to remove output to stdout and stderr when I find them. This should be updated now.

mwaxmonsky (approved):
Thanks for addressing those comments @mattldawson!

WardF:
Reviewing now as well, although I'm sure I'll be the slowest due to relative unfamiliarity; I don't anticipate finding any issues, but I'm trying to digest what I'm looking at instead of skimming it :)

include/micm/util/error.hpp
boulderdaze: How about not including copyright for this specific file? I think that makes sense that we don't want to duplicate the error file for fortran but it could also be confusing for MICM being standalone CPP project to have this unusual usage of macro to cater for fortran. What do you think?

WardF (approved):

boulderdaze (approved):
Looks great! Thank you for the work! I have one suggestion but it shouldn't prevent this branch from being merged.

mattldawson:
I agree about not making things confusing just to cater to fortran, but I think we should try to have the copyright and license on each source code file. What if we revisit this after the fortran wrapper is updated? It could be the case that we don't even need it in fortran and can just use normal comments.

boulderdaze:
> > I agree about not making things confusing just to cater to fortran, but I think we should try to have the copyright and license on each source code file. What if we revisit this after the fortran wrapper is updated? It could be the case that we don't even need it in fortran and can just use normal comments.
> 
> (I added this to our post-development review issue #211)

Sounds good, thank you!

PR #474, opened 2024-04-25
Update the CUDA implementation of CudaLinearSolver with the CUDA matrix class

sjsprecious:
This PR updates the input type for the CUDA implementation of `LinearSolver` function with the `CudaSparseMatrix` and `CudaDenseMatrix` class.

All the 42 tests all passed on Derecho's A100 GPU by using `nvhpc/23.7`.

PR #483, opened 2024-04-30
Explicitly call the function from the base class in a CUDA class

sjsprecious:
This PR makes all the CUDA class explicitly call the functions from the base class if the input argument is not a CUDA matrix. This helps confuse the compiler less.

Since this branch is based on the `update_cuda_matrix_class_for_cuda_linear_solver` branch, there are lots of duplicated changes from PR #474 before that PR is merged.


mattldawson (approved):

mwaxmonsky (approved):
Great work @sjsprecious !

Issue #458, opened 2024-04-12
Need a test for movement assignment and constructor


sjsprecious:
While working on the PR #452 , Matt discovered that in the movement assignment and constructor, we needed to manually reset `other.param_.d_data_ = nullptr` after the move operator. Otherwise, the `cudaFree` will be performed twice for the same device data and an error will be triggered later.

It will be better to add a test for this so that we won't be trapped by this again.

PR #484, opened 2024-04-30
Check if the device pointer is initialized to nullptr in the move constructor

sjsprecious:
This PR fixes the issue found in #458 based on my knowledge but I am happy to check more things if they are missing.

If I misunderstand the issue somehow, please let me know as well.


mattldawson (approved):

mwaxmonsky (commented):
Looks great @sjsprecious !!

@mattldawson , would it make sense to implement these using the error handling that came in from https://github.com/NCAR/micm/commit/0212b857ed4ff53691a30975729d5572fa645624?

mwaxmonsky (approved):
@sjsprecious , apologies, didn't recognize this was at the test level.  Looks good to me!

PR #485, opened 2024-05-01
400 backward euler solver option

### LINKED ISSUE #400: Backward Euler solver option
Add an option either to the Rosenbrock solver or a new ODE solver type that can recreate the MOZART Backward Euler solver

**Acceptance Criteria**
- Users can create an instance of a MICM solver whose solver function follows the same logic as the CAM-Chem MOZART solver

K20shores:
Closes #400 

- adds begin, end iterators to the matrix class so that we can do math on any type of matrix, needs to be added to vectorized matrix
-


include/micm/solver/backward_euler.inl
```diff
@@ -0,0 +1,173 @@
...
+
+    Process::UpdateState(processes_, state);
+
+    bool singular = false;
+
+    while(t < time_step) {
+
+      bool converged = true;
+
+      do {
```
sjsprecious, line 100: I think we are missing the iteration for the Newton method here. See the Fortran implementation at https://github.com/ESCOMP/CAM/blob/cam_development/src/chemistry/pp_trop_strat_mam4_vbs/mo_imp_sol.F90#L259.

K20shores: oh hey we are. Thanks for that

include/micm/solver/backward_euler.inl
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  enum class LastTimeStepCut
+  {
+    None,
+    Half,
+    Tenth
+  };
+
+  inline void BackwardEuler::Solve(double time_step, auto state, auto linear_solver, auto process_set)
```
sjsprecious, line 67: Probably we can not do a strict apple-to-apple comparison here. But CAM-Chem calculates a chunk of data for the chemistry solver (https://github.com/ESCOMP/CAM/blob/cam_development/src/chemistry/pp_trop_strat_mam4_vbs/mo_imp_sol.F90#L198) rather than all the grid cells at once, which is done in MICM right now.

include/micm/solver/backward_euler.inl
```diff
@@ -0,0 +1,205 @@
...
+
+        // always make sure the solution is positive regardless of which iteration we are on
+        // remove negatives, accept this solution and continue
+        yn1_iter = Yn1.begin();
+        yn_iter = Yn.begin();
+        for(; yn1_iter != Yn1.end(); ++yn1_iter, ++yn_iter) {
+          *yn_iter = *yn1_iter = std::max(0.0, *yn1_iter);
+        }
+
+        if (iterations == 0) {
```
sjsprecious, line 147: We need to increase `iterations` first. Otherwise it is never updated.

include/micm/solver/backward_euler.inl
```diff
@@ -0,0 +1,205 @@
...
+    std::size_t cut_in_tenth_limit = 1;
+    const std::array<double, 5> time_step_reductions{ 0.5, 0.5, 0.5, 0.5, 0.1 };
+
+    auto Yn = state.variables_;
+    auto Yn1 = state.variables_;
+
+    Process::UpdateState(processes, state);
+
+    while(t < time_step) {
+      bool converged = true;
```
sjsprecious, line 86: I think we need to set `converged = false` instead. Otherwise after the first iteration, we will always fail the condition `!converged && iterations < max_iter` (because we do not check the convergence for the first iteration) and no second iteration will be calculated.

include/micm/solver/backward_euler.inl
```diff
@@ -0,0 +1,205 @@
...
+      std::size_t iterations = 0;
+
```
sjsprecious, line 88: Before the Newton iteration begins, we should define something like below:
```
      auto iter_invariant = Yn;
      for(auto& elem : iter_invariant.AsVector())
      {
        elem /= H;
      }
```
This refers to the constant term when calculating the residual, and it should not change between different Newton iterations.

include/micm/solver/backward_euler.inl
```diff
@@ -0,0 +1,205 @@
...
+        auto yn1_iter = Yn1.begin();
+        auto yn_iter = Yn.begin();
+        auto residual_iter = residual.begin();
+        auto forcing_iter = forcing.begin();
+        // forcing_blk in camchem
+        // residual = (Yn1 - Yn) / H - forcing;
+        for(; yn1_iter != Yn1.end(); ++yn1_iter, ++yn_iter, ++residual_iter, ++forcing_iter) {
+          *residual_iter = (*yn1_iter - *yn_iter) / H - *forcing_iter;
+        }
```
sjsprecious, line 124: This should be replaced by:
```
        for(; yn_iter != Yn.end(); ++yn_iter, ++iter_invariant_iter, ++residual_iter, ++forcing_iter) {
          *residual_iter = *yn_iter / H - *iter_invariant_iter - *forcing_iter;
        }
```
Both `yn1` and `yn` are updated during the Newton iterations, but we want to subtract a constant term instead here.

include/micm/solver/backward_euler.inl
sjsprecious: Something like:
```
        ++iterations;
        if (iterations == 1) {
          // we always want to accept the first iteration
          continue;
        }
```

include/micm/solver/backward_euler.inl
```diff
@@ -0,0 +1,205 @@
...
+        // is less than a tolerance times the absolute value of the solution
+        for(; temp_iter != temp.end(); ++temp_iter, ++yn1_iter) {
+          // changes that are much smaller than the tolerance are negligible and we assume can be accepted
+          if (std::abs(*temp_iter) > small) {
+            if (std::abs(*temp_iter) > tol * std::abs(*yn1_iter)) {
+              converged = false;
+              break;
+            }
+          }
+        }
```
sjsprecious, line 167: I would change it to:
```
        for(; temp_iter != temp.end(); ++temp_iter, ++yn1_iter) {
          if (std::abs(*temp_iter) > small) {
            if (std::abs(*temp_iter) > tol * std::abs(*yn1_iter)) {
              converged = false;
              break;
            }
            else { converged = true; }
          }
          else { converged = true; }
        }
```

include/micm/solver/backward_euler.inl
```diff
@@ -0,0 +1,205 @@
...
+        H *= time_step_reductions[n_convergence_failures++];
+      }
+      else {
+        t += H;
+
+        // when we accept two solutions in a row, we can increase the time step
+        n_successful_integrations++;
+        n_convergence_failures = 0;
+        if (n_successful_integrations >= 2) {
+          H *= 2.0;
```
sjsprecious, line 195: May need to set `n_successful_integrations = 0` after doubling the time step.

include/micm/solver/backward_euler.inl
```diff
@@ -0,0 +1,205 @@
...
+        };
+
+        H *= time_step_reductions[n_convergence_failures++];
+      }
+      else {
+        t += H;
+
+        // when we accept two solutions in a row, we can increase the time step
+        n_successful_integrations++;
+        n_convergence_failures = 0;
```
sjsprecious, line 193: CAM-Chem does not reset this value though.

include/micm/solver/backward_euler.inl
sjsprecious: Also we need to save the `Yn` value here for substepping. Something like:
```
auto yn_bk = Yn;
```

include/micm/solver/backward_euler.inl
```diff
@@ -0,0 +1,205 @@
...
+          // we have failed to converge, accept the solution
+          // TODO: continue on with the current solution to get the full solution
+          n_convergence_failures = 0;
+          // give_up = true;
+          t += H;
+          break;
+          throw std::system_error(make_error_code(MicmBackwardEulerErrc::FailedToConverge), "Failed to converge");
+        };
+
+        H *= time_step_reductions[n_convergence_failures++];
```
sjsprecious, line 186: After doing the time substepping here, we need to reset `Yn` as:
```
          Yn = yn_bk;
```
since `Yn` has already been updated during this iteration.

sjsprecious (commented):
I made a couple of suggestions that I think are needed to be consistent with the CAM-Chem implementation. However, those changes do not seem to affect the results. Hmm....

include/micm/solver/backward_euler.inl
K20shores: We solve multiple grid cells all at once, just the same as our Rosenbrock implementation. If we need to solve chunks of grid cells, I think we should push to have that logic outside of the solve routine. The camchem implementation is complicated and the readability gained from not handling chunked solving I think is far more important than the small computational costs we'll have from stack frames added without a for loop on the outside of the solve function that loops over chunks

include/micm/solver/backward_euler.inl
K20shores: Ah, true. I have done that on my machine now

include/micm/solver/backward_euler.inl
K20shores: good catch

include/micm/solver/backward_euler.inl
K20shores: I don't see how this is better. We only need to find a single species that didn't converge. The else's are superfluous when we set `converged = true;` before the loop

include/micm/solver/backward_euler.inl
K20shores: Since it's `>= 2` I don't think that will have any logical difference; this is the only place that the value of `n_successful_integrations` matters

include/micm/solver/backward_euler.inl
K20shores: Is this the same thing as reseting Yn to the initial value for this timestep?

sjsprecious: I think so.

include/micm/solver/backward_euler.inl
```diff
@@ -0,0 +1,206 @@
...
+    auto Yn1 = state.variables_;
+
+    Process::UpdateState(processes, state);
+
+    while(t < time_step) {
+      // std::cout << "H: " << H << " t: " << t << std::endl;
+      bool converged = false;
+      std::size_t iterations = 0;
+
+      auto original_Yn = Yn;
```
sjsprecious, line 88: with the new implementation, we probably do not need `original_Yn` anymore.

include/micm/solver/backward_euler.inl
```diff
@@ -0,0 +1,206 @@
...
+          // we have failed to converge, accept the solution
+          // TODO: continue on with the current solution to get the full solution
+          n_convergence_failures = 0;
+          // give_up = true;
+          t += H;
+          throw std::system_error(make_error_code(MicmBackwardEulerErrc::FailedToConverge), "Failed to converge");
+          break;
+        };
+
+        Yn = original_Yn;
```
sjsprecious, line 185: We do not need this line since `Yn` is not updated if the solution does not converge.

a.out
sjsprecious, line 1: Probably we should remove this `a.out` file?

thing.cpp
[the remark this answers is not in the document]
K20shores: hahaha no we do not. That was just me messing around with auto

a.out
K20shores: yeah, that's true

include/micm/solver/backward_euler.inl
```diff
@@ -0,0 +1,204 @@
...
+    std::size_t n_convergence_failures = 0;
+    const std::array<double, 5> time_step_reductions{ 0.5, 0.5, 0.5, 0.5, 0.1 };
+
+    auto Yn = state.variables_;
+    auto Yn1 = state.variables_;
+
+    Process::UpdateState(processes, state);
+
+    while(t < time_step) {
+      // std::cout << "H: " << H << " t: " << t << std::endl;
```
mattldawson, line 84:
```suggestion
```

mattldawson (approved):
very nice!

include/micm/solver/backward_euler.hpp
```diff
@@ -0,0 +1,55 @@
...
+        const std::vector<Process>& processes);
+
+    virtual ~BackwardEuler() = default;
+
+    /// @brief Advances the given step over the specified time step
+    /// @param time_step Time [s] to advance the state by
+    /// @return A struct containing results and a status code
+    void Solve(double time_step, auto& state, auto linear_solver, auto process_set, const std::vector<micm::Process>& processes, auto jacobian_diagonal_elements);
+  };
+
```
boulderdaze, line 52: Could you update the comments for `BackwardEuler`?` I'm in favor of specifying types in the function declaration for readability. I'd love to hear what you and the team think!

include/micm/solver/backward_euler.inl
```diff
@@ -0,0 +1,204 @@
...
+        linear_solver.Solve(residual, temp, state.lower_matrix_, state.upper_matrix_);
+
+        // solution_blk in camchem
+        // Yn1 = Yn1 + temp;
+        // always make sure the solution is positive regardless of which iteration we are on
+        // remove negatives, accept this solution and continue
+        auto temp_iter = temp.begin();
+        yn1_iter = Yn1.begin();
+        for(; temp_iter != temp.end(); ++temp_iter, ++yn1_iter) {
+          *yn1_iter = std::max(0.0, *yn1_iter + *temp_iter);
```
sjsprecious, line 141: @K20shores @mattldawson According to the implementation of CAM-Chem, we probably could get rid of the `temp` and `residual` vectors here and just use the `forcing` vector. This may be an optimization step or only applicable to the CAM-Chem version.

include/micm/solver/backward_euler.inl
```diff
@@ -0,0 +1,204 @@
...
+    // A fully implicit euler implementation is given by the following equation:
+    // y_{n+1} = y_n + H * f(t_{n+1}, y_{n+1})
+    // This is a root finding problem because you need to know y_{n+1} to compute f(t_{n+1}, y_{n+1})
+    // you need to solve the equation y_{n+1} - y_n - H f(t_{n+1}, y_{n+1}) = 0
+    // We will also use the same logic used by cam-chem to determine the time step
+    // That scheme is this: 
+    // Start with H = time_step
+    // if that fails, try H = H/2 several time
+    // if that fails, try H = H/10 once
+    // if that fails, accept the current integration and move on. Continue on with the current H until the end of the time step
```
sjsprecious, line 63: Since I forgot this all the time, can you please add a sentence saying **We accept the current H but do not update the Yn vector**?

sjsprecious (approved):
Thanks @K20shores for your efforts on implementing the backward Euler solver. It is very nice and helpful for our future performance evaluation.

Just have two minor comments but they shouldn't hold the merge of this PR.

test/unit/solver/test_backward_euler.cpp
```diff
@@ -0,0 +1,52 @@
...
+    std::vector<micm::Process>{ r1, r2 },
+    options
+  };
+
+  auto state = rosenbrock.GetState();
+  auto linear_solver = rosenbrock.linear_solver_;
+  auto process_set = rosenbrock.process_set_;
+  auto processes = std::vector<micm::Process>{ r1, r2 };
+
+  EXPECT_NO_THROW(be.Solve(time_step, state, linear_solver, process_set, processes, rosenbrock.state_parameters_.jacobian_diagonal_elements_));
```
boulderdaze, line 51: I am a bit confused to see rosenbrock solver in the test for backward euler. Would it be still in-progress or am I missing something?

boulderdaze (changes requested):
Looks good! I like that you removed the template for the type `double`, cleaner! I just have a couple minor change requests, thanks!

include/micm/solver/backward_euler.hpp
K20shores: I'm intentionally using auto so that we don't have to put template everywhere.

include/micm/solver/backward_euler.inl
K20shores: done

include/micm/solver/backward_euler.hpp
K20shores: This isn't a finalized implementation. We need to split out the logic of constructing what we need to solve so that it's easy to use both rosenbrock and backward euler, but that will require some work so I left it for a future issue. Maybe at that point we will add some explicit types here but using auto means caring less about the template stuff

include/micm/solver/backward_euler.hpp
```diff
@@ -0,0 +1,55 @@
...
+    /// @brief Default constructor
+    BackwardEuler();
+
+    /// @brief Builds a Rosenbrock solver for the given system, processes, and solver parameters
+    /// @param system The chemical system to create the solver for
+    /// @param processes The collection of chemical processes that will be applied during solving
+    /// @param parameters Rosenbrock algorithm parameters
+    BackwardEuler(
+        const System& system,
+        const std::vector<Process>& processes);
```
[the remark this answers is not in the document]
K20shores: done

include/micm/solver/backward_euler.hpp
K20shores: also comment have been updated

include/micm/solver/backward_euler.inl
K20shores: I applied that and everything seems to work

include/micm/solver/backward_euler.hpp
boulderdaze: Ah that makes sense. Thank you for the explanation!

include/micm/solver/backward_euler.inl
```diff
@@ -0,0 +1,195 @@
...
+        // is less than a tolerance times the absolute value of the solution
+        do {
+          // changes that are much smaller than the tolerance are negligible and we assume can be accepted
+          converged = (std::abs(*forcing_iter) <= small) || (std::abs(*forcing_iter) <= tol * std::abs(*yn1_iter));
+          ++forcing_iter, ++yn1_iter;
+        }
+        while (converged && forcing_iter != forcing.end());
+
+        if (!converged) {
+          std::cout << "failed to converge within the newton iteration\n";
```
mwaxmonsky, line 157: Is this `cout` redundant with the `throw:std::system_error` below?

If this is slightly different from the error below, it might be worth setting a different flag and throwing an error based off that instead?

include/micm/solver/backward_euler.inl
```diff
@@ -0,0 +1,195 @@
...
+        }
+      } while(!converged && iterations < max_iter);
+
+      if (!converged) {
+        std::cout << "failed to converge\n";
```
mwaxmonsky, line 162: If throwing an error, is this needed?

mwaxmonsky (approved):
Looks great @K20shores!  Just had a question or two but nothing worth holding up the PR.

include/micm/solver/backward_euler.inl
K20shores: Failing to converge within the loop once doesn't mean that we will fail to converge in subsequent iterations. The error messages below are when an integration does fail, but to match camchem we have to accept some level of integration and continue on.

include/micm/solver/backward_euler.inl
K20shores: Yes, we shouldn't throw an error; I placed that there to make sure we do whatever camchem does. A future PR will remove the error throwing and then at that time logging something would be appropriate, but maybe to `cerr`

PR #493, opened 2024-05-06
Fix the profiler driver 

### LINKED ISSUE #490: Fix the profiler driver
A bug is found: 
```
micm/examples/profile_example.cpp:55:3: error: unknown type name 'ConfigParseStatus'
  ConfigParseStatus status = solverConfig.ReadAndParse(config_path);
```

boulderdaze:
Closes #490 

Removed the outdated `ConfigParseStatus` 
Added it to the git actions

### DIFF
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -22,7 +22,7 @@ jobs:
       - uses: actions/checkout@v3
 
       - name: Run Cmake 
-        run: cmake -S . -B build -D CMAKE_BUILD_TYPE=${{ matrix.build_type }}
+        run: cmake -S . -B build -D CMAKE_BUILD_TYPE=${{ matrix.build_type }} -D MICM_ENABLE_PROFILE=ON
 
       - name: Build
         run: cmake --build build --parallel 10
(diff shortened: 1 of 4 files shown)


K20shores (approved):
Does this get run in any of the github actions?

boulderdaze:
> Does this get run in any of the github actions?

Looks like it doesn't. If `MICM_ENABLE_PROFILE` is `ON`, the executable is created but the default is `OFF` unlike `MICM_ENABLE_EXAMPLES`, which makes sense. I'd like to have it added to the github actions but am not sure which is the best way to do. Creating a docker for this could be easier but am not sure that's what we want to do. What do you think?

K20shores:
@boulderdaze I think it makes sense to add this option to the mac, ubuntu, and windows actions

boulderdaze:
> @boulderdaze I think it makes sense to add this option to the mac, ubuntu, and windows actions

boulderdaze:
> @boulderdaze I think it makes sense to add this option to the mac, ubuntu, and windows actions

Yes, I updated it. Could you review them one more?

K20shores (approved):

PR #495, opened 2024-05-07
Update the assign function inside both CPU and GPU class

### LINKED ISSUE #471: Rewrite the assign function in the Rosenbrock solve function for CPU/GPU
According to the discussion with @mattldawson , the `assign` method (e.g., https://github.com/NCAR/micm/blob/main/include/micm/solver/rosenbrock.inl#L257 and https://github.com/NCAR/micm/blob/main/include/micm/solver/rosenbrock.inl#L363) in the `Solve` function of Rosenbrock class (https://github.com/NCAR/micm/blob/main/include/micm/solver/rosenbrock.inl#L257) probably does not work on the GPU.

Potential solutions:
- Overload the `assign` method for GPU
- Implement a new `operator=` in the matrix constructor to achieve the same functionality of `assign` (maybe we have one already?)

Potential questions:
- Shall we do host->host, device->device, or all together for the assign operation?

Final decision:
- Implement a function called `Copy` in both CPU and GPU dense matrix class, where in the CPU class, this function just calls `assign` method of std::vector class; in the GPU class, this function just does `CudaMemcpy` from device to device
- Add an assert check to make sure that the size of both matrices must match 
- In the existing Copy assignment operator of GPU matrix class, deallocate the GPU memory space if the `d_data_` pointer is not `nullptr`.

sjsprecious:
This PR introduces a new `Copy` function in both the CPU and GPU matrix class.

In the CPU matrix class, it simply calls the `assign` function.

In the GPU matrix class, it only copies the device data.

fix #471

### DIFF
--- a/test/unit/util/test_cuda_dense_matrix.cpp
+++ b/test/unit/util/test_cuda_dense_matrix.cpp
@@ -535,3 +535,39 @@ TEST(CudaDenseMatrix, Axpy)
   double sum = std::accumulate(gpu_y.AsVector().begin(), gpu_y.AsVector().end(), 0);
   EXPECT_EQ(sum, (10 * 2.0 + 20.0) * 198 + 20.0 * 2.0 + 20.0 + 30.0 * 2.0 + 20.0);
 }
+
+TEST(CudaDenseMatrix, CopyFunction)
+{
+  std::vector<std::vector<double>> h_vector{ { 0.1, 2.9 }, { 7.3, 4.5 } };
+  auto matrix = micm::CudaDenseMatrix<double, 2>(h_vector);
+
+  std::vector<std::vector<double>> h_vector2{ { 2.1, 5.9 }, { 6.3, 4.8 } };
+  auto matrix2 = micm::CudaDenseMatrix<double, 2>(h_vector2);
+
+  matrix[0][0] = 8.5;
+
+  EXPECT_EQ(matrix[0][0], 8.5);
+  EXPECT_EQ(matrix[0][1], 2.9);
+  EXPECT_EQ(matrix[1][0], 7.3);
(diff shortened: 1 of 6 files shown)


test/unit/util/test_cuda_dense_matrix.cpp
```diff
@@ -535,3 +535,39 @@ TEST(CudaDenseMatrix, Axpy)
...
+  EXPECT_EQ(matrix[1][0], 53.29);
+  EXPECT_EQ(matrix[1][1], 20.25);
+  for(int i = 0; i < 2; i++)
+  {
+    for(int j = 0; j < 2; j++)
+    {
+      EXPECT_EQ(matrix2[i][j], matrix[i][j]);
+    }
+  }
+}
```
mwaxmonsky, line 573: Missing newline at end of file.

mwaxmonsky (approved):
Looks great @sjsprecious!

test/unit/util/test_cuda_dense_matrix.cpp
mwaxmonsky: I don't believe it is currently. I just checked the past few auto-format commits and found a few (https://github.com/NCAR/micm/commit/d73f05233a447587768c6918101d32c274d88a00#diff-4872e82f861b90b032479d762c8b25d8d0f60e2aff9c415531a943d6077150f7, https://github.com/NCAR/micm/commit/990572ddb343ddf316fd742f26a9505f173eeb70#diff-2ffe8c5559a315998835495219bd20a6366c5d511ea6d3746b0184e16baf3c0b) that left the missing newline.  Might be an issue with the formatter.

mattldawson (approved):
looks great!

PR #511, opened 2024-05-09
Configure clang-tidy to uniformise the style in naming classes, class data members and variables

### LINKED ISSUE #494: Configure clang-tidy to uniformise the style in naming classes, class data members and variables
- [x] Use `CamelCase` for class name
- [x] Use `lower_case` for data members with a trailing underscore

boulderdaze:
Closes #494 
 - Use `CamelCase` for class name
 - Use `lower_case` for data members with a trailing underscore
 - Use `UPPER_CASE` for constants


.clang-tidy
```diff
@@ -54,6 +54,20 @@ CheckOptions:
...
+  - key:             readability-identifier-naming.ClassCase
+    value:           CamelCase
+  - key:             readability-identifier-naming.ClassIgnoredRegexp
+    value:           'is_error_condition_enum'
+  - key:             readability-identifier-naming.MemberCase
+    value:           lower_case
+  - key:             readability-identifier-naming.MemberSuffix
+    value:           _
+  - key:             readability-identifier-naming.MemberIgnoredRegexp
+    value:           '^._$|^k0_._$|^kinf_._$|^Fc_$|^niLU_.*$|[LU]ij_.*$'
```
boulderdaze, line 66: Rethinking this. Maybe we shouldn't require certain case, so we don't need to list exception cases for math parameters. @mattldawson, @K20shores, @mwaxmonsky, what do you think? 
```suggestion
  - key:             readability-identifier-naming.MemberCase
    value:           aNy_CasE
  - key:             readability-identifier-naming.MemberSuffix
    value:           _
```

K20shores: That seems reasonable; we do work in a math field after all and not every variable can easily be named with camel case

K20shores (approved):

.clang-tidy
boulderdaze: I agree with you. I updated it to allow any case for variable names

mattldawson (approved):

mwaxmonsky (approved):
Looks great @boulderdaze!

PR #527, opened 2024-05-17
514 create a builder class that can make a backward euler solver

### LINKED ISSUE #514: Create a builder class that can make a Backward euler solver

K20shores:
Closes #514 

- Temporarily removes backward euler from running in the analytical policy test until we finish implementing the builder such that any solver could be run with those tests
- Adds a solver builder, which has options to build different solvers, right now only BackwardEuler is supported
- Adds a solver class which wraps any solver (backward euler, rosenbrock). This may be a useless class
- Removes template template parameters from many different places

### DIFF
--- a/include/micm/solver/backward_euler.hpp
+++ b/include/micm/solver/backward_euler.hpp
@@ -9,6 +9,7 @@
 #include <micm/profiler/instrumentation.hpp>
 #include <micm/solver/linear_solver.hpp>
 #include <micm/solver/state.hpp>
+#include <micm/solver/backward_euler_solver_parameters.hpp>
 #include <micm/system/system.hpp>
 #include <micm/util/jacobian.hpp>
 
@@ -52,6 +53,7 @@ namespace micm
     void Solve(
         double time_step,
         auto& state,
+        BackwardEulerSolverParameters parameters,
         auto linear_solver,
         auto process_set,
         const std::vector<micm::Process>& processes,
--- a/include/micm/solver/backward_euler.hpp
+++ b/include/micm/solver/backward_euler.hpp
@@ -35,11 +35,6 @@ namespace micm
     /// @brief Default constructor
     BackwardEuler();
 
-    /// @brief Builds a backward eulder solver for the given system and processes
-    /// @param system The chemical system to create the solver for
-    /// @param processes The collection of chemical processes that will be applied during solving
-    BackwardEuler(const System& system, const std::vector<Process>& processes);
-
     virtual ~BackwardEuler() = default;
 
     /// @brief Advances the given step over the specified time step
--- a/include/micm/solver/backward_euler.hpp
+++ b/include/micm/solver/backward_euler.hpp
@@ -7,9 +7,9 @@
 #include <micm/process/process.hpp>
 #include <micm/process/process_set.hpp>
 #include <micm/profiler/instrumentation.hpp>
+#include <micm/solver/backward_euler_solver_parameters.hpp>
 #include <micm/solver/linear_solver.hpp>
 #include <micm/solver/state.hpp>
-#include <micm/solver/backward_euler_solver_parameters.hpp>
 #include <micm/system/system.hpp>
 #include <micm/util/jacobian.hpp>
 
@@ -29,30 +29,40 @@ namespace micm
 {
 
   /// @brief An implementation of the fully implicit backward euler method
+  template <class LinearSolverPolicy, class ProcessSetPolicy> 
   class BackwardEuler
   {
+    BackwardEulerSolverParameters parameters_;
+    LinearSolverPolicy linear_solver_;
+    ProcessSetPolicy process_set_;
+    std::vector<std::size_t> jacobian_diagonal_elements_;
+    std::vector<micm::Process> processes_;
+
    public:
     /// @brief Default constructor
-    BackwardEuler();
+    BackwardEuler(
+      BackwardEulerSolverParameters parameters, 
+      LinearSolverPolicy linear_solver, 
+      ProcessSetPolicy process_set,
+      std::vector<std::size_t> jacobian_diagonal_elements,
+      std::vector<micm::Process>& processes
+      ) : parameters_(parameters),
+          linear_solver_(linear_solver),
+          process_set_(process_set),
+          jacobian_diagonal_elements_(jacobian_diagonal_elements),
+          processes_(processes)
+    {
+    }
 
     virtual ~BackwardEuler() = default;
 
     /// @brief Advances the given step over the specified time step
     /// @param time_step Time [s] to advance the state by
     /// @param state The state to advance
-    /// @param linear_solver The linear solver to use
-    /// @param process_set The process set to use
-    /// @param processes The collection of chemical processes that will be applied during solving
-    /// @param jacobian_diagonal_elements The diagonal elements of the jacobian matrix
     /// @return A struct containing results and a status code
     void Solve(
         double time_step,
-        auto& state,
-        BackwardEulerSolverParameters parameters,
-        auto linear_solver,
-        auto process_set,
-        const std::vector<micm::Process>& processes,
-        auto jacobian_diagonal_elements);
+        auto& state);
   };
 
 }  // namespace micm
--- a/include/micm/solver/backward_euler_solver_parameters.hpp
+++ b/include/micm/solver/backward_euler_solver_parameters.hpp
@@ -1,0 +1,24 @@
+/* Copyright (C) 2023-2024 National Center for Atmospheric Research
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#pragma once
+
+#include <cstddef>
+#include <iostream>
+#include <vector>
+#include <array>
+
+namespace micm
+{
+
+  /// @brief BackwardEuler solver parameters
+  struct BackwardEulerSolverParameters
+  {
+    std::vector<double> absolute_tolerance_;
+    double small { 1.0e-40 };
+    size_t max_number_of_steps_{ 11 };
+    const std::array<double, 5> time_step_reductions{ 0.5, 0.5, 0.5, 0.5, 0.1 };
+  };
+
+}  // namespace micm
--- a/include/micm/solver/solver_builder.inl
+++ b/include/micm/solver/solver_builder.inl
@@ -47,25 +47,25 @@ inline std::error_code make_error_code(MicmSolverBuilderErrc e)
 
 namespace micm
 {
-  SolverBuilder& SolverBuilder::SetSystem(micm::System system)
+  inline SolverBuilder& SolverBuilder::SetSystem(micm::System system)
   {
     system_ = system;
     return *this;
   }
 
-  SolverBuilder& SolverBuilder::SetReactions(std::vector<micm::Process> reactions)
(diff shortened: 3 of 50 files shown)


test/unit/solver/test_cuda_linear_solver.cpp
```diff
@@ -103,22 +90,12 @@ void verify_gpu_against_cpu()
   std::vector<double> cpu_x = linearSolverGenerator<
       Group10000VectorMatrix,
       Group10000SparseVectorMatrix,
-      micm::LinearSolver<double, Group10000SparseVectorMatrix>>(
-      [](const Group10000SparseVectorMatrix<double>& matrix,
-         double initial_value) -> micm::LinearSolver<double, Group10000SparseVectorMatrix> {
-        return micm::LinearSolver<double, Group10000SparseVectorMatrix>{ matrix, initial_value };
-      },
-      10000);
+      micm::LinearSolver<Group10000SparseVectorMatrix>>(10000);
```
sjsprecious, line 93: why don't we need the lambda function here any more?

test/unit/solver/test_cuda_lu_decomposition.cpp
```diff
@@ -1,4 +1,4 @@
-#pragma once
+#include "test_lu_decomposition_policy.hpp"
```
sjsprecious, line 1: Why do we need to remove `pragma once` here and include a new header file?

test/unit/solver/test_cuda_lu_decomposition.cpp
```diff
@@ -81,15 +42,15 @@ void testRandomMatrix(size_t n_grids)
   gpu_A.CopyToDevice();
   gpu_LU.first.CopyToDevice();
   gpu_LU.second.CopyToDevice();
-  gpu_lud.Decompose<double, GPUSparseMatrixPolicy>(gpu_A, gpu_LU.first, gpu_LU.second);
+  gpu_lud.Decompose<GPUSparseMatrixPolicy>(gpu_A, gpu_LU.first, gpu_LU.second);
   gpu_LU.first.CopyToHost();
   gpu_LU.second.CopyToHost();
-  check_results<double, GPUSparseMatrixPolicy>(
+  check_results<typename GPUSparseMatrixPolicy::value_type, GPUSparseMatrixPolicy>(
```
sjsprecious, line 48: It looks like the definition of `check_results` function has been deleted so I wonder how this execution here could work?

sjsprecious (commented):
Thanks @K20shores for working on this giant PR. Really nice work and I just have a few clarification questions besides the three failing tests.

test/unit/solver/test_cuda_linear_solver.cpp
K20shores: It's not clear to me that we ever needed it here. we can call the construction directly in the linear solver function. I can put these back if we like having the lamdas with explicit creation

test/unit/solver/test_cuda_lu_decomposition.cpp
K20shores: `pragma once` is only meant to be in header files. We are now sharing more code with the CPU side of things by including this header file, I think specifically something that checks results

test/unit/solver/test_cuda_lu_decomposition.cpp
K20shores: that's now coming from the new header file

test/unit/solver/test_cuda_linear_solver.cpp
sjsprecious: Thanks. @mattldawson do you recall why we use the lambda function argument before?

test/unit/solver/test_cuda_lu_decomposition.cpp
sjsprecious: Ah, I see. Thanks for the clarification and it makes sense.

include/micm/solver/backward_euler.hpp
```diff
@@ -28,34 +29,40 @@ namespace micm
...
-    /// @param jacobian_diagonal_elements The diagonal elements of the jacobian matrix
     /// @return A struct containing results and a status code
     void Solve(
         double time_step,
-        auto& state,
-        auto linear_solver,
-        auto process_set,
-        const std::vector<micm::Process>& processes,
-        auto jacobian_diagonal_elements);
+        auto& state);
```
boulderdaze, line 65: Could you remove the comment since the return type is void?

include/micm/util/matrix.hpp
```diff
@@ -24,10 +24,21 @@ namespace micm
...
+  using StandardDenseMatrix = Matrix<double>;
+
   /// @brief A 2D array class with contiguous memory
-  template<class T>
+  template<class T = double>
   class Matrix
   {
+    public:
+    // Diagonal markowitz reordering requires an int argument, make sure one is always accessible
+    using IntMatrix = Matrix<int>;
```
boulderdaze, line 38: What would be the use of `Matrix<int>`? The comment suggests it is because of `Diagonal markowitz reordering`. Would it be specific for backward euler? I am curious.

boulderdaze (changes requested):
Looks good! Just have minor change requests. I like the removal of template template parameter.

include/micm/util/matrix.hpp
K20shores: When the solvers are made, we can choose to reorder the species matrix or not. I don't know the details but it apparently minimizes the number of zeros or something, so we should have better cache coherency or something. Any solver can use this, but the function needs a Matrix<int> to do things, and mostly we make a Matrix<double>. Since we aren't passing `template<template<class> Matrix` around anymore, I needed to do something like this so that we always have the int version available.

test/unit/solver/test_cuda_linear_solver.cpp
K20shores: It was for the JITed version of things. We needed to pass the jit compiler around, but, I made the jit compiler into a singleton and now the interface to all solvers and their components are the same

boulderdaze (approved):
That makes sense that the Jit class is singleton. Looks cleaner!

test/tutorial/test_README_example.cpp
```diff
@@ -30,21 +30,52 @@ int main(const int argc, const char *argv[])
 
   std::vector<Process> reactions{ r1, r2 };
 
-  RosenbrockSolver<> solver{ chemical_system, reactions, RosenbrockSolverParameters::ThreeStageRosenbrockParameters() };
+#if 0
+  auto rosenbrock_solver = Solver::Create()
+                    .SetSystem(chemical_system)
+                    .SetReactions(reactions)
+                    .SetNumberOfGridCells(19)
+                    .Rosenbrock<MatrixPolicy, SparseMatrixPolicy>(RosenbrockSolverParameters::ThreeStageRosenbrockParameters());
```
mattldawson, line 38: I think we were figuring out how the `SolverBuilder` would work in this test, but I wonder if we should put this somewhere else and keep this test matching what's in the README, so it keeps getting tested?

mattldawson (approved):
looks great! this was a lot of work! Just a couple minor questions, but nothing important

K20shores:
Intel failure seems to be in [other repositories](https://github.com/ggerganov/llama.cpp/issues/7507) as well

sjsprecious (approved):
Thanks @K20shores for working on this giant PR.

All the 44 tests pass on Derecho's A100 GPU, using `nvhpc/23.7`.

mofosyne:
If you rebase on the current master it should include the temporary intel workflow disable and allow all the other test to run through. (Might not be too important as you can always merge in anyway, but good for OCD)

include/micm/solver/solver_builder.inl
```diff
@@ -0,0 +1,277 @@
...
+    {
+      throw std::runtime_error("Not implemented yet");
+    }
+    if (std::holds_alternative<BackwardEulerSolverParameters>(options_))
+    {
+      return BuildBackwardEulerSolver();
+    }
+
+    throw std::runtime_error("No solver type set");
+  }
```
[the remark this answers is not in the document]
K20shores: I don't have a preference on if we do that or not. Maybe you could do that in the next PR? This one already has so much

include/micm/solver/backward_euler_solver_parameters.hpp
```diff
@@ -0,0 +1,24 @@
...
+
+namespace micm
+{
+
+  /// @brief BackwardEuler solver parameters
+  struct BackwardEulerSolverParameters
+  {
+    std::vector<double> absolute_tolerance_;
+    double small { 1.0e-40 };
+    size_t max_number_of_steps_{ 11 };
```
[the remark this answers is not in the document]
K20shores: not if we want to allow them to be modified. And with that statement I should probably remove const on the time step reductions

test/integration/analytical_policy.hpp
```diff
@@ -162,18 +160,18 @@ void test_analytical_troe(
...
-      be.Solve(
-          time_step, be_state, linear_solver, process_set, processes, solver.state_parameters_.jacobian_diagonal_elements_);
-    }
+    // if constexpr (std::is_same_v<decltype(solver.process_set_), micm::ProcessSet>)
+    // {
+    //   auto linear_solver = solver.linear_solver_;
+    //   auto process_set = solver.process_set_;
+    //   be.Solve(
+    //     time_step, be_state, micm::BackwardEulerSolverParameters(), linear_solver, process_set, processes, solver.state_parameters_.jacobian_diagonal_elements_);
+    // }
```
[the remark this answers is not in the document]
K20shores: Oh, I do think we can delete this. I'm hoping that all of these tests can be passed any type of solver and check the results. I can dream and hope to make it come true by deleting those

test/integration/analytical_policy.hpp
```diff
@@ -162,18 +160,18 @@ void test_analytical_troe(
...
     EXPECT_EQ(result.state_, (micm::SolverState::Converged));
     EXPECT_NEAR(k1, state.rate_constants_.AsVector()[0], 1e-8);
     EXPECT_NEAR(k2, state.rate_constants_.AsVector()[1], 1e-8);
     model_concentrations[i_time] = result.result_.AsVector();
-    be_model_concentrations[i_time] = be_state.variables_[0];
+    // be_model_concentrations[i_time] = be_state.variables_[0];
```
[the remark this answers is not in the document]
K20shores: yes

test/unit/solver/test_solver_builder.cpp
```diff
@@ -0,0 +1,96 @@
...
+  //                           .SolverParameters(micm::ThreeStageRosenbockSolverParameters{})
+  //                           .Build();
+
+  // auto rosenbrock_vector = micm::CpuSolverBuilder<VectorMatrix<double, L>, SparseVectorMatrix<double, L>>()
+  //                           .SetSystem(the_system)
+  //                           .SetReactions(reactions)
+  //                           .SetNumberOfGridCells(1)
+  //                           .SolverParameters(micm::ThreeStageRosenbockSolverParameters{})
+  //                           .Build();
+}
```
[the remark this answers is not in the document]
K20shores: this is meant to be here. It will be uncommented when this gets working in upcoming PRs

PR #543, opened 2024-05-31
Share CUBLAS handle between different CUBLAS matrices

### LINKED ISSUE #428: Ownership of a cublas handle
Thanks to @mwaxmonsky , it turns out that the cublas handle should be associated with a pair of host thread/rank and device (see the Section `2.1.2. cuBLAS Context` at https://docs.nvidia.com/cuda/cublas/#example-code).

Currently a cublas handle is generated for each object of class `CudaVectorMatrix`. This is fine but just redundant. In addition, creating/destroying a cublas handle is way much expensive compared to calling a cublas API. Thus even if they are done in the constructor/destructor, we should still try to reduce its frequency as much as possible for the efficiency purpose.

Proposed solution (not sure if it is doable):
- Each MPI rank / OMP thread will generate a cublas handle at the beginning (even before the solver object is created), then it passes the handle to whatever function that needs it.

This issue should be treated as an optimization step and we won't worry about it until we have a multi-GPU test.

Ideas:
- Generate a singleton like the `GetInstance()` function for Jitcompiler https://github.com/NCAR/micm/blob/555be4edd17220c923e9e4ae298662c690939d81/include/micm/jit/jit_compiler.hpp#L108-L121
- In the Cuda matrix constructor, call the `GetInstance()` function instead of creating the cublas handle directly.
- Whichever the first Cuda matrix will generate the cublas handle; the rest Cuda matrices will just share the same pointer later.

sjsprecious:
This PR implements the generation of CUBLAS handle as singleton and shares between CUDA matrices. This will avoid the unnecessary creation of CUBLAS handle for each CUDA matrix and save some computational costs.

All the 44 tests pass on Derecho's A100 GPU using `nvhpc/23.7`.

fix #428


include/micm/util/cuda_dense_matrix.hpp
```diff
@@ -240,5 +236,18 @@ namespace micm
...
+    {
+      static cublasHandle_t cublas_handle = NULL;
+      if (!cublas_handle && !is_destructor) CHECK_CUBLAS_ERROR(cublasCreate(&cublas_handle), "CUBLAS initialization failed...");
+      if (!cublas_handle && is_destructor)
+      {
+        CHECK_CUBLAS_ERROR(cublasDestroy(cublas_handle), "CUBLAS finalization failed...");
+        cublas_handle = NULL;
+      }
+      return cublas_handle;
+    }
```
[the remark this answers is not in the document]
K20shores: Would something like this work? It adds a custom deleter to the shared pointer so that we don't have to check the count of the shared pointer and we no longer need the boolean to decide when to destroy. Yes, I helped generate this with chatgpt


```
#include <memory>
#include <stdexcept>
#include <iostream>
#include <mutex>
#include <cublas_v2.h>

class CublasHandleSingleton {
public:
    // Get the cublas handle as a shared pointer with custom deleter
    static std::shared_ptr<cublasHandle_t> getHandle() {
        std::lock_guard<std::mutex> lock(getMutex()); // Lock the mutex
        static CublasHandleSingleton instance; // Ensures only one instance
        return instance.handle;
    }

private:
    std::shared_ptr<cublasHandle_t> handle;
    static std::mutex& getMutex() {
        static std::mutex mutex;
        return mutex;
    }

    // Private constructor to prevent instantiation
    CublasHandleSingleton() : handle(new cublasHandle_t, [](cublasHandle_t* handle) {
        if (*handle != nullptr) {
            cublasStatus_t status = cublasDestroy(*handle);
            if (status != CUBLAS_STATUS_SUCCESS) {
                std::cerr << "CUBLAS finalization failed: " << status << std::endl;
            }
            delete handle;
        }
    }) {
        cublasStatus_t status = cublasCreate(handle.get());
        if (status != CUBLAS_STATUS_SUCCESS) {
            throw std::runtime_error("CUBLAS initialization failed.");
        }
    }

    // Delete copy constructor and assignment operator
    CublasHandleSingleton(const CublasHandleSingleton&) = delete;
    CublasHandleSingleton& operator=(const CublasHandleSingleton&) = delete;

    // Destructor
    ~CublasHandleSingleton() = default; // Custom deleter in shared_ptr handles destruction
};

// Usage example
int main() {
    try {
        std::shared_ptr<cublasHandle_t> handle = CublasHandleSingleton::getHandle();
        // Use the handle for cuBLAS operations

        // When the shared_ptr goes out of scope, the custom deleter will be called
    } catch (const std::exception& e) {
        std::cerr << e.what() << std::endl;
        return EXIT_FAILURE;
    }
    return EXIT_SUCCESS;
}

```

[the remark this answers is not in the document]
sjsprecious: @K20shores @mwaxmonsky thank you both for suggesting the implementation of singleton class for the cublas handle. After merging @mattldawson 's latest solver builder changes that address the GPU memory leak in the cuda rosenbrock test, the singleton class now works for the CUBLAS code and this PR is ready for your review again.

include/micm/util/cublas_handle_singleton.hpp
```diff
@@ -0,0 +1,74 @@
...
+            std::cout << "Mapping multiple OpenMP threads to the same GPU...\n";
+            std::lock_guard<std::mutex> lock(GetMutex()); // Lock the mutex so that only one thread will execute the following code at a time;
+                                                            // Unlock the mutex when the thread goes out of scope;
+                                                            // I think this is needed if multiple threads are mapped to the same GPU.
+            }
+        }
+        else
+        {
+            throw std::runtime_error("Please set the environment variables OMP_NUM_THREADS and CUDA_VISIBLE_DEVICES.\n");
+        }
```
K20shores, line 34:
```suggestion
        auto lock = GetMutex();
```
I see no problem with getting the lock each time, even if we are in a single threaded context. Do we need all of this logic?

K20shores (approved):

sjsprecious:
When using the `nvhpc/23.7` compiler on Derecho with `gcc-toolchain/12`, I got the the following warning messages:
```
nvcc warning : The -std=c++20 flag is not supported with the configured host compiler. Flag will be ignored.
/glade/derecho/scratch/sunjian/micm/include/micm/util/cublas_handle_singleton.hpp(40): warning #3033-D: inline variables are a C++17 feature
        inline static cublasHandle_t handle_;
        ^

Remark: The warnings can be suppressed with "-diag-suppress <warning-number>"

/glade/derecho/scratch/sunjian/micm/include/micm/util/cublas_handle_singleton.hpp(21): warning #2919-D: C++17-style initializer is nonstandard in this mode
          if (auto search = cublas_handle_map.find(device_id); search == cublas_handle_map.end())
                                                             ^

/glade/derecho/scratch/sunjian/micm/include/micm/util/cublas_handle_singleton.hpp(40): warning #3033-D: inline variables are a C++17 feature
        inline static cublasHandle_t handle_;
        ^

Remark: The warnings can be suppressed with "-diag-suppress <warning-number>"

/glade/derecho/scratch/sunjian/micm/include/micm/util/cublas_handle_singleton.hpp(21): warning #2919-D: C++17-style initializer is nonstandard in this mode
          if (auto search = cublas_handle_map.find(device_id); search == cublas_handle_map.end())
                                                             ^

/glade/derecho/scratch/sunjian/micm/include/micm/util/cublas_handle_singleton.hpp:40:39: warning: inline variables are only available with -std=c++1z or -std=gnu++1z
       inline static cublasHandle_t handle_;
                                       ^~~~~  
/glade/derecho/scratch/sunjian/micm/include/micm/util/cublas_handle_singleton.hpp:41:48: warning: inline variables are only available with -std=c++1z or -std=gnu++1z
       inline static std::map<int, cublasHandle_t> cublas_handle_map;
                                                ^~~~~~~~~~~~~~~~~
/glade/derecho/scratch/sunjian/micm/include/micm/util/cublas_handle_singleton.hpp: In static member function ‘static micm::CublasHandleSingleton& micm::CublasHandleSingleton::GetInstance()’:
/glade/derecho/scratch/sunjian/micm/include/micm/util/cublas_handle_singleton.hpp:21:5: warning: init-statement in selection statements only available with -std=c++1z or -std=gnu++1z
         if (auto search = cublas_handle_map.find(device_id); search == cublas_handle_map.end())
     ^   
/glade/derecho/scratch/sunjian/micm/include/micm/util/cublas_handle_singleton.hpp:26:5: warning: init-statement in selection statements only available with -std=c++1z or -std=gnu++1z
         if (auto search = cublas_handle_map.find(device_id); search == cublas_handle_map.end())
```
@K20shores @mwaxmonsky @mattldawson just wonder whether there is a way to address these warning messages?

include/micm/util/cublas_handle_singleton.hpp
sjsprecious: Thanks @K20shores for your comment. I have added the `std::map` implementation as we discussed.

include/micm/util/cublas_handle_singleton.hpp
```diff
@@ -0,0 +1,72 @@
...
+      CublasHandleSingleton(const CublasHandleSingleton&) = delete;
+      CublasHandleSingleton& operator=(const CublasHandleSingleton&) = delete;
+
+      // Get the static instance of CublasHandleSingleton class
+      static CublasHandleSingleton& GetInstance(int device_id)
+      {
+        if (auto search = cublas_handle_map_.find(device_id); search == cublas_handle_map_.end())
+        {
+          std::lock_guard<std::mutex> lock(GetMutex()); // no cublas handle if found; lock the mutex and generate a new cublas handle below
+        }
```
K20shores, line 22: Placing the lock inside of the if will immediately grab the lock but then let it go once the if block ends because the lock will go out of scope. So we should just grab the lock every time 

```suggestion  
        std::lock_guard<std::mutex> lock(GetMutex()); // no cublas handle if found; lock the mutex and generate a new cublas handle below
```

K20shores (changes requested):

include/micm/util/cublas_handle_singleton.hpp
sjsprecious: Thanks for catching it. Fixed!

sjsprecious:
@K20shores @mwaxmonsky After discussing with @mattldawson , it seems that we do not need the singleton class at all, but just a function to access the static map variable. Thus I further simplify the implementation in the latest commit.

src/util/cuda_util.cu
```diff
@@ -26,5 +30,21 @@ namespace micm
...
+      CHECK_CUDA_ERROR(cudaGetDevice(&device_id), "Failed to get device ID...");
+      std::lock_guard<std::mutex> lock(mutex); // lock the mutex and generate a new cublas handle below
+      if (auto search = cublas_handles_map.find(device_id); search == cublas_handles_map.end())
+      {
+        cublasHandle_t handle;
+        CHECK_CUBLAS_ERROR(cublasCreate(&handle), "CUBLAS initialization failed..."); // create the cublas handle
+        cublas_handles_map[device_id] = handle;                                       // save the cublas handle
+      }
+      return cublas_handles_map[device_id];
+    }
```
K20shores, line 48: We don't seem to destroy the handle this way

```suggestion
    auto cublas_handle_deleter = [](cublasHandle_t* handle) {
        if (handle != nullptr) {
            CHECK_CUBLAS_ERROR(cublasDestroy(*handle), "CUBLAS finalization failed");
            delete handle;
        }
    };

    // Define the smart pointer type using decltype for the custom deleter
    using CublasHandlePtr = std::unique_ptr<cublasHandle_t, decltype(cublas_handle_deleter)>;
    
    CublasHandlePtr CreateCublasHandle() {
      cublasHandle_t handle;
      CHECK_CUBLAS_ERROR(cublasCreate(&handle), "CUBLAS initialization failed...");
      return CublasHandlePtr(new cublasHandle_t(handle), &cublasDestroy);
    }
    
    cublasHandle_t& GetCublasHandle()
    {
      static std::map<int, CublasHandlePtr> cublas_handles_map;
      static std::mutex mutex;
      int device_id;
      CHECK_CUDA_ERROR(cudaGetDevice(&device_id), "Failed to get device ID...");
      std::lock_guard<std::mutex> lock(mutex); // lock the mutex and generate a new cublas handle below
      if (auto search = cublas_handles_map.find(device_id); search == cublas_handles_map.end())
      {
        cublas_handles_map[device_id] = CreateCublasHandle();
      }
      return *cublas_handles_map[device_id];
    }
```

This uses a smart pointer with a custom deleter so that we still release the cuda memory

K20shores (changes requested):

src/util/cuda_util.cu
sjsprecious: Thanks @K20shores for your suggestions. I have pushed the suggested changes with some issues fixed by Copilot (hopefully they make sense to you).

One thing I want to clarify: the cublas handle itself does not seem to allocate any memory on the device (the GPU memory check passes for my implementation).

K20shores (approved):

src/util/cuda_util.cu
K20shores: The [cuda documentation](https://docs.nvidia.com/cuda/cublas/index.html?highlight=cublasDestroy#cublasdestroy) says that some hardware resources are released. I imagine it doesn't really matter for us since we only require the resource once and not multiple times during a single run. But it seems like it's a good idea to ensure we release them

mattldawson (approved):
looks great!

src/util/cuda_util.cu
sjsprecious: Agreed!

mwaxmonsky (approved):

PR #547, opened 2024-06-04
546 update cmakeliststxt to support latest cuda project type and architectures

### LINKED ISSUE #546: Update CMakeLists.txt to support latest CUDA project type and architectures
**Is your feature request related to a problem? Please describe.**
Current implementation hard codes sm80 so this will not be usable on v100s or h100s.

**Describe the solution you'd like**
A flag that turns on a particular version or set of supported gpu architectures.

**Describe alternatives you've considered**
N\A

**Additional context**
N\A

mwaxmonsky:
Simplifies CMake infrastructure by removing openACC support and modernizes CUDA integration with CMake.

Closes #546


K20shores (commented):
Lots of red. Beautiful! and I like the completeness of the architectures

K20shores (approved):

mwaxmonsky:
> Lots of red. Beautiful! and I like the completeness of the architectures

I wanted more red but CMake isn't quite there for CUDA yet.  :(

I also was thinking would it make sense to add a default architecture if none is provided (ie. if `MICM_GPU_TYPE` is _not_ set but `MICM_ENABLE_CUDA` is, maybe set `MICM_GPU_ARCH=all-major`?) (https://cmake.org/cmake/help/latest/prop_tgt/CUDA_ARCHITECTURES.html)

cmake/dependencies.cmake
```diff
@@ -114,12 +114,33 @@ endif()
...
+  set(cuda_arch_map_v100 70)
+  set(cuda_arch_map_h100 90a)
+  set(cuda_arch_map_h200 90a)
+  set(cuda_arch_map_b100 95)
+  set(cuda_arch_map_b200 95)
+  set(MICM_GPU_ARCH ${cuda_arch_map_${MICM_GPU_TYPE}})
+
+  if(NOT MICM_GPU_ARCH)
+    message(FATAL_ERROR "MICM_GPU_TYPE (${MICM_GPU_TYPE}) is not recognized. Available options are [a100, v100, h1/200].")
+  endif()
```
sjsprecious, line 131: Maybe we can combine these output to line 118 and just ask the users to provide a valid GPU type.

sjsprecious: also is the `MICM_GPU_TYPE` input case sensitive?

src/CMakeLists.txt
```diff
@@ -47,59 +47,18 @@ if(MICM_ENABLE_LLVM)
...
     PRIVATE micm
   )
 
   target_link_libraries(micm_cuda 
-    PUBLIC cudart cublas culibos
+    PUBLIC CUDA::cudart CUDA::cublas
   )
-
-  target_include_directories(micm_cuda PUBLIC /usr/local/cuda/include)
+  set_target_properties(micm_cuda PROPERTIES CUDA_ARCHITECTURES "${MICM_GPU_ARCH}")
```
sjsprecious, line 61: I look at the `CMakeCache.txt` file and the `CMAKE_CUDA_ARCHITECTURES` is not set to what we want here. The way that works in my case is to set the `CUDAARCHS` env variable instead (https://github.com/NCAR/micm/blob/428-singleton-for-cublas-handle/src/CMakeLists.txt#L55-L64).

src/CMakeLists.txt
```diff
@@ -47,59 +47,18 @@ if(MICM_ENABLE_LLVM)
...
   add_library(musica::micm_cuda ALIAS micm_cuda)
-  set(CMAKE_CUDA_ARCHITECTURES 80)
+  enable_language(CUDA)
   target_link_libraries(micm_cuda 
     PRIVATE micm
   )
 
   target_link_libraries(micm_cuda 
-    PUBLIC cudart cublas culibos
+    PUBLIC CUDA::cudart CUDA::cublas
```
sjsprecious, line 59: Do we need `CUDA::` here?

src/CMakeLists.txt
```diff
@@ -47,59 +47,18 @@ if(MICM_ENABLE_LLVM)
...
-endif()
-
 if(MICM_ENABLE_CUDA)
-  message(STATUS "GPU architecture ${MICM_GPU_ARCH}")
-  set(CMAKE_CUDA_ARCHITECTURES ${MICM_GPU_ARCH})
-
   add_library(micm_cuda)
   add_library(musica::micm_cuda ALIAS micm_cuda)
-  set(CMAKE_CUDA_ARCHITECTURES 80)
+  enable_language(CUDA)
```
sjsprecious, line 53: We have enabled CUDA in the `cmake/dependencies.cmake` file. Do we need to re-enable it here?

sjsprecious (changes requested):
Thanks @mwaxmonsky for working on this fix. It looks nice and I just have a few questions.

src/CMakeLists.txt
mwaxmonsky: Good catch!  Removed!

src/CMakeLists.txt
mwaxmonsky: Yup!  The old `find_package(cuda)` is deprecated and requires using the new `enable_language(cuda)` instead but we still need to find the libraries.  CMake brought in a new `find_package(CUDAToolkit)` command which requires `CUDA::*` before each of the found libraries (https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html#cuda-runtime-library).

src/CMakeLists.txt
sjsprecious: According to @mwaxmonsky 's debugging, the `CMakeCache.txt` file does not seem to matter here and the correct CUDA flag is used for a given GPU architecture (see the `build/src/CMakeFiles/micm_cuda.dir/flags.make` file for details).

src/CMakeLists.txt
sjsprecious: Thanks for your clarification. That is good to know.

cmake/dependencies.cmake
mwaxmonsky: Based on conversations in the virtual office, we do want to combine these so that is incoming!

src/CMakeLists.txt
mwaxmonsky: After our discussion and with some additional testing, we are now setting `ENV{CUDAARCHS}` as you are correct and this overrides all of the architecture variables.

This means the build directory will be locked to the architecture passed in on the command line, if provided, but I would not want to trust the build artifacts to update appropriately even if we did allow this.

cmake/dependencies.cmake
mwaxmonsky: Removed!

cmake/dependencies.cmake
```diff
@@ -113,13 +113,30 @@ endif()
...
+  set(cuda_arch_map_a100 80)
+  set(cuda_arch_map_v100 70)
+  set(cuda_arch_map_h100 90a)
+  set(cuda_arch_map_h200 90a)
+  set(cuda_arch_map_b100 95)
+  set(cuda_arch_map_b200 95)
+  set(cuda_arch_map_ all-major)
+  # Setting CUDAARCHS does not override CMAKE_CUDA_ARCHITECTURES or CUDA_ARCHITECTURES
+  # until the current process has returned to the caller site in CMake.
+  set(ENV{CUDAARCHS} ${cuda_arch_map_${MICM_GPU_TYPE}})
```
sjsprecious, line 126: So this logic won't handle the GPU type other than NVIDIA gpus. Would the following one be more appropriate?
```
if(${MICM_GPU_TYPE} STREQUAL "v100")
  set(ENV{CUDAARCHS} 70)
elseif (${MICM_GPU_TYPE} STREQUAL "a100")
  set(ENV{CUDAARCHS} 80)
elseif (${MICM_GPU_TYPE} STREQUAL "h100" OR ${MICM_GPU_TYPE} STREQUAL "h200")
  set(ENV{CUDAARCHS} 90a)
elseif (${MICM_GPU_TYPE} STREQUAL "b100" OR ${MICM_GPU_TYPE} STREQUAL "b200")
  set(ENV{CUDAARCHS} 95)
else()
  message(FATAL_ERROR "Unsupported GPU type: ${MICM_GPU_TYPE}.")
endif()
```

CMakeLists.txt
```diff
@@ -34,22 +34,14 @@ option(MICM_ENABLE_COVERAGE "Enable code coverage output" OFF)
 option(MICM_ENABLE_MEMCHECK "Enable memory checking in tests" OFF)
 option(MICM_ENABLE_JSON "Enable json configuration file reading" ON)
 option(MICM_BUILD_DOCS "Build the documentation" OFF)
-option(MICM_ENABLE_CUDA "Build with Cuda support" OFF)
-option(MICM_ENABLE_OPENACC "Build with OpenACC Support" OFF)
 option(MICM_ENABLE_LLVM "Build with LLVM support for JIT-compiling" OFF)
 option(MICM_ENABLE_TESTS "Build the tests" ON)
 option(MICM_ENABLE_EXAMPLES "Build the examples" ON)
 option(MICM_ENABLE_PROFILE "Profile MICM Solver" OFF)
+set(MICM_GPU_TYPE "None" CACHE STRING "The GPU type being targeted")
```
sjsprecious, line 41: shall we lower case the `MICM_GPU_TYPE` here or somewhere later?

mwaxmonsky: Apologies!  Forgot to add that until just now in my latest commit.  I added it in the dependencies.cmake since that is where we are doing our checking but am happy to move it elsewhere.

cmake/dependencies.cmake
mwaxmonsky: I like that the architecture value is listed once here but I think we want to avoid this for the time being because we are going to have to do this for each new gpu type that we want to support and this might get verbose.

We will also need to refactor the current implementation anyway when we move to supporting Intel and AMD and I suspect this will change significantly then as well with newer CMake support.

CMakeLists.txt
K20shores: Why lowercase this option?

sjsprecious: so we can allow a user to provide either `a100` or `A100`

cmake/dependencies.cmake
sjsprecious: Thanks for the clarification. That is clear to me now.

sjsprecious (approved):
Thanks @mwaxmonsky for working on this fix. It looks great to me.

All the 44 tests pass on Derecho's A100 GPU, using `nvhpc/23.7`.

K20shores (approved):

mattldawson (approved):
nice!

PR #551, opened 2024-06-05
Add option for creating a third-body species in the MICM configuration 

boulderdaze:
Allow users to specify that a species should be treated as a "third-body" species whose concentration is automatically set to the total air density (mol m-3).

Closes musica issue [#115](https://github.com/NCAR/musica/issues/115)

- Updated github action to version 4
- Removed invalid code that sets third body name when it matches `M`. Setting third body type is done by parser.


K20shores (approved):

include/micm/configure/solver_config.hpp
```diff
@@ -366,7 +372,14 @@ namespace micm
         {
           if (value.is_string())
           {
-            species.SetProperty<std::string>(key, value);
+            if (key == TRACER_TYPE && value == THIRD_BODY)
+            {
+              species = Species::ThirdBody();
```
mattldawson, line 377: If the user gives a name other than "M" for a third-body species, I think we should either set the species name they provided or throw an error saying third-body species are always named "M"

mattldawson (changes requested):
Looks great! Just one minor suggestion

include/micm/configure/solver_config.hpp
boulderdaze: My understanding was that this function means to overwrite any names provided by users for third body and always set it as `M`.  Does it mean this function shouldn't set the third body name?
```c++
  inline Species Species::ThirdBody()
  {
    Species third_body{ "M" };
    third_body.parameterize_ = [](const Conditions& c) { return c.air_density_; };
    return third_body;
  }
```
Another question is that the two approaches sound very different, either change the name or throw an error; Can we decide which way makes more sense?

mattldawson: I think changing the name to what the user supplied makes the most sense, instead of throwing an error.

The name set by `Species::ThirdBody()` is just a default, but if it makes sense to pass a name to the function maybe that would be better?

boulderdaze: Yeah, I removed `static` keyword for `ThirdBody()` function so it actually has access to the memeber data including the name. 

Is[ this code](https://github.com/NCAR/micm/blob/main/examples/example.cpp#L63-L64) still valid? Currently, the configuration has to specify `"tracer type" : "THIRD_BODY"` to declare that the species is `third body` type. It has nothing to do with species name because we accept that users can name it whatever they want. 
```c++
    if (species.name_ == "M")
      species.parameterize_ = [](const Conditions& c) { return c.air_density_; }; 
```
If this is not valid anymore and `M` doesn't necessarily mean third body, this code should be removed.

mattldawson: I'm not sure I follow about needing to remove the static keyword. I thought the function should return a new instance of `Species`, something like:
```
  inline Species Species::ThirdBody(const std::string name)
  {
    Species third_body{ name };
    third_body.parameterize_ = [](const Conditions& c) { return c.air_density_; };
    return third_body;
  }
```

And, yes, I agree that the code that depends on the name should be removed. We should probably also update any examples that have an `M` species to make sure that they now include the `"tracer type": "THIRD_BODY"`

boulderdaze: In the [parser](https://github.com/NCAR/micm/blob/third_body/include/micm/configure/solver_config.hpp#L366-L383), `Species` is created before setting their properties. Species object is already created and we want to maintain the name that the parser read, so `ThirdBody` function needs access to its instance name. 
Species that becomes third body is still a non-static member and by declaring `Thirdbody` function static, we get to create species in stack again and has to call copy assignment.  
While I understand that your suggestion also makes sense, when I think about what `third body` does is just to set the `parameterize_`. I am not sure why the function should be static, but I might miss some information that makes sense for third body function to be static?

mattldawson: That makes sense. I'm fine with that approach too.

mattldawson (approved):
looks great!

PR #554, opened 2024-06-05
Add error checks for all the CUDA APIs

### LINKED ISSUE #550: Add error check for all the CUDA APIs
We have implemented the error checking functions for CUDA (https://github.com/NCAR/micm/blob/main/include/micm/util/cuda_matrix.cuh#L53) and CUBLAS (https://github.com/NCAR/micm/blob/main/include/micm/util/cuda_matrix.cuh#L60) APIs.
 
However, there are many CUDA APIs in the MICM CUDA code (https://github.com/NCAR/micm/blob/main/src/solver/rosenbrock.cu#L51), where we do not check the returned values and this could lead to a malfunctioning CUDA kernel later.

**Describe the solution you'd like**
Add the existing error check functions to all the CUDA APIs in MICM.

sjsprecious:
This PR adds the error checks for all the CUDA APIs used in the MICM CUDA code.

With this change, we find a potential GPU memory issue in the `test_cuda_rosenbrock` test. After discussing with @mattldawson , we will update the solver builder in the `test_cuda_rosenbrock` and hopefully the test will pass later.

fix #550

### DIFF
--- a/include/micm/util/cuda_dense_matrix.hpp
+++ b/include/micm/util/cuda_dense_matrix.hpp
@@ -5,6 +5,7 @@
 #pragma once
 
 #include <micm/util/cuda_matrix.cuh>
+#include <micm/util/cuda_util.cuh>
 #include <micm/util/error.hpp>
 #include <micm/util/vector_matrix.hpp>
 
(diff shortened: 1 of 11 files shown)


mattldawson (approved):
Looks good! My only suggestion would be to move the error macros to a `include/micm/util/cuda_error.hpp` or `include/micm/util/cuda_util.hpp` header so that files that need them don't have to import a matrix header, but it's up to you if you want to do this.

mwaxmonsky (approved):

PR #556, opened 2024-06-05
removing irrelevant cmake tests

K20shores:
I feel like these tests don't do anything for us and we should remove them. What do y'all think? We inherently test that we can use fetch content by using it in musica


boulderdaze (approved):

sjsprecious (approved):

mattldawson (approved):
fine by me!

mwaxmonsky (approved):

PR #559, opened 2024-06-07
Modify regex in clang format

boulderdaze:
Modify regex in clang format

- `IndentPPDirectives: AfterHash`: current
```c++
#if FOO
#  if BAR
#    include <foo>
#  endif
#endif
```
- `IndentPPDirectives: BeforeHash`: this PR
```c++
#if FOO
  #if BAR
    #include <foo>
  #endif
#endif
```
The current REGEX failed to list an external header with `-` above system headers. (The example is from tuvx, which I found the failure.)
Current:
```c++: 
#include <tuvx/util/config_yaml.h>

#include <cstring>
#include <fstream>

#include <yaml-cpp/yaml.h>
```
This PR:
```c++: 
#include <tuvx/util/config_yaml.h>

#include <yaml-cpp/yaml.h>

#include <cstring>
#include <fstream>
```


boulderdaze:
> @boulderdaze, could you update the description with a few examples of what will change when the format bot runs after this change?

mwaxmonsky:
> > @boulderdaze, could you update the description with a few examples of what will change when the format bot runs after this change?
> 
> Yes, that's a good idea. I updated the comments with the examples.

Thank you!

mwaxmonsky (approved):
Looks great @boulderdaze!

mattldawson (approved):

K20shores (approved):

Issue #499, opened 2024-05-07
Refactor tests to use both solvers for all analytical tests


K20shores:
Refactor test cases so that either Rosenbrock or Backward Euler can be used.

Blocked by #500

**Acceptance Criteria**
- All analytical ODE solver tests are applied to the Backward Euler solver, and they all pass

**Ideas**
- Create an `analytical_backward_euler.cpp` test following the logic of https://github.com/NCAR/micm/blob/main/test/integration/analytical_rosenbrock.cpp

PR #560, opened 2024-06-07
Apply solver builder to Rosenbrock

### LINKED ISSUE #496: Create solver builder for rosenbrock, backward euler
Create a solver builder class which makes the linear solver, the state, the jacobian information, etc. so that either backward euler or rosenbrock can be created

- [x] #514
- [x] #538
- [x] #529
- [x] #539

**Acceptance criteria**
- The builder wrapper should be only called once during a simulation.
- We add a new builder for the Rosenbrock solver and a test for it (does not break the existing tests).
- We add a new builder for the Backward Euler solver and a test for it (does not break the existing tests).
- We update the existing Rosenbrock tests with the new solver builder and fix the broken tests.

**Ideas**
- In the test driver, we can generate a Rosenbrock solver like:
```
  auto rosenbrock_solver = Solver::Create()
                    .SetSystem(chemical_system)
                    .SetReactions(reactions)
                    .SetNumberOfGridCells(19)
                    .Rosenbrock<MatrixPolicy, SparseMatrixPolicy>(RosenbrockSolverParameters::ThreeStageRosenbrockParameters());
```
and a backward Euler solver like:
```
  auto backward_euler_solver = Solver::Create()
                    .SetSystem(chemical_system)
                    .SetReactions(reactions)
                    .SetNumberOfGridCells(19)
                    .BackwardEuler(BackwardEulerParameters::CamChemParameters());
```
and a CUDA rosenbrock solver like:
```
  auto cuda_solver = Solver::Create()
                    .SetSystem(chemical_system)
                    .SetReactions(reactions)
                    .SetNumberOfGridCells(19)
                    .Rosenbrock<CudaDenseMatrixPolicy, CudaSparseMatrixPolicy>(RosenbrockSolverParameters::ThreeStageRosenbrockParameter());
```
- then we can generate a state variable for different solvers like:
```
  auto rosenbrock_state = rosenbrock_solver.GetState();
  auto backward_euler_state = backward_euler_solver.GetState();
  auto cuda_state = cuda_solver.GetState();
``` 
- then we can initialize the conditions like:
```
  rosenbrock_state.conditions_[0].temperature_ = 287.45;  // K
  rosenbrock_state.conditions_[0].pressure_ = 101319.9;   // Pa
  rosenbrock_state.SetConcentration(foo, 20.0);           // mol m-3

  backward_euler_state.conditions_[0].temperature_ = 287.45;  // K
  backward_euler_state.conditions_[0].pressure_ = 101319.9;   // Pa
  backward_euler_state.SetConcentration(foo, 20.0);           // mol m-3

  cuda_state.conditions_[0].temperature_ = 287.45;  // K
  cuda_state.conditions_[0].pressure_ = 101319.9;   // Pa
  cuda_state.SetConcentration(foo, 20.0);           // mol m-3
```
- then we can establish a time loop for the cuda solver:
```
  cuda_state.PrintHeader();
  for (int i = 0; i < 10; ++i)
  {
    cuda_state.CalculateRateConstants(); // calculate rate constants on the host
    cuda_state.SyncToDevice(); // copy the updated rate constants to the device
    auto result = cuda_solver.Solve(500.0, cuda_state); // solve on the device
    cuda_state.variables_ = result.result_; // device to device copy with overloaded operator=
    cuda_state.SyncToHost();
    cuda_state.PrintState(i * 500);
  }
```

### LINKED ISSUE #529: Update the builder to create a rosenbrock solver
Adapt the `RosenbrockSolver` class to be created using the `SolverBuilder`

### LINKED ISSUE #538: Move the update state function to the state class and call it `CalculateRateConstant`

### LINKED ISSUE #539: Update the tests to use the solver builder

mattldawson:
Modifies the Rosenbrock solver to be creatable using the `SolverBuilder` class. This work coincidentally also removed the remaining template template parameters. Also, moves the rate constant calculation step out of the solver function, and modifies the GH Action for coverage to run on pushes to `main`, so that CodeCov diffs stay up-to-date

closes #529 
closes #538
closes #539 
closes #496
closes musica/134


include/micm/solver/linear_solver.hpp
```diff
@@ -57,6 +57,11 @@ namespace micm
     /// @brief default constructor
     LinearSolver(){};
 
+    LinearSolver(const LinearSolver&) = delete;
+    LinearSolver& operator=(const LinearSolver&) = delete;
```
sjsprecious, line 61: what is the reason to delete the copy constructor and assignment operator for almost all the class in this PR?

include/micm/solver/backward_euler.hpp
```diff
@@ -29,38 +28,47 @@ namespace micm
 {
 
   /// @brief An implementation of the fully implicit backward euler method
-  template<class LinearSolverPolicy, class ProcessSetPolicy>
+  template<class LinearSolverPolicy, class RatesPolicy>
```
sjsprecious, line 31: Is changing `ProcessSetPolicy` to `RatesPolicy` only a name change or its functionality also differs?

sjsprecious (commented):
@mattldawson can you give a brief flowchart or introduction about the general steps to use the new `solverbuilder` class to build a solver for CPU/GPU? Thanks.

include/micm/solver/solver_builder.hpp
```diff
@@ -17,102 +17,99 @@
 #include <micm/util/jacobian.hpp>
 #include <micm/util/matrix.hpp>
 #include <micm/util/sparse_matrix.hpp>
+#include <micm/util/vector_matrix.hpp>
 
 #include <system_error>
 #include <variant>
```
K20shores, line 23:
```suggestion
```

I think we no longer need this

test/integration/analytical_rosenbrock.cpp
```diff
@@ -13,120 +15,107 @@ using SparseMatrixTest = micm::SparseMatrix<double>;
 
 TEST(AnalyticalExamples, Troe)
 {
-  test_analytical_troe<micm::RosenbrockSolver<micm::Matrix, SparseMatrixTest>>();
+  auto builder = micm::CpuSolverBuilder<micm::RosenbrockSolverParameters>(micm::RosenbrockSolverParameters::ThreeStageRosenbrockParameters());
```
K20shores, line 18:
```suggestion
  auto rosenbrock = micm::CpuSolverBuilder<micm::RosenbrockSolverParameters>(micm::RosenbrockSolverParameters::ThreeStageRosenbrockParameters());
  auto backward_euler = micm::CpuSolverBuilder<micm::BackwardEulerParameters>(micm::BackwardEulerParameters::BackwardEulerParameters());
  test_analytical_troe(rosenbrock);
  test_analytical_troe(backward_euler);
```

Is something like this possible now? If so, can we add it?

include/micm/solver/solver.hpp
```diff
@@ -15,26 +17,30 @@ namespace micm
...
           number_of_species_(number_of_species),
           number_of_reactions_(number_of_reactions),
-          state_parameters_(state_parameters)
+          state_parameters_(state_parameters),
+          processes_(std::move(processes))
     {
     }
 
-    void Solve(double time_step, StatePolicy& state)
+    SolverResult Solve(double time_step, StatePolicy& state)
```
K20shores, line 41: Should we call `CalculateRateConstants` here so that it doesn't need to be called by the user? I'm not sure what the exact requirements for the gpu solver will be, but maybe they don't require a precondition to the solve function?

K20shores (commented):
Only a couple of comments, but I love all the red!

include/micm/solver/linear_solver.hpp
mattldawson: It was the combination of a few things. First, we don't actually need the copy constructor/assignment for anything we're currently doing. Second, for the JIT classes, we would have to figure out how to copy the generated JIT functions and resource managers, and it wasn't obvious how to do this. And lastly, for the remaining classes, I removed them so that the solver builder follows the same logic for all solvers (JIT and non-JIT).

include/micm/solver/backward_euler.hpp
mattldawson: It's just a name change, but I think `RatesPolicy` better reflects how objects of this type are now used. For example, the analytical tests implement this policy to provide the forcing and Jacobian functions for non-chemical systems.

mattldawson:
> @mattldawson can you give a brief flowchart or introduction about the general steps to use the new `solverbuilder` class to build a solver for CPU/GPU? Thanks.

A CPU solver is built in the README example here:
https://github.com/NCAR/micm/blob/6b973a054282386cf03a86fec916ec07b95cfe00/test/tutorial/test_README_example.cpp#L34-L37

I don't think we have a full CUDA Rosenbrock example yet in the tutorial, but I think it would follow logic similar to the JIT solver:
https://github.com/NCAR/micm/blob/6b973a054282386cf03a86fec916ec07b95cfe00/test/tutorial/test_jit_tutorial.cpp#L104-L108

include/micm/solver/solver.hpp
mattldawson: I think the reasoning for having the user do this is that for the CUDA solver, the rate constants have to be calculated on the host, and then synced to the device. Doing this in the `Solver` would require the `Solver` to do something special for a CUDA ODE solver. My vote would be to leave this in the user space and make an issue for calculating rate constants on the device. At that point we could move in into `Solver::Solve()` with no custom logic for CUDA solvers. What do you think?

include/micm/solver/backward_euler.hpp
```diff
@@ -29,38 +28,47 @@ namespace micm
...
+    /// @brief Solver parameters typename
+    using ParametersType = BackwardEulerSolverParameters;
+
     /// @brief Default constructor
+    /// @param parameters Solver parameters
+    /// @param linear_solver Linear solver
+    /// @param rates Rates calculator
+    /// @param jacobian Jacobian matrix
     BackwardEuler(
         BackwardEulerSolverParameters parameters,
```
boulderdaze, line 49: Can it be cons ref to avoid a copy of the arguement?
```suggestion
        const BackwardEulerSolverParameters& parameters,
```

boulderdaze (changes requested):
Looks good! I'm excited for all those changes this PR brought in

include/micm/solver/linear_solver.hpp
sjsprecious: Thanks for the clarification.

include/micm/solver/backward_euler.hpp
sjsprecious: Thanks for the clarification.

include/micm/solver/backward_euler_solver_parameters.hpp
```diff
@@ -11,10 +11,16 @@
...
 {
+  
+  template<class RatesPolicy, class LinearSolverPolicy>
+  class BackwardEuler;
 
   /// @brief BackwardEuler solver parameters
   struct BackwardEulerSolverParameters
   {
+    template<class RatesPolicy, class LinearSolverPolicy>
+    using SolverType = BackwardEuler<LinearSolverPolicy, RatesPolicy>;
```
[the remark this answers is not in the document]
mattldawson: see following comment...

include/micm/solver/cuda_rosenbrock.hpp
```diff
@@ -32,69 +30,76 @@
...
-  template<
-      template<class>
-      class MatrixPolicy,
-      class SparseMatrixPolicy,
-      class LinearSolverPolicy = CudaLinearSolver<SparseMatrixPolicy>,
-      class ProcessSetPolicy = CudaProcessSet>
-
-  class CudaRosenbrockSolver
-      : public RosenbrockSolver<MatrixPolicy, SparseMatrixPolicy, LinearSolverPolicy, ProcessSetPolicy>
+  template<class RatesPolicy, class LinearSolverPolicy>
```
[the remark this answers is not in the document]
mattldawson: done!

include/micm/solver/solver_builder.hpp
mattldawson: removed

test/integration/analytical_rosenbrock.cpp
mattldawson: It is now possible, but when I applied the analytical tests to the Backward Euler solver, none of them passed. They mostly ended up with answers that were close but not within the 1.0e-5 tolerance that we use in the tests.

I will open up a separate issue for testing the Backward Euler solver with the analytical systems.

include/micm/solver/rosenbrock.hpp
```diff
@@ -40,179 +39,96 @@
...
+        RosenbrockSolverParameters parameters,
+        LinearSolverPolicy linear_solver,
+        RatesPolicy rates,
+        auto& jacobian)
+        : parameters_(parameters),
+          linear_solver_(std::move(linear_solver)),
+          rates_(std::move(rates)),
+          jacobian_diagonal_elements_(jacobian.DiagonalIndices(0))
+    {
+    }
```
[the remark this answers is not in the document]
mattldawson: Thanks for catching that! fixed!

include/micm/solver/rosenbrock.hpp
```diff
@@ -40,179 +39,96 @@
...
+    /// @param linear_solver Linear solver
+    /// @param rates Rates calculator
+    /// @param jacobian Jacobian matrix
+    ///
+    /// Note: This constructor is not intended to be used directly. Instead, use the SolverBuilder to create a solver
     RosenbrockSolver(
-        const System& system,
-        const std::vector<Process>& processes,
-        const RosenbrockSolverParameters& parameters);
+        RosenbrockSolverParameters parameters,
```
boulderdaze, line 69: Looks good! Thank you for addressing the comments! Just one more comment but I don't need to review for this

Does it make sense to make it `const ref`? 
```suggestion
        const RosenbrockSolverParameters& parameters,
```

boulderdaze (approved):

test/integration/analytical_rosenbrock.cpp
sjsprecious: Note that the backward euler solver has a lower accuracy than the rosenbrock solver. Thus it is possible that it fails the same criteria used by the rosenbrock solver.

sjsprecious (commented):
Thanks @mattldawson for working on this giant PR. It is lots of work!

It seems that two tests on the windows platform fail the CI workflow somehow and the error message indicates an internal compiler error. Are they just old versions and should be removed? Or do they indicate a potential code bug somewhere?

include/micm/solver/solver.hpp
K20shores: That's fine with me

mattldawson:
> Thanks @mattldawson for working on this giant PR. It is lots of work!
> 
> It seems that two tests on the windows platform fail the CI workflow somehow and the error message indicates an internal compiler error. Are they just old versions and should be removed? Or do they indicate a potential code bug somewhere?

They just mean there is a bug in the compiler code. I don't think there is a way to infer anything beyond that. I will create a separate issue to look into that.

sjsprecious (approved):

K20shores:
Does this also close #496?

mattldawson:
> Does this also close #496 and #499?

I think it closes #496, but I didn't include the analytical tests for Backward Euler because they were failing (probably needs different solver tolerances, or lower expectations as @sjsprecious mentioned)

K20shores (approved):

Issue #575, opened 2024-06-20
Test all solvers against Robertson tests


K20shores:
All of the mentioned tests are very stiff systems.

## Acceptance Criteria
- [x] CPU Rosenbrock passes all these analytical tests (dense matrix)
- [x]  CPU Rosenbrock passes all these analytical tests (vector matrix)
- [x] CPU JIT Rosenbrock passes all these analytical tests
- [x] GPU Rosenbrock passes all these analytical tests

K20shores:
@sjsprecious do you think it's worth testing the cuda solver against these analytical tests? We'd have to write cuda kernels fro E5, Oregonator, and HIRES

sjsprecious:
@K20shores possibly not if they require a specific implementation of CUDA kernels. I think it is enough if we can get all the CUDA tests that are covered by the JIT tests working. This is just my two cents.

K20shores:
@sjsprecious I'll change it to just the robertson tests then; it won't require kernels

K20shores:
Closed by #596 when the vector tests were added to analytical policy

Issue #498, opened 2024-05-07
Mimic camchem substep convergence failure integration acceptance


K20shores:
If the Backward Euler solver fails to converge, the state at the end of the last successful solver time step should be returned.

**Acceptance Criteria**
- When the Backward Euler solver fails to converge, the state at the end of the last successful solver time step is returned.

**Ideas**
- The Robertson analytical example can be used to test this. As of now, backward euler fails to converge for the robertson example
- This block of code could be modified to set `state.variables_ = Yn` prior to throwing then error:
https://github.com/NCAR/micm/blob/b819e5770cd7e83d60d23181c2a90e1bbd660afb/include/micm/solver/backward_euler.inl#L171-L179

PR #576, opened 2024-06-20
testing backward euler with most of the analytical tests

### LINKED ISSUE #499: Refactor tests to use both solvers for all analytical tests

K20shores:
Closes #499 

Backward euler passes all tests, but the robertson tests (the solver fails to converge and seems to get stuck, will be investigated by #498). I don't know how to test backward euler against HIRES, Oregonator, E5, but hopefully that can be fixed with #575

### DIFF
--- a/test/integration/analytical_rosenbrock.cpp
+++ b/test/integration/analytical_rosenbrock.cpp
@@ -12,102 +12,93 @@
 
 using SparseMatrixTest = micm::SparseMatrix<double>;
 
+auto rosenbrock = micm::CpuSolverBuilder<micm::RosenbrockSolverParameters>(
+    micm::RosenbrockSolverParameters::ThreeStageRosenbrockParameters());
+auto backward_euler = micm::CpuSolverBuilder<micm::BackwardEulerSolverParameters>(
+    micm::BackwardEulerSolverParameters());
+
 TEST(AnalyticalExamples, Troe)
 {
-  auto builder = micm::CpuSolverBuilder<micm::RosenbrockSolverParameters>(
-      micm::RosenbrockSolverParameters::ThreeStageRosenbrockParameters());
-  test_analytical_troe(builder);
+  test_analytical_troe(rosenbrock);
+  test_analytical_troe(backward_euler, 1e-4);
 }
 
 TEST(AnalyticalExamples, TroeSuperStiffButAnalytical)
 {
-  auto builder = micm::CpuSolverBuilder<micm::RosenbrockSolverParameters>(
-      micm::RosenbrockSolverParameters::ThreeStageRosenbrockParameters());
-  test_analytical_stiff_troe(builder);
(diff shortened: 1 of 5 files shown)


test/integration/analytical_rosenbrock.cpp
```diff
@@ -12,102 +12,93 @@
...
+  // test_analytical_robertson(backward_euler, 1e-1);
 }
 
 TEST(AnalyticalExamples, SurfaceRxn)
 {
-  auto builder = micm::CpuSolverBuilder<micm::RosenbrockSolverParameters>(
-      micm::RosenbrockSolverParameters::ThreeStageRosenbrockParameters());
-  test_analytical_surface_rxn(builder);
+  test_analytical_surface_rxn(rosenbrock, 1e-5);
+  test_analytical_surface_rxn(backward_euler, 1e-1);
```
sjsprecious, line 101: `1e-1` seems large to me. This makes me think whether we should write a function of calculating the relative error instead of using `expect_near`, which compares the absolute difference.

K20shores: yeah these values are on the small side. I'll make that change

boulderdaze (commented):
How do you decide the tolerance values? It is interesting to see they vary between each solver type and rate constant type.

K20shores:
> How do you decide the tolerance values? It is interesting to see they vary between each solver type and rate constant type.

it was super duper scientific. I set a value of `1e-8` and when that failed for a particular test I subtracted one from the exponent until it passed.

A more intelligent way would be to play with the solver tolerances to choose values that minimize the error

boulderdaze:
> > How do you decide the tolerance values? It is interesting to see they vary between each solver type and rate constant type.
> 
> it was super duper scientific. I set a value of `1e-8` and when that failed for a particular test I subtracted one from the exponent until it passed.
> 
> A more intelligent way would be to play with the solver tolerances to choose values that minimize the error

Haha I see. Yeah that makes sense but would require a good amount of work to investigate that. I guess we don't need to worry about it for now

boulderdaze (approved):

test/integration/analytical_rosenbrock.cpp
K20shores: done, only was able to use 0.05 for the relative error though

mattldawson (approved):

sjsprecious (approved):

mwaxmonsky (approved):

PR #579, opened 2024-06-21
Add CUDA Rosenbrock tests

### LINKED ISSUE #540: Add test of CUDA Rosenbrock with device-resident state throughout solve
Add a test where the rate constants are calculated on the host, then all state data is copied to the device, the solve is performed, output is synched back to host.

**Acceptance Criteria**
- Tests of the CUDA Rosenbrock solver are included in the test suite and pass for all analytical chemistry systems (doesn't include E5, HIRES, Oregonator)
- CUDA Rosenbrock tests calculate rate constants on the host before solving on the device and then evaluating results back on the host

**Ideas**
- Add `SyncInputsToDevice()` and `SyncOutputsToHost()` functions on `micm::State` that can be called before and after the ODE solve that copy matrix data in the State between the host and device
- The https://github.com/NCAR/micm/blob/main/test/integration/analytical_jit_rosenbrock.cpp test could be used as a template for setting up the CUDA Rosenbrock tests

sjsprecious:
This PR adds the CUDA tests for all the analytical problems that are covered by the JIT tests.

We also remove unnecessary calls to the functions from the base class in a derived class.

All the GPU tests pass on Derecho's A100 GPU by using `nvhpc/23.7`.

fix #540


include/micm/solver/state.hpp
```diff
@@ -8,6 +8,7 @@
 #include <micm/util/jacobian.hpp>
 #include <micm/util/matrix.hpp>
 #include <micm/util/sparse_matrix.hpp>
+#include <micm/util/cuda_dense_matrix.hpp>
```
sjsprecious, line 11: @mattldawson @mwaxmonsky @K20shores because I add the `cuda_dense_matrix.hpp` header here, all the CI tests will now try to include the CUDA stuff. I think this is the reason most CI tests fail as those containers do not install CUDA. Is there a better way to handle the issue here?

mattldawson: You could extend this class with a `CudaState` class that implements the synching functions. You would probably also need to modify the `Solver` class to return the correct type of state class from the `GetState` function.

sjsprecious: Thanks @mattldawson . I have extended the `state` class to `cudastate` class at https://github.com/NCAR/micm/pull/579/commits/cd5735edde54872fb4e2a27ac5a278c73576c8d4 (hopefully I did it correctly) but I don't know how to revise the `Solver` class to get the `cudastate` for the CUDA tests. Can you provide more details here?

test/integration/analytical_cuda_rosenbrock.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+constexpr std::size_t L = 1;
+
+TEST(AnalyticalExamplesCudaRosenbrock, Troe)
+{
+  auto builder = micm::CudaSolverBuilder<micm::CudaRosenbrockSolverParameters, L>(
+      micm::RosenbrockSolverParameters::ThreeStageRosenbrockParameters());
+  test_analytical_troe(builder, [](auto& state) -> void { state.SyncInputsToDevice(); },
+                                [](auto& state) -> void { state.SyncOutputsToHost(); });
+}
```
K20shores, line 21: Is this really necessary? @mattldawson I know we are trying to be explicit about memory moves, but now we have a different interface for cpu vs gpu. Since it's required for memory to be moved to and from the host for the GPU, I think it makes sense for the cuda rosenbrock solver to do these operations in its own functions

test/integration/analytical_policy.hpp
```diff
@@ -73,8 +73,8 @@ using yields = std::pair<micm::Species, double>;
 
 using SparseMatrixTest = micm::SparseMatrix<double>;
 
-template<class BuilderPolicy>
-void test_analytical_troe(BuilderPolicy& builder)
+template<class BuilderPolicy, class PrepareFunc, class PostpareFunc>
+void test_analytical_troe(BuilderPolicy& builder, PrepareFunc prepare_for_solve, PostpareFunc postpare_for_solve)
```
K20shores, line 77: If we end up keeping the functions, can they be arguments instead of templates??
```suggestion
template<class BuilderPolicy>
void test_analytical_troe(BuilderPolicy& builder, auto prepare_for_solve = [](auto& state) { },, auto  postpare_for_solve = [](auto& state) { })
```

test/integration/analytical_policy.hpp
```diff
@@ -159,8 +159,10 @@ void test_analytical_troe(BuilderPolicy& builder)
   {
     times.push_back(time_step);
     solver.CalculateRateConstants(state);
+    prepare_for_solve(state);
     // Model results
     auto result = solver.Solve(time_step, state);
+    postpare_for_solve(state);
```
K20shores, line 165: These are functions that wrap a single line of code. I really think moving the data onto and off of the device should be done by the solver when it needs to.

K20shores (commented):
Having to modify so many of the tests to match the API needed to use the GPU solver when we are aiming to have the same API between GPU and CPU to me says we should rethink where the memory movement needs to happen

include/micm/solver/state.hpp
K20shores: We'll need to figure out how to make the solver builder return the appropriate state

https://github.com/NCAR/micm/blob/85232c5e80f1b0b3c06b6aa236c08e588b86ba2a/include/micm/solver/solver_builder.inl#L300-L307

test/integration/analytical_policy.hpp
sjsprecious: I had a similar discussion with Matt last Friday about this and he made a good point: if we run MICM in a CPU-only host model like CAM, then the data movement here is necessary to ensure correctness; if we run MICM in a fully GPU-enabled model like FastEddy, we actually do not need any data movement here. Thus it should really be the user's liability/obligation to decide when the data movement needs to happen. Hiding the data movement inside the `solver` may lead to unnecessary data movement and hurt the performance later.

include/micm/solver/state.hpp
K20shores: One way would be to add another function onto the builder, `BuildForGPU` that will do the same things as in this function, but return a `Solver<SolverPolicy, CudaState<DenseMatrixPolicy, SparseMatrixPolicy>>`

test/integration/analytical_policy.hpp
K20shores: Ah, I hadn't thought of that

include/micm/solver/state.hpp
sjsprecious: Thanks @mattldawson and @K20shores for your help. I have added the `CudaState` class and updated the cuda solver builder accordingly.

test/integration/analytical_policy.hpp
sjsprecious: I tried to use arguments but it did not compile. Thus I assume that we have to use templates here.

test/integration/analytical_cuda_rosenbrock.cpp
mattldawson: I think it will be important to have users do this themselves, because for a GPU-resident model like FastEddy they may do the copies at different times (or not at all) compared to a primarily CPU-based model that only uses GPUs for MICM. I also think it's reasonable to have a different API for `CudaState` than for `State`, and expect users to know that they have different responsibilities when `CudaState`. Doing memory transfers inside solver functions I think would lead to problems down the line.

include/micm/process/process.hpp
```diff
@@ -176,7 +176,6 @@ namespace micm
       const std::vector<Process>& processes,
       State<DenseMatrixPolicy, SparseMatrixPolicy>& state)
   {
-    MICM_PROFILE_FUNCTION();
```
mattldawson, line 179: was this removed by accident?

include/micm/solver/cuda_state.hpp
```diff
@@ -0,0 +1,39 @@
...
+  template<class DenseMatrixPolicy, class SparseMatrixPolicy>
+  struct CudaState : public State<DenseMatrixPolicy, SparseMatrixPolicy>
+  {
+    public:
+      CudaState(const CudaState&) = delete;
+      CudaState& operator=(const CudaState&) = delete;
+      CudaState(CudaState&&) = default;
+      CudaState& operator=(CudaState&&) = default;
+
+      /// @brief
```
mattldawson, line 22: this is missing the description

mattldawson (approved):
looks great! just two very minor comments, but I think it's fine to merge

include/micm/solver/cuda_state.hpp
sjsprecious: Thanks. Just added.

include/micm/process/process.hpp
sjsprecious: Good catch. Just put it back.

test/integration/analytical_policy.hpp
```diff
@@ -78,8 +78,9 @@ using yields = std::pair<micm::Species, double>;
 
 using SparseMatrixTest = micm::SparseMatrix<double>;
 
-template<class BuilderPolicy>
-void test_analytical_troe(BuilderPolicy& builder, double tolerance = 1e-8)
+
+template<class BuilderPolicy, class PrepareFunc, class PostpareFunc>
+void test_analytical_troe(BuilderPolicy& builder, PrepareFunc prepare_for_solve, PostpareFunc postpare_for_solve, double tolerance = 1e-8)
```
K20shores, line 83: Do we need templates for the before and after functions? Couldn't we just have arguements with default values so that the CPU tests don't have to pass in empty functions?

K20shores (approved):

PR #581, opened 2024-06-24
read version in the docs from the cmake files

K20shores:
Now we won't have to manually update the version in the docs


.github/workflows/pr_documentation_link.yml
```diff
@@ -0,0 +1,31 @@
...
+  pull_request:
+    types: [opened]
+
+jobs:
+  add-doc-link:
+    runs-on: ubuntu-latest
+    
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
```
boulderdaze, line 14: This feature would be nice! Could you update the version to `v4`? 
```suggestion
      uses: actions/checkout@v4
```

boulderdaze (commented):
This feature would be nice!

.github/workflows/pr_documentation_link.yml
K20shores: I couldn't get the action to work so I deleted the action

boulderdaze (approved):

mattldawson (approved):

Issue #578, opened 2024-06-21
Inspect the Robertson integration test for accuracy


K20shores:
We pulled many of our analytical tests from [this website](https://www.unige.ch/~hairer/testset/testset.html). One of them is the Robertson.

During #576 @sjsprecious asked me to try to reduce the tolerance for `test_analytical_surface_rxn`, but that made me think of the tolerance we had on robertson which was 0.1. This seemed like a large tolerance.

So, I mimiced [the fortran driver code](https://www.unige.ch/~hairer/testset/stiff/rober/driver_rodas.f) in c++ with micm to test a range of solver tolerances and we found that for some reason we plateau at a relative error of 1e-1.

![three_stage_literal_translation](https://github.com/NCAR/micm/assets/13836216/0ef72051-22d1-4fc6-9bd1-9972c1ba8575)

However, this is probably fine since the concentrations of A and B are essentially zero at the end of the simulation and so the large percent difference doesn't really affect anything. However, the lack of effect of the different tolerances mathematically was worrying. 

If in the driver code, at the beginning of each time step we reset the initial concentrations and solve, we get a different distribution of relative errors that's more expected.

![three_stage_reset_contribution](https://github.com/NCAR/micm/assets/13836216/80db9cfe-fe95-49be-81bc-9518ad032bb1)

We need to figure out if we are actually testing what we really think we are in the Robertson analytical test or not. The branch that contains the robertson test is `many_solver_tolerances`. In case that gets deleted, here's the modified robertson test function

```
template<class BuilderPolicy>
void test_analytical_robertson(BuilderPolicy& builder, double tolerance = 1e-8)
{
  /*
   * A -> B, k1 = 0.04
   * B + B -> C + B, k2 = 3e7
   * B + C -> A + C, k3 = 1e4
   *
   * this problem is described in
   * Hairer, E., Wanner, G., 1996. Solving Ordinary Differential Equations II: Stiff and Differential-Algebraic Problems, 2nd
   * edition. ed. Springer, Berlin ; New York. Page 3
   *
   * solutions are provided here
   * https://www.unige.ch/~hairer/testset/testset.html
   */

  auto a = micm::Species("A");
  auto b = micm::Species("B");
  auto c = micm::Species("C");

  micm::Phase gas_phase{ std::vector<micm::Species>{ a, b, c } };

  micm::Process r1 = micm::Process::Create()
                         .SetReactants({ a })
                         .SetProducts({ Yields(b, 1) })
[...]

PR #582, opened 2024-06-25
498 mimic camchem substep convergence failure integration acceptance

### LINKED ISSUE #498: Mimic camchem substep convergence failure integration acceptance

K20shores:
Closes #498 

This also tests backward euler against hires and e5. The oregonator systems is apparently too stiff for backward euler.

From my understanding, if you fail to converge too many times in a row and try to cut the time step too many times in camchem, you forgo the current integration, but accept what you have done up to the last successful integration and continue on.

I don't actually know how to test this


include/micm/solver/backward_euler.inl
```diff
@@ -157,29 +157,22 @@ namespace micm
...
-          // TODO: continue on with the current solution to get the full solution
-          n_convergence_failures = 0;
-          // give_up = true;
-          t += H;
-          throw std::system_error(make_error_code(MicmBackwardEulerErrc::FailedToConverge), "Failed to converge");
-        };
-
-        H *= time_step_reductions[n_convergence_failures++];
+          std::cerr << "Failed to converge too many times in a row. Accepting the current integration and continuing on.\n";
+          break;
```
sjsprecious, line 169: according to https://github.com/ESCOMP/CAM/blob/cam_development/src/chemistry/pp_trop_strat_mam4_vbs/mo_imp_sol.F90#L385-L397, we should advance `t` to `t+H` here. The comments are confusing to me as we only advance the time step, but do not accept the changed solution (i.e., the `base_sol` in the F90 code is not updated).

include/micm/solver/backward_euler.inl
```diff
@@ -194,13 +187,13 @@ namespace micm
           n_successful_integrations = 0;
           H *= 2.0;
         }
+        // Don't let H go past the time step
+        H = std::min(H, time_step - t);
```
sjsprecious, line 191: I think this check is needed for both `if` and `else` blocks, so it should be moved to its original place?

K20shores: well, the if block reduces the time step (even though it's multiplying something), but there's no harm in me moving it back to how it was

include/micm/solver/backward_euler.inl
K20shores: ah good catch. I thought I was missing something but struggled to understand the linked code

include/micm/solver/backward_euler.inl
```diff
@@ -195,12 +189,12 @@ namespace micm
           H *= 2.0;
         }
       }
-
       // Don't let H go past the time step
       H = std::min(H, time_step - t);
     }
```
sjsprecious, line 194: it will be great if we could add some comments after these `}` to indicate which loop/if-else block it ends.

test/integration/analytical_rosenbrock.cpp
```diff
@@ -91,7 +91,7 @@ TEST(AnalyticalExamples, BranchedSuperStiffButAnalytical)
 TEST(AnalyticalExamples, Robertson)
 {
   test_analytical_robertson(rosenbrock, 1e-1);
-  // test_analytical_robertson(backward_euler, 1e-1);
+  test_analytical_robertson(backward_euler, 1e-1);
```
sjsprecious, line 94: Interesting that backward euler solver could pass the test with the same tolerance as rosenbrock solver.

test/integration/analytical_rosenbrock.cpp
```diff
@@ -104,243 +104,45 @@ using LinearSolverTest = micm::LinearSolver<SparseMatrixTest, micm::LuDecomposit
...
-  {
-    double rel_diff = relative_difference(model_concentrations[i][_a], analytical_concentrations[i][0]);
-    EXPECT_TRUE(rel_diff < tol) << "Arrays differ at index (" << i << ", " << 0 << ")";
-    rel_diff = relative_difference(model_concentrations[i][_b], analytical_concentrations[i][1]);
-    EXPECT_TRUE(rel_diff < tol) << "Arrays differ at index (" << i << ", " << 1 << ")";
-    rel_diff = relative_difference(model_concentrations[i][_c], analytical_concentrations[i][2]);
-    EXPECT_TRUE(rel_diff < tol) << "Arrays differ at index (" << i << ", " << 2 << ")";
-  }
+  test_analytical_oregonator(rosenbrock_solver, 1e-3);
+  // test_analytical_oregonator(backward_euler_solver, 1e-1);
```
sjsprecious, line 123: Does the backward euler solver not work for the Oregonator test?

test/integration/analytical_rosenbrock.cpp
```diff
@@ -104,243 +104,45 @@ using LinearSolverTest = micm::LinearSolver<SparseMatrixTest, micm::LuDecomposit
...
-  };
-
-  auto state = solver.rates_.GetState();
 
-  state.variables_[0] = model_concentrations[0];
+  auto rosenbrock_solver = OregonatorTest::template CreateSolver<RosenbrockTest<OregonatorTest>, LinearSolverTest>(params, 1);
+  auto backward_euler_params = micm::BackwardEulerSolverParameters();
+  double abolute_tolerance = 1e-16;
+  backward_euler_params.absolute_tolerance_ = { abolute_tolerance, abolute_tolerance, abolute_tolerance };
+  backward_euler_params.relative_tolerance_ = 1e-10;
```
sjsprecious, line 119: Looks like you only override the `backward_euler_params` for this test. Is there a specific reason to do so?

test/integration/analytical_rosenbrock.cpp
K20shores: It does not. It's not even close to working. The very first time step we already have almost a 100% relative difference

test/integration/analytical_rosenbrock.cpp
K20shores: I was trying different tolerances to see if I could get backward euler to pass, but nothing I tried seemed to work

include/micm/solver/backward_euler.inl
K20shores: I'm not sure that this is a style we want to support. Code folding is more useful than comments, I feel, since it allows whomever is reading to choose which parts of they control flow they want to hide

test/integration/analytical_rosenbrock.cpp
K20shores: I thought so, too. I think it's the way we are comparing the numbers, but anything I've tried always leaves us with the same tolerance. Maybe in resolving #578 we could come up with a better way to compare these numbers, or figure out some set of tolerances that make us match the analytical results directly

boulderdaze (approved):
Looks good!

test/integration/analytical_rosenbrock.cpp
K20shores: For reference, these are the equations that compose the Oregonator

Screenshot 2024-06-26 at 12 55 25 PM

Image from 
Hairer, E., Wanner, G., 1996. Solving Ordinary Differential Equations II: Stiff and Differential-Algebraic Problems, 2nd edition. ed. Springer, Berlin ; New York.

include/micm/solver/backward_euler.inl
```diff
@@ -157,32 +162,27 @@ namespace micm
...
-          // we have failed to converge, accept the solution
-          // TODO: continue on with the current solution to get the full solution
-          n_convergence_failures = 0;
-          // give_up = true;
           t += H;
-          throw std::system_error(make_error_code(MicmBackwardEulerErrc::FailedToConverge), "Failed to converge");
-        };
-
-        H *= time_step_reductions[n_convergence_failures++];
+          std::cerr << "Failed to converge too many times in a row. Accepting the current integration and continuing on.\n";
```
mattldawson, line 175: I think we should avoid writing to stdout and stderr in favor of letting host applications handle all logging information. Why not throw the FailedToConverge error as before?

include/micm/solver/backward_euler.inl
```diff
@@ -157,32 +162,27 @@ namespace micm
...
+          break;
+        }
+        else {
+          std::cerr << "Failed to converge. Reducing the time step.\n";
```
mattldawson, line 179:
```suggestion
```

mattldawson (changes requested):
looks great! just one suggestion for the error output

test/integration/analytical_rosenbrock.cpp
sjsprecious: Oh, I did not realize that this is for the Robertson test. We surely need to understand the #578 issue more first. Thanks for the clarificadtion.

test/integration/analytical_rosenbrock.cpp
sjsprecious: Thanks for the additional information. Not sure why it does not work for the BE solver but this may highlight the importance of using high-order numerical solver for the chemistry problem.

sjsprecious: Probably just remove this test for BE solver since it is not working?

test/integration/analytical_rosenbrock.cpp
sjsprecious: If we decide to remove the test for OREGO, we probably do not need this change either?

include/micm/solver/backward_euler.inl
K20shores: Camchem assumes we can continue on even when the integration fails. Throwing an error would prevent us from checking our implementation of backward euler against camcehm. I'm not really sure what we should do. What if we add another state to the result enum `AcceptingUnconvergedIntegration`? @sjsprecious @mattldawson

test/integration/analytical_rosenbrock.cpp
K20shores: removed

test/integration/analytical_rosenbrock.cpp
K20shores: agreeed. It's been removed.

include/micm/solver/backward_euler.inl
sjsprecious: You are correct that CAM-Chem can continue the runs even if the solver does not converge for this time step. It will write out a warning message to the log file. Adding another state to the result seems fine to me.

sjsprecious (approved):

mattldawson (approved):

PR #585, opened 2024-06-28
In-place linear solve

### LINKED ISSUE #492: Condense the first two arguments of the linear solver so that we don't need to make a copy
The first and second argument of the linear solver are the same and we always have to make a [copy](https://github.com/NCAR/micm/blob/main/include/micm/solver/rosenbrock.inl#L369-L370). In #485 the forcing vector was just passed in for both arguments and things worked. That means we can remove one of the arguments and prevent a copy. 

**Acceptance**
- There is only one matrix passed into linear solve
- The copy of the matrix is removed from the CPU and GPU versions of Rosenbrock

K20shores:
Closes #492 

Solving  $Ax=B$ with the linear solve can be done in place. $x$ was always a copy of $b$ and we would copy $b$ into places in the linear solve. This removes the $b$ argument from all versions of the linear solver and reduces the memory footprint by one matrix


mattldawson (approved):
looks great! my only suggestion would be to add a description of the parameters to the base class solve function and specify that `b` should be passed in as `x` and will be overwritten

sjsprecious (approved):
Thanks for working on this.

PR #593, opened 2024-07-01
577 test all parameter types of the dense matrix cpu rosenbrock on the analytical policy tests

### LINKED ISSUE #577: Test all parameter types of the dense matrix cpu rosenbrock on the analytical policy tests
## Acceptance criteria

All stage types available are tested against the analytical policy tests

- 2 stage
- 3 stage
- 4 stage
- 4 stage differential algebraic
- 6 stage differential algebraic

K20shores:
Closes #577

- Converts HIRES, Oregonator, E5 to chemical equations so that they can be tested on the GPU as well
- Many analytical tests now test for the relative difference. Some don't. Why? Because sometimes expect near was easier to use than the relative difference


test/integration/cuda/test_cuda_analytical_rosenbrock.cpp
```diff
@@ -18,6 +18,7 @@ TEST(AnalyticalExamplesCudaRosenbrock, Troe)
       micm::RosenbrockSolverParameters::ThreeStageRosenbrockParameters());
```
sjsprecious, line 18: Looks like all the CUDA tests fail to compile. Also, do we want to test all the rosenbrock parameters for the CUDA tests as well?

test/integration/analytical_rosenbrock.cpp
```diff
@@ -12,120 +12,156 @@
...
-  test_analytical_troe(
-      rosenbrock, [](auto& state) {}, [](auto& state) {});
-  test_analytical_troe(
-      backward_euler, [](auto& state) {}, [](auto& state) {}, 1e-4);
+  test_analytical_troe(rosenbrock_2stage, 1e-7);
+  test_analytical_troe(rosenbrock_3stage);
+  test_analytical_troe(rosenbrock_4stage);
+  test_analytical_troe(rosenbrock_4stage_da);
+  test_analytical_troe(rosenbrock_6stage_da);
+  test_analytical_troe(backward_euler, 1e-4);
```
sjsprecious, line 34: Should all the backward euler solver test be put to a separate file instead?

K20shores: Yeah, probably. I'll do that

test/integration/cuda/test_cuda_analytical_rosenbrock.cpp
```diff
@@ -3,160 +3,212 @@
 
 #include "../analytical_policy.hpp"
 #include "../analytical_surface_rxn_policy.hpp"
+#include "../oregonator.hpp"
+#include "../e5.hpp"
+#include "../hires.hpp"
```
sjsprecious, line 8: All these three tests fail on Derecho's A100 GPU somehow.

sjsprecious: After discussion with Kyle, we do not want to include these three tests into the CUDA integration test suite because they hardcode the Jacobian and forcing terms, and we need to write specific CUDA kernels for the GPU tests.

boulderdaze (approved):
Looks good!

mattldawson (approved):
nice!

test/integration/test_analytical_backward_euler.cpp
K20shores, line 1: @sjsprecious @mattldawson the backward euler tests have to pass with a relative error of 0.5, some with a relative error of 1. Most of the 0.5 are for very early on in the integration. 1 is restricted to some of the analytical tests. I suppose I'd have to add in the absolute tolerance checks as well to lower this

test/integration/analytical_policy.hpp
```diff
@@ -704,32 +701,32 @@ void test_analytical_ternary_chemical_activation(
   }
 
   std::vector<std::string> header = { "time", "A", "B", "C" };
-  writeCSV("analytical_concentrations.csv", header, analytical_concentrations, times);
-  writeCSV("model_concentrations.csv", header, model_concentrations, times);
+  writeCSV("tca_analytical_concentrations.csv", header, analytical_concentrations, times);
+  writeCSV("tca_model_concentrations.csv", header, model_concentrations, times);
```
mattldawson, line 705: is writing the results to csv files needed for the test? if so, does this prevent us from running tests in parallel?

test/unit/jit/solver/test_jit_solver_builder.cpp
```diff
@@ -0,0 +1,69 @@
...
+  //                           .SetNumberOfGridCells(1)
+  //                           .SolverParameters(micm::BackwardEulerSolverParameters{})
+  //                           .Build();
+
+  // auto cuda_rosenbrock = micm::CudaSolverBuilder<L>()
+  //                           .SetSystem(the_system)
+  //                           .SetReactions(reactions)
+  //                           .SetNumberOfGridCells(1)
+  //                           .SolverParameters(micm::ThreeStageRosenbockSolverParameters{})
+  //                           .Build();
```
mattldawson, line 68: can this be uncommented now? or is there still more to do for building CUDA Rosenbrock solvers?

mattldawson (approved):
looks great! just a couple minor questions, but looks good to go to me!

test/integration/analytical_policy.hpp
K20shores: it's not needed. We can run them in parallel and they'll overwrite each others results but these are only used when looking at the output. I changed the name so that other tests don't overwrite these results

test/unit/jit/solver/test_jit_solver_builder.cpp
K20shores: This test does exist in test/unit/cuda/test_cuda_solver_builder.cpp. This file is in the jit and we don't need this code

sjsprecious (approved):
Thanks @K20shores for working on this giant PR. It looks great!

Also @K20shores reports potential memory leak for the analytical CUDA tests and it should be addressed in a separate effort.

test/integration/analytical_policy.hpp
```diff
@@ -1923,55 +2113,143 @@ void test_analytical_hires(auto& solver, double tolerance = 1e-8)
...
-   * This problem is described in
+   * A1 -> A2 + A3,  k1 = 7.89e-10
+   * A2 + A3 -> A5,  k2 = 1.13e9
+   * A1 + A3 -> A4,  k3 = 1.1e7
+   * A4 -> A3 + A6,  k4 = 1.13e3
+   * 
+   * this problem is described in
    * Hairer, E., Wanner, G., 1996. Solving Ordinary Differential Equations II: Stiff and Differential-Algebraic Problems, 2nd
-   * edition. ed. Springer, Berlin ; New York. Page 145
+   * edition. ed. Springer, Berlin ; New York. Page 3
```
mwaxmonsky, line 2172: Possibly add DOI if this is correct?
https://doi.org/10.1007/978-3-642-05221-7

mwaxmonsky (approved):
Great work @K20shores!  Just one minor comment but otherwise looks good to me!

boulderdaze (approved):
Looks good!

PR #594, opened 2024-07-02
removing reset of L and U matrices

### LINKED ISSUE #587: Check if resetting the L and U matrices before solving with Rosenbrock is needed
Currently the lower and upper matrices on the `State` are reset to zero at the beginning of the Rosenbrock solve, but this shouldn't actually be necessary. It was likely just causing an error before the normalized error function was set to properly work with vectorizable matrices where the number of grid cells was not even divisible by the vector dimension. See if this can be removed now.

**Acceptance Criteria**
- The reseting of the lower and upper matrices at the beginning of the solve is removed or an explanation of why it is necessary is included as a comment in the code

K20shores:
Closes #587 

All tests pass after removing the reset of the matrices

### DIFF
--- a/include/micm/solver/rosenbrock.inl
+++ b/include/micm/solver/rosenbrock.inl
@@ -13,9 +13,6 @@ namespace micm
 
     SolverResult result{};
     result.state_ = SolverState::Running;
-    // reset the upper, lower matrix. (this may not be needed anymore. see https://github.com/NCAR/micm/issues/587)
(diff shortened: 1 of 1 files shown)


boulderdaze (approved):

mattldawson (approved):
great!

mwaxmonsky (approved):

sjsprecious (approved):

PR #596, opened 2024-07-09
Backware Euler with vectorizable matrix types

### LINKED ISSUE #500: Adapt Backward Euler to work with vectorizable matrix types
Backward Euler currently uses iterators for the underlying `std::vector` of matrix types to loop over matrix elements. This works with the standard matrix type, but for vectorizable matrices will include buffer elements that should not be operated on.

**Acceptance Criteria**
- The Backward Euler solver works with vectorizable dense and sparse matrices, including when the number of grid cells is not evenly dividable by the vector dimension of the matrices
- Buffer elements in the vector matrices are not accessed during the Backward Euler solve

**Ideas**
- Implement math operations for the matrix classes (`+`, `-`, `/`. `*`) - this will likely be challenging because of temporary variables. Maybe we could try this out during a Hack-a-thon. Focus on VectorMatrix types first.
- Or, use the `ForEach()` or `Axpy()` function in place of existing loops (although this won't work with an eventual CUDA BE solver)

mattldawson:
Adapts the Backward Euler solver to work with vectorizable matrices.

closes #500 

Also:
- moves physical constants and labels into dedicated namespaces
- adapts analytical tests to solve for multiple grid cells
- results in a net increase of only one line of code!


test/integration/jit/test_jit_analytical_rosenbrock.cpp
```diff
@@ -4,125 +4,134 @@
...
+auto param_six_da = micm::RosenbrockSolverParameters::SixStageDifferentialAlgebraicRosenbrockParameters();
 
 TEST(AnalyticalExamplesJitRosenbrock, Troe)
 {
-  test_analytical_troe<builderType, stateType>(two);
-  test_analytical_troe<builderType, stateType>(three);
-  test_analytical_troe<builderType, stateType>(four);
-  test_analytical_troe<builderType, stateType>(four_da);
-  test_analytical_troe<builderType, stateType>(six_da);
+  test_analytical_troe<BuilderType<NUM_CELLS>, StateType<NUM_CELLS>>(BuilderType<NUM_CELLS>(param_two), 1e-3);
```
K20shores, line 31: Where does `NUM_CELLS` come from?

K20shores (approved):
Nice! Only a few questions, but nothing that would hold this up

include/micm/util/property_keys.hpp
```diff
@@ -4,5 +4,11 @@
...
 #include <string>
 
-static const std::string GAS_DIFFUSION_COEFFICIENT = "diffusion coefficient [m2 s-1]";
-static const std::string MOLECULAR_WEIGHT = "molecular weight [kg mol-1]";
+namespace micm
+{
+  namespace keys
+  {
+    static const std::string GAS_DIFFUSION_COEFFICIENT = "diffusion coefficient [m2 s-1]";
+    static const std::string MOLECULAR_WEIGHT = "molecular weight [kg mol-1]";
```
boulderdaze, line 12: Could the namespace be `property_keys` or something more specific? I think `keys` maybe too general.

include/micm/solver/backward_euler.inl
```diff
@@ -89,46 +86,30 @@ namespace micm
...
-        }
-        for (std::size_t i_block = 0; i_block < state.jacobian_.NumberOfBlocks(); ++i_block)
-        {
-          auto jacobian_vector = std::next(state.jacobian_.AsVector().begin(), i_block * state.jacobian_.FlatBlockSize());
-          for (const auto& i_elem : jacobian_diagonal_elements_)
-            jacobian_vector[i_elem] -= 1 / H;
-        }
-
+        // add the inverse of the time step from the diagonal
+        state.jacobian_.AddToDiagonal(1 / H);
```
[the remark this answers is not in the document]
sjsprecious: I think we need to negate the Jacobian matrix for Backward Euler solver first and then add `-1/H` as Kyle suggested. Probably this explains why you need to use such a higher tolerance for the unit test of Backward Euler solver.

sjsprecious (changes requested):
Thanks @mattldawson for working on this giant PR and this functionality looks great. Just have one question as listed below.

boulderdaze (approved):
Looks good! I don't see anything other than what others have commented on. I left one comment but it shouldn't hold this PR getting merged.

include/micm/solver/backward_euler.inl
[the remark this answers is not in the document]
sjsprecious: After Matt explains the details, these changes seem correct (look at the corresponding changes at line 111).

sjsprecious:
Running on Derecho with NVHPC, I got the following failed tests:
```
         34 - analytical_rosenbrock_integration (Failed)
         35 - analytical_backward_euler (Failed)
         38 - cuda_analytical_rosenbrock_integration (Subprocess aborted)
```
Looking at the output log, the difference is slightly above the given tolerance. I wonder whether we can add a check to the relative error (like https://github.com/NCAR/micm/blob/main/test/unit/solver/test_rosenbrock.cpp#L138-L139) and pass the test if the relative error is small enough (say < 1.e-14?).

test/unit/util/test_sparse_matrix_policy.hpp
```diff
@@ -500,3 +500,37 @@ MatrixPolicy<double, OrderingPolicy> testSetScalar()
...
+  matrix[0][1][1] = 1;
+  matrix[1][1][1] = 3;
+  matrix[2][1][1] = 5;
+
+  matrix.AddToDiagonal(7);
+
+  EXPECT_EQ(matrix[0][1][1], 8);
+  EXPECT_EQ(matrix[1][1][1], 10);
+  EXPECT_EQ(matrix[2][1][1], 12);
+
```
mwaxmonsky, line 534: We might want to add some checks on the non-diagonal to make sure we are only incrementing the expected indices.

mwaxmonsky (approved):
Looks great @mattldawson!  Just one quick recommendation but nothing worth holding things up.

test/integration/jit/test_jit_analytical_rosenbrock.cpp
mattldawson: it's from `analytical_policy.hpp`

include/micm/solver/backward_euler.inl
```diff
@@ -43,25 +43,22 @@ inline std::error_code make_error_code(MicmBackwardEulerErrc e)
...
-    // the number of time step reduction is controlled by the time_step_reductions parameter
+    // A series of time step reductions are used after failed solves to try to find a solution
+    // These reductions are controlled by the time_step_reductions parameter in the solver parameters
+    // if the last attempt to reduce the timestep fails,
+    // accept the current H but do not update the Yn vector
 
     SolverResult result;
 
-    double small = parameters_.small;
+    double small = parameters_.small_;
```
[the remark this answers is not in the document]
mattldawson: done!

include/micm/util/property_keys.hpp
mattldawson: agreed. changed!

test/unit/util/test_sparse_matrix_policy.hpp
mattldawson: added!

mattldawson:
> Running on Derecho with NVHPC, I got the following failed tests:
> 
> ```
>          34 - analytical_rosenbrock_integration (Failed)
>          35 - analytical_backward_euler (Failed)
>          38 - cuda_analytical_rosenbrock_integration (Subprocess aborted)
> ```
> 
> Looking at the output log, the difference is slightly above the given tolerance. I wonder whether we can add a check to the relative error (like https://github.com/NCAR/micm/blob/main/test/unit/solver/test_rosenbrock.cpp#L138-L139) and pass the test if the relative error is small enough (say < 1.e-14?).

I modified the analytical policy tests to include a relative error check and the two CPU tests pass now. Still waiting on the CUDA analytical test to run.

sjsprecious (approved):
Thanks @mattldawson for working on this and addressing all the comments. I could confirm that all the tests passed on Derecho's A100 GPU.

PR #600, opened 2024-07-10
Fix GPU memory leak for the CUDA unit tests

### LINKED ISSUE #598: GPU memory leak for the CUDA kernels
As identified by @K20shores in his PR #593 , there are many memory leaks for the GPU tests that are not captured correctly by `compute-sanitizer`. Kyle has found one leak at https://github.com/NCAR/micm/blob/main/src/solver/lu_decomposition.cu#L184 but there are more.

**Acceptance Criteria**
- All the GPU memory tests are passed with the `--leak-check full` option for `compute-sanitizer`.

sjsprecious:
This PR:
- use the correct option `--leak-check full` for `compute-sanitizer` to check the memory leak on the GPU.
- fix a GPU memory leak due to the missing of `cudaFree` for a bool pointer on the GPU.
- fix multiple GPU memory leaks by using `std::swap` rather than `std::move` in the move constructor and assignment in a CUDA class.

All the GPU unit tests are now passed for GPU memory check on Derecho's A100 GPU. The integration test is not checked for GPU memory leak because it takes too long.

fix #598

### DIFF
--- a/cmake/test_util.cmake
+++ b/cmake/test_util.cmake
@@ -4,7 +4,7 @@
 if(MICM_ENABLE_MEMCHECK)
   if(MICM_ENABLE_CUDA)
     find_program(MEMORYCHECK_COMMAND "compute-sanitizer")
-    set(MEMORYCHECK_COMMAND_OPTIONS "--error-exitcode=1 --show-backtrace device --tool=memcheck --launch-timeout=0")
+    set(MEMORYCHECK_COMMAND_OPTIONS "--error-exitcode=1 --leak-check full --show-backtrace device --launch-timeout=0")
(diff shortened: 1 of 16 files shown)


K20shores (approved):
neat!

mwaxmonsky (approved):
Great work @sjsprecious!

mattldawson (approved):
looks great!

sjsprecious:
@mattldawson @K20shores @mwaxmonsky I ran the GPU memory leak check for the CUDA integration test on Derecho's A100 GPU. It passed the check within 10963461 ms (~3 hours and 10 minutes).

PR #603, opened 2024-07-15
572 check for singularity when the solver parameters flag is turned on

### LINKED ISSUE #572: Check for singularity when the solver parameters flag is turned on
There is a flag in `RosenbrockSolverParameters` to allow checking for singularity, but it currently doesn't do anything. Check for singularity when this flag is enabled.

**Acceptance Criteria**
- A singular matrix will result in a convergence failure for the current solver time step when the `RosenbrockSolverParameters::check_singularity_` flag is `true`

**Ideas**
- Possibly add a `LinearSolver::SafeFactor()` function that checks for singularity during the factoring, and can be called here:
https://github.com/NCAR/micm/blob/e076450a72532f1c780b47fc4e16a44dd8ba5e2b/include/micm/solver/rosenbrock.inl#L262-L263
- I think checking for singularity means checking whether the denominators here:
https://github.com/NCAR/micm/blob/e076450a72532f1c780b47fc4e16a44dd8ba5e2b/include/micm/solver/linear_solver.inl#L165
and here:
https://github.com/NCAR/micm/blob/e076450a72532f1c780b47fc4e16a44dd8ba5e2b/include/micm/solver/linear_solver.inl#L181
are 0, but we should ask a math person.
- The `JitLinearSolver` and `CudaLinearSolver` could probably override `SafeFactor` and just throw a "not yet implemented" error until we decide this would be useful to implement with JIT and/or on the GPUs

K20shores:
Closes #572 

This PR fixes an unknown bug where a zero in the bottom right position of the U matrix (which leads to a singular matrix) was not detected by always checking for zero in the bottom right element. Additionally, two tests are added to check for singular matrices when running rosenbrock

### DIFF
--- a/include/micm/solver/lu_decomposition.hpp
+++ b/include/micm/solver/lu_decomposition.hpp
@@ -112,16 +112,6 @@ namespace micm
     /// @param A Sparse matrix to decompose
     /// @param L The lower triangular matrix created by decomposition
     /// @param U The upper triangular matrix created by decomposition
-    template<class SparseMatrixPolicy>
-    requires(SparseMatrixConcept<SparseMatrixPolicy>) void Decompose(
-        const SparseMatrixPolicy& A,
-        SparseMatrixPolicy& L,
-        SparseMatrixPolicy& U) const;
-
-    /// @brief Perform an LU decomposition on a given A matrix
-    /// @param A Sparse matrix to decompose
-    /// @param L The lower triangular matrix created by decomposition
-    /// @param U The upper triangular matrix created by decomposition
     /// @param is_singular Flag that is set to true if A is singular; false otherwise
     template<class SparseMatrixPolicy>
     requires(!VectorizableSparse<SparseMatrixPolicy>) void Decompose(
--- a/include/micm/solver/lu_decomposition.inl
+++ b/include/micm/solver/lu_decomposition.inl
@@ -164,16 +164,6 @@ namespace micm
   }
 
   template<class SparseMatrixPolicy>
-  requires(SparseMatrixConcept<SparseMatrixPolicy>) inline void LuDecomposition::Decompose(
-      const SparseMatrixPolicy& A,
-      SparseMatrixPolicy& L,
-      SparseMatrixPolicy& U) const
-  {
-    bool is_singular = false;
-    Decompose<SparseMatrixPolicy>(A, L, U, is_singular);
-  }
-
-  template<class SparseMatrixPolicy>
   requires(!VectorizableSparse<SparseMatrixPolicy>) inline void LuDecomposition::Decompose(
       const SparseMatrixPolicy& A,
       SparseMatrixPolicy& L,
@@ -328,7 +318,9 @@ namespace micm
       }
 
       auto cell_U_bottom_right = std::next(U.AsVector().begin(), i_group * U_GroupSizeOfFlatBlockSize + U_GroupSizeOfFlatBlockSize - 1);
-      if (*cell_U_bottom_right == 0) is_singular = true;
+        if (*cell_U_bottom_right == 0){
+            is_singular = true;
+        }
     }
   }
 
--- a/include/micm/solver/lu_decomposition.inl
+++ b/include/micm/solver/lu_decomposition.inl
@@ -223,16 +223,19 @@ namespace micm
             L_vector[lki_nkj->first] -= L_vector[lkj_uji->first] * U_vector[lkj_uji->second];
             ++lkj_uji;
           }
+
           if (U_vector[*uii] == 0.0)
           {
             is_singular = true;
-            return;
           }
           L_vector[lki_nkj->first] /= U_vector[*uii];
           ++lki_nkj;
           ++uii;
         }
       }
+      // the singularity check inside the for loop won't detect a zero in the bottom right position
+      auto cell_U_bottom_right = std::next(U.AsVector().begin(), i_block * U.FlatBlockSize() + U.FlatBlockSize() - 1);
+      if (*cell_U_bottom_right == 0) is_singular = true;
     }
   }
 
@@ -316,7 +319,6 @@ namespace micm
             if (U_vector[uii_deref + i_cell] == 0.0)
             {
               is_singular = true;
-              return;
             }
             L_vector[lki_nkj_first + i_cell] /= U_vector[uii_deref + i_cell];
           }
@@ -324,6 +326,9 @@ namespace micm
           ++uii;
         }
       }
+
+      auto cell_U_bottom_right = std::next(U.AsVector().begin(), i_group * U_GroupSizeOfFlatBlockSize + U_GroupSizeOfFlatBlockSize - 1);
+      if (*cell_U_bottom_right == 0) is_singular = true;
     }
   }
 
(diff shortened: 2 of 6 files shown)


include/micm/solver/lu_decomposition.inl
```diff
@@ -223,16 +213,19 @@ namespace micm
             L_vector[lki_nkj->first] -= L_vector[lkj_uji->first] * U_vector[lkj_uji->second];
             ++lkj_uji;
           }
+
           if (U_vector[*uii] == 0.0)
           {
             is_singular = true;
-            return;
```
sjsprecious, line 229: Shall we still `return` here if we already know it is singular?

include/micm/solver/lu_decomposition.hpp
```diff
@@ -108,16 +108,6 @@ namespace micm
...
 
-    /// @brief Perform an LU decomposition on a given A matrix
-    /// @param A Sparse matrix to decompose
-    /// @param L The lower triangular matrix created by decomposition
-    /// @param U The upper triangular matrix created by decomposition
-    template<class SparseMatrixPolicy>
-    requires(SparseMatrixConcept<SparseMatrixPolicy>) void Decompose(
-        const SparseMatrixPolicy& A,
-        SparseMatrixPolicy& L,
-        SparseMatrixPolicy& U) const;
```
sjsprecious, line 119: What is it used for before? I don't think we ever use this function?

sjsprecious (changes requested):
Thanks @K20shores for addressing this issue. I guess the CUDA version of LU decomposition (https://github.com/NCAR/micm/blob/main/src/solver/lu_decomposition.cu#L96-L102) needs to be revised accordingly as well? And update the `test_cuda_rosenbrock.cpp`?

include/micm/solver/lu_decomposition.hpp
K20shores: It was a pass through. When we called this, we declared a local boolean variable and then called the actual function with the boolean variable. Since this is on our hot path, I figured it's a slight, and useful, optimization to remove one branch (in the LinearFactor function of rosenbrock) and one stack call by eliminating this pass through

include/micm/solver/lu_decomposition.inl
K20shores: that only makes sense when we want to check for singularity. We haven't passed that information down here so we have to continue and expect the rosenbrock implementation to handle singularity properly

sjsprecious: Ah, that makes sense. Thanks for the clarification.

include/micm/solver/lu_decomposition.hpp
sjsprecious: Thanks for the clarification. That makes sense.

sjsprecious (approved):
Thanks @K20shores for working on this and figuring out this bug. I could confirm that all the tests passed on Derecho's A100 GPU.

I also talked with Allison today, and she told me that we could construct a singular matrix with linear dependence between two rows or two columns (which is the definition of "singularity"). Not sure whether this is easy to do but this may avoid the usage of negative reaction rates.

mattldawson (approved):
looks great!

include/micm/solver/lu_decomposition.inl
```diff
@@ -316,15 +312,25 @@ namespace micm
...
       }
+      std::size_t uii_deref = *uii;
+      if (n_cells != A_GroupVectorSize) {
+        // check the bottom right corner of the matrix
+        for (std::size_t i_cell = 0; i_cell < n_cells; ++i_cell)
+        {
+          if (U_vector[uii_deref + i_cell] == 0.0)
+          {
+            is_singular = true;
+          }
```
boulderdaze, line 330: Does it make sense to break the loop after setting the singularity flag?

```c++
        for (std::size_t i_cell = 0; i_cell < n_cells; ++i_cell)
        {
          if (U_vector[uii_deref + i_cell] == 0.0)
          {
            is_singular = true;
            break;
          }
        }
```

include/micm/solver/rosenbrock.inl
```diff
@@ -240,26 +240,21 @@ namespace micm
   {
     MICM_PROFILE_FUNCTION();
 
-    auto jacobian = state.jacobian_;
     uint64_t n_consecutive = 0;
     singular = false;
     while (true)
     {
+      auto jacobian = state.jacobian_;
```
boulderdaze, line 247: Does `jacobian` need copy at every iteration? Does it mean we did wrong previously?

boulderdaze (commented):
Looks good! Just  a couple questions

include/micm/solver/lu_decomposition.inl
sjsprecious: The `break` is removed here to allow a user bypass the singularity check even if the matrix is singular, just like what the current MICM is doing.

include/micm/solver/rosenbrock.inl
sjsprecious: Yes, the `jacobian` needs to be copied every iteration because `alpha` is changed due to the time-step refinement. The previous implementation was correct because we never triggered this iteration about time-step refinement. Thus it did not matter where the copy happened.

include/micm/solver/lu_decomposition.inl
boulderdaze: hmm I don't get it. The loop just sets `is_singular` true or false. If it sets true at an early iteration, it doesn't matter whether continuing checking and setting it. Can you explain how it is related to allowing users to bypass the check?

include/micm/solver/rosenbrock.inl
boulderdaze: Interesting, what is time step refinement?

include/micm/solver/lu_decomposition.inl
sjsprecious: If you `break` here, the LU decomposition will end up with incompletion. If a user wants to bypass singularity check by setting `parameters_.check_singularity_ = false` 
https://github.com/NCAR/micm/blob/fc66c175a5a7857a7e2c76cf1726d6e55855b8c3/include/micm/solver/rosenbrock.inl#L255
the solver will proceed with wrong LU results and probably blow up later.

include/micm/solver/rosenbrock.inl
sjsprecious: This line: https://github.com/NCAR/micm/blob/fc66c175a5a7857a7e2c76cf1726d6e55855b8c3/include/micm/solver/rosenbrock.inl#L261

include/micm/solver/lu_decomposition.inl
boulderdaze: hmm, is there any chance you might be looking at the different line for singularity checks? It should be true for [the above check](https://github.com/NCAR/micm/blob/572-check-for-singularity-when-the-solver-parameters-flag-is-turned-on/include/micm/solver/lu_decomposition.inl#L309-L316) but doesn't seem like it for this check

include/micm/solver/rosenbrock.inl
boulderdaze: ah, thanks!

include/micm/solver/lu_decomposition.inl
sjsprecious: Oh my bad. I am looking at the wrong section of the code. I think you are right here and sorry for the confusion.

K20shores: This particular one is fine to break at since it will only stop the singularity check for one sub block of the matrix. The whole LU decomposition will continue on though

boulderdaze (approved):

mwaxmonsky (approved):
Looks great @K20shores!

PR #614, opened 2024-08-07
Correct usage of third body species

### LINKED ISSUE #613: Third body species are not properly used
In #551 we added support for species to be labeled as third-body, meaning a species that is non reacting, whose concentration doesn't change, but can still be present in a list of products or reactants.

However, if you set any of these M species in our examples as a third body by setting them to something like

```
        {
            "name": "M",
            "type": "CHEM_SPEC",
            "tracer type": "THIRD_BODY"
        },
```

https://github.com/NCAR/micm/blob/c74099cba7df9d5e12898980b9f1e1d108e27765/examples/configs/TS1/species.json#L1274-L1278

https://github.com/NCAR/micm/blob/c74099cba7df9d5e12898980b9f1e1d108e27765/examples/configs/chapman/species.json#L3-L7

https://github.com/NCAR/micm/blob/c74099cba7df9d5e12898980b9f1e1d108e27765/examples/configs/carbon_bond_5/species.json#L131-L134

https://github.com/NCAR/micm/blob/c74099cba7df9d5e12898980b9f1e1d108e27765/test/unit/unit_configs/small_mechanism/species.json#L7-L11

we end up with test failures

```
94% tests passed, 4 tests failed out of 69

Total Test time (real) =  17.88 sec

The following tests FAILED:
         49 - chapman_integration (Failed)
         66 - carbon_bond_5_example (Subprocess aborted)
         67 - chapman_example (Subprocess aborted)
         69 - ts1_example (Subprocess aborted)
Errors while running CTest
```

Each of them fails with the same message

```
:: ctest -R chapman_integration --verbose
49: [ RUN      ] ChapmanIntegration.CanBuildChapmanSystemUsingConfig
49: unknown file: Failure
49: C++ exception with description "M: Reactant does not exist" thrown in the test body.
```

This is the same failure that we see in [music box](https://github.com/NCAR/music-box/issues/169) when trying to set a third body species.

The reason this happens is that products and reactants are read in as names from their reaction lists and turned into `Species` objects, which look like our regular species.

https://github.com/NCAR/micm/blob/c74099cba7df9d5e12898980b9f1e1d108e27765/include/micm/configure/solver_config.hpp#L417-L453

The problem is that a species being a third body is only known when reading from the species list.

https://github.com/NCAR/micm/blob/c74099cba7df9d5e12898980b9f1e1d108e27765/include/micm/configure/solver_config.hpp#L346-L373

When we indicate that a species is a third body, we set a value called `parameterize_` which holds a lambda function to return the air density, which is the value that needs to be used for the concentration of a third body species when calculating reaction rates.

But, we don't know that `M`, or whatever species is marked as third body, should be a third body species when parsing the reactants and products.

The error occurs when we try to build a solver which has read in a configuration with a third body species present. Specifically it fails during construction of the process set. In the snippet below, `reactant` will eventually be `M`, but because it was read from the reactant or product list, it will never have been set as a parameterized species. And, because `variable_map` contains only the non-parameterized species, we will throw an error on line 148.

https://github.com/NCAR/micm/blob/c74099cba7df9d5e12898980b9f1e1d108e27765/include/micm/process/process_set.hpp#L136-L165

**Acceptance criteria**
- All configurations used in our tests with `M` have `M` set as a third body species
- Configurations can build and run

**Ideas**
- Remove the idea of the third body species entirely, or
- In the [configuration reader](https://github.com/NCAR/micm/blob/main/include/micm/configure/solver_config.hpp#L125), use a map of strings to species and lookup the species to set as a reactant or product when constructing the products or reactants

K20shores:
Fix #613. The full writeup is in the issue, but ensuring that the species object in the list of products and reactants contains information on whether something is a third body by copying the value from the species array during a read of a configuration file allows the process set to properly detect if a species is a third body or not and fixes our usage of third body species

### DIFF
--- a/examples/configs/TS1/species.json
+++ b/examples/configs/TS1/species.json
@@ -1274,6 +1274,7 @@
     {
       "name": "M",
       "type": "CHEM_SPEC",
+      "tracer type": "THIRD_BODY",
       "__description": "third-body species"
     }
 
--- a/examples/configs/carbon_bond_5/species.json
+++ b/examples/configs/carbon_bond_5/species.json
@@ -130,7 +130,8 @@
     },
     {
       "name": "M",
-      "type": "CHEM_SPEC"
+      "type": "CHEM_SPEC",
+      "tracer type": "THIRD_BODY"
     },
     {
       "name": "MEO2",
--- a/examples/configs/chapman/species.json
+++ b/examples/configs/chapman/species.json
@@ -3,7 +3,7 @@
         {
             "name": "M",
             "type": "CHEM_SPEC",
-            "tracer type": "CONSTANT"
+            "tracer type": "THIRD_BODY"
         },
         {
             "name": "O2",
--- a/include/micm/configure/solver_config.hpp
+++ b/include/micm/configure/solver_config.hpp
@@ -122,7 +122,7 @@ namespace micm
     using json = nlohmann::json;
 
    public:
-    std::vector<Species> species_arr_;
+    std::map<std::string, Species> species_;
 
     std::vector<UserDefinedRateConstant> user_defined_rate_arr_;
     std::vector<ArrheniusRateConstant> arrhenius_rate_arr_;
@@ -247,7 +247,7 @@ namespace micm
       }
 
       // Clear vectors and maps
-      species_arr_.clear();
+      species_.clear();
       user_defined_rate_arr_.clear();
       arrhenius_rate_arr_.clear();
       troe_rate_arr_.clear();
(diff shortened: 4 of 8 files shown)


mattldawson (approved):
nice!

boulderdaze (approved):
Looks good! One question here, 
Can `"tracer_type": "CONSTANT"` be always converted to `"tracer type": "THIRD_BODY"`?

K20shores:
> Looks good! One question here, Can `"tracer_type": "CONSTANT"` be always converted to `"tracer type": "THIRD_BODY"`?

I don't know. I don't think we really respect `CONSTANT` right now

carl-drews:
"tracer type": "THIRD_BODY",
causes musica.micm_solve() to crash.
            "tracer type": "CONSTANT",
does not.

PR #616, opened 2024-08-07
correcting solver builder constructor

K20shores:
Correct the constructor of the solver. Apparently this was preventing part of musica from building. I don't understand it

### DIFF
--- a/include/micm/solver/solver.hpp
+++ b/include/micm/solver/solver.hpp
@@ -40,13 +40,13 @@ namespace micm
     Solver& operator=(const Solver&) = delete;
 
     Solver(Solver&& other)
-        : number_of_grid_cells_(other.number_of_grid_cells_),
+        : solver_(std::move(other.solver_)),
+          processes_(std::move(other.processes_)),
+          number_of_grid_cells_(other.number_of_grid_cells_),
           number_of_species_(other.number_of_species_),
           number_of_reactions_(other.number_of_reactions_),
           state_parameters_(other.state_parameters_)
     {
-      std::swap(this->solver_, other.solver_);
-      std::swap(this->processes_, other.processes_);
     }
     Solver& operator=(Solver&& other)
     {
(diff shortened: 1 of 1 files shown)


boulderdaze (approved):

mattldawson (approved):

include/micm/solver/solver.hpp
```diff
@@ -40,13 +40,13 @@ namespace micm
...
-        : number_of_grid_cells_(other.number_of_grid_cells_),
+        : solver_(std::move(other.solver_)),
+          processes_(std::move(other.processes_)),
+          number_of_grid_cells_(other.number_of_grid_cells_),
           number_of_species_(other.number_of_species_),
           number_of_reactions_(other.number_of_reactions_),
           state_parameters_(other.state_parameters_)
     {
-      std::swap(this->solver_, other.solver_);
-      std::swap(this->processes_, other.processes_);
```
sjsprecious, line 49: @K20shores @mattldawson I thought doing `swap` here would make sure when `other` is destroyed by the move constructor, it did not free the space now owned by `this`. Kyle's change basically means this is not necessary?

boulderdaze: This `Solver(&&)` move constructor caused an error in musica, and I'm working on it. Also the move assignment has a bug in it. We might have to add test for this.  

```c++
    Solver& operator=(Solver&& other)
    {
      std::swap(this->solver_, other.solver_);
      number_of_grid_cells_ = other.number_of_grid_cells_;
      number_of_species_ = other.number_of_species_;
      number_of_reactions_ = other.number_of_reactions_;
      state_parameters_ = other.state_parameters_;
      std::swap(this->solver_, other.processes_); <---- this should get corrected
      return *this;
    }
```

boulderdaze: no problem! yeah I don't see the need to have move assignment.

PR #618, opened 2024-08-09
Relax the criteria to pass the GPU test with nvhpc/24.7 on Derecho

### LINKED ISSUE #617: Broken GPU test with newer nvhpc compiler
All the CUDA tests in MICM will fail on Derecho if we use a NVHPC compiler newer than `23.7`.

The output indicates that the difference is round-off level. Thus we may just want to relax the criteria for comparison.

sjsprecious:
This PR fixes the broken GPU unit tests with nvhpc/24.7 on Derecho by using a relative tolerance rather than an exact equal match. This seems to be a compiler issue as nvhpc/23.7 on Derecho can pass the equal check with the same setup.

fix #617

### DIFF
--- a/test/unit/cuda/process/test_cuda_process_set.cpp
+++ b/test/unit/cuda/process/test_cuda_process_set.cpp
@@ -14,6 +14,7 @@
 #include <iostream>
 #include <random>
 #include <vector>
+#include <cmath>
 
 using index_pair = std::pair<std::size_t, std::size_t>;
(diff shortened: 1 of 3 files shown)


K20shores (approved):
Incidentally, this may also fix the tests on my consumer-grade GPU! I will have to test it sometime today to see if that's true

mattldawson (approved):

PR #626, opened 2024-08-21
Update fill function for CUDA matrix

### LINKED ISSUE #624: The Fill function for the CUDA matrix is inefficient
Currently the `Fill` function for the CUDA matrix is implemented by filling a host matrix first and then copying it to the device (https://github.com/NCAR/micm/blob/main/include/micm/cuda/util/cuda_dense_matrix.hpp#L232-L233).

A better implementation should use either `CudaMemset` function or a customized CUDA kernel to do it on the device directly.

**Acceptance Criteria**
- Pass all the CUDA unit tests
- No more data transfer for this function

sjsprecious:
This PR updates the `Fill` function for the CUDA matrix so that it no longer triggers a data transfer between host and device.

@mattldawson @K20shores both CudaDenseMatrix and CudaSparseMatrix are using the same `Fill` function. Is it possible to define the `Fill` function only at one place instead?
 
fix #624

### DIFF
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -67,7 +67,7 @@ jobs:
     strategy:
       matrix:
         compiler:
-          - { cpp: g++-11, c: gcc-11}
+          - { cpp: g++-14, c: gcc-14}
           - { cpp: g++-12, c: gcc-12}
           - { cpp: clang++, c: clang}
         build_type: [Release]
--- a/include/micm/cuda/util/cuda_dense_matrix.hpp
+++ b/include/micm/cuda/util/cuda_dense_matrix.hpp
@@ -229,8 +229,14 @@ namespace micm
     /// @param val Value to set each element to
     void Fill(T val)
     {
-      std::fill(this->data_.begin(), this->data_.end(), val);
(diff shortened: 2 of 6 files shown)


K20shores:
@sjsprecious  [GCC11 was removed from github actions](https://github.com/actions/runner-images/issues/10213). Would you mind removing 11 in the workflow file for mac and adding gcc14 as part of this PR? It'll fix the failing test

K20shores (approved):
If you want to use the same function, you'd need to take the contents out of both matrices and put them in a separate function and then call that function. However, you'd still have two functions called `Fill`, once on each matrix, that just pass through to another function. I think what we have is good enough as is; if it were much larger we'd want a different solution though

sjsprecious:
Thanks @K20shores for your comments. I just updated the CI test on Mac to use gcc14 instead of gcc11.

mattldawson (approved):
looks great!

I'm not sure off-hand of a way to merge the two `Fill` functions in a way that wouldn't make the code more difficult to read/maintain, but maybe you could create a discussion for it if you think there is a good way to do this?

PR #636, opened 2024-08-29
Remove redundant variable and optimize the copy assignment for the CUDA matrix

### LINKED ISSUE #634: Remove unnecessary forcing variables
After revisiting the `solve` function, I think the usage of `forcing` and `initial_forcing` variables here (https://github.com/NCAR/micm/blob/main/include/micm/solver/rosenbrock.inl#L20-L21) is unnecessary and can be replaced by the `K` vector.

**Acceptance Criteria**
- Remove the `forcing` and `initial_forcing` variables if possible.
- Pass all the unit tests on Derecho's CPU and GPU.

### LINKED ISSUE #635: The copy assignment operator for the CUDA matrix is not optimal
According to Matt, the copy assignment operator for the CUDA matrix will always reallocate a new device memory space before copying the data (https://github.com/NCAR/micm/blob/main/include/micm/cuda/util/cuda_dense_matrix.hpp). This is not necessary if the matrix size is the same between left and right hand side.

**Acceptance Criteria**
- Only free and reallocate device memory space when the matrix size is different.
- Pass all the unit tests on Derecho's CPU and GPU.

sjsprecious:
This PR:
- removes the usage of `forcing` variable in the `solve` function
- reallocates the device memory in the copy assignment operator only when the matrix size is different

fix #634 
fix #635

### DIFF
--- a/include/micm/cuda/util/cuda_dense_matrix.hpp
+++ b/include/micm/cuda/util/cuda_dense_matrix.hpp
@@ -129,10 +129,12 @@ namespace micm
     CudaDenseMatrix& operator=(const CudaDenseMatrix& other)
     {
       VectorMatrix<T, L>::operator=(other);
-      if (this->param_.d_data_ != nullptr)
+      if (this->param_.number_of_elements_ != other.param_.number_of_elements_)
+      {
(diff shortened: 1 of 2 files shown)


mattldawson (approved):
nice!

mwaxmonsky (approved):

PR #641, opened 2024-09-04
Add CUDA stream for asynchronous kernel launch

sjsprecious:
This PR:
- adds a singleton class to generate the CUDA streams
- makes all the CUDA kernel launches asynchronously
- modifies the default `main()` function for GTest to clean up the CUDA streams before the program exits

All the unit tests passed on Derecho's A100 GPU with `nvhpc/24.7`.

### DIFF
--- a/include/micm/cuda/util/cuda_util.cuh
+++ b/include/micm/cuda/util/cuda_util.cuh
@@ -5,7 +5,10 @@
 #include <cublas_v2.h>
 #include <cuda_runtime.h>
 
+#include <map>
 #include <string>
+#include <memory>
+#include <mutex>
 
 #define CHECK_CUDA_ERROR(err, msg)   micm::cuda::CheckCudaError(err, __FILE__, __LINE__, msg)
 #define CHECK_CUBLAS_ERROR(err, msg) micm::cuda::CheckCublasError(err, __FILE__, __LINE__, msg)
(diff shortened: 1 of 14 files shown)


include/micm/cuda/util/cuda_util.cuh
```diff
@@ -30,5 +33,54 @@ namespace micm
...
+
+    /// @brief Define the smart pointer type using the functor for the custom deleter
+    using CudaStreamPtr = std::unique_ptr<cudaStream_t, CudaStreamDeleter>;
+
+    /// @brief Singleton class to manage CUDA streams
+    class CudaStreamSingleton
+    {
+    public:
+
+      CudaStreamSingleton() = default;
```
K20shores, line 59: I believe this should be `private` in a singleton implementation

include/micm/cuda/util/cuda_util.cuh
```diff
@@ -30,5 +33,54 @@ namespace micm
...
+      
+      ~CudaStreamSingleton() = default;
```
K20shores, line 61: I think this function should be defined, but it should call the `CleanUp` function below. That way we don't have to remember to call cleanup. In fact, if we never need to manually call `CleanUp`, then we could put the contents of `CleanUp` into the destructor

test/unit/cuda/util/cuda_gtest_main.cpp
```diff
@@ -0,0 +1,9 @@
+#include <gtest/gtest.h>
+#include <micm/cuda/util/cuda_util.cuh>
+
+int main(int argc, char **argv) {
+    ::testing::InitGoogleTest(&argc, argv);
+    int result = RUN_ALL_TESTS();
+    micm::cuda::CudaStreamSingleton::GetInstance().CleanUp(); // Ensure cleanup
```
K20shores, line 7: Is this necessary? I think if you define the destructor as I suggested above that we would no longer need this file because the destructor would always run and cleanup the streams

K20shores (changes requested):

include/micm/cuda/util/cuda_util.cuh
sjsprecious: Originally I implemented it exactly what you suggested, but the `CudaStream` resource was something different. When putting the `cleanup` function in the destructor, the CUDA driver is already turned down before the destructor is called and it will lead to an error saying `driver is shutting down`. That is why I have to call the `cleanup` function separately as shown below to clean up the CUDA resource explicitly and avoid GPU memory leak.

test/unit/cuda/util/cuda_gtest_main.cpp
sjsprecious: See comment above.

K20shores (approved):

include/micm/cuda/util/cuda_util.cuh
sjsprecious: Thanks for catching it. Fixed.

PR #642, opened 2024-09-04
Add const to solver functions

mattldawson:
Adds the `const` qualifier to solver functions that should not be modifying object data. Also removes collection of profiling data during actions to reduce runtime.


sjsprecious (approved):
Looks nice!

PR #649, opened 2024-09-05
Replace json to yaml 619

### LINKED ISSUE #619: Remove nlohman::json in favor of yaml-cpp since it parses both yaml and json
The current MICM configuration parser wraps the `nlohman::json` library. Modify the parser and the CMake build scripts to use `yaml-cpp` library instead, which will allow for mechanisms in JSON or YAML.

**Acceptance Criteria**
- `nlohman::json` library is no longer a dependency of MICM
- The MICM configuration parser uses`yaml-cpp` for parsing configuration files
- Include at least one test that reads a YAML configuration

**Ideas**
- Files that may need editing:
  - `cmake/dependencies.cmake`
  - `CMakeLists.txt`
    - In here we define a variable called `MICM_ENABLE_JSON`, maybe change that to `MICM_ENABLE_CONFIG_READER` or something similar
  - `src/CMakeLists.txt`
  - `include/micm/configure/solver_config.hpp`
  - `test/unit/CMakeLists.txt`
  - Many things in `test/unit/configure`
  - There may be others, but this is likely enough to start
- Possibly create a test in the configuration parser unit tests for a YAML configuration (the translation of an existing JSON configuration should be fairly easy in Python or an online converter https://github.com/NCAR/micm/blob/6b1c58a9be14095e3f3c6df403c91c4e800f23de/test/unit/configure/test_solver_config.cpp#L102-L138

montythind:
Replaces nholman json with Yaml-cpp library

Closes #619


include/micm/configure/solver_config.hpp
```diff
@@ -361,51 +366,66 @@ namespace micm
...
-          }
-        }
+        } 
       }
       species_[name] = species;
+    }        
+
+    // Utility functions to check types and perform conversions
+    bool isBool(const std::string& value) {
+        return (value == "true" || value == "false");
```
montythind, line 410: I am not sure where to put these functions.

include/micm/configure/solver_config.hpp
```diff
@@ -361,51 +366,66 @@ namespace micm
...
-          else if (value.is_boolean())
-          {
-            species.SetProperty<bool>(key, value);
-          }
-          else
-          {
-            throw std::system_error{ make_error_code(MicmConfigErrc::InvalidType), key };
-          }
-        }
+        }
```
boulderdaze, line 403: I am fine with this change. Maybe I'd suggest
```c++
if (key != NAME && key != TYPE)
{       
  std::string stringValue = value.as<std::string>();
  if (!stringValue.empty())
  {
    if (isInt(stringValue)){...}
    else if (isFloat(stringValue)){...}
    else if (isBool(stringValue)) {...}
    else {throw error}
  }
  else if (key == TRACER_TYPE && stringValue == THIRD_BODY) {...}
  else 
  {
    species.SetProperty<std::string>(key, stringValue);
  }
}
```

include/micm/configure/solver_config.hpp
boulderdaze: I think putting them below all the parsing functions would be good for consistency. We use CamelCase for function names (refers to our [clang tidy](https://github.com/NCAR/micm/blob/main/.clang-tidy)). Could you modify them in CamelCase?
eg) `isBool` ->`IsBool()`

boulderdaze (changes requested):
Looks good! I have some comments, hopefully they are minor. The issue description mentioned adding a test using a yaml configuration to verify. Could you add a test for that?

include/micm/configure/solver_config.hpp
boulderdaze: Now I see this again, I'm wondering why `stringValue.empty()` is checked only for `int`, `float`, `bool` type. Maybe it could be 
```C++
if (key != NAME && key != TYPE)
{       
  std::string stringValue = value.as<std::string>();
  if (stringValue.empty()) {throw error}
  
  if (isInt(stringValue)){...}
  else if (isFloat(stringValue)){...}
  else if (isBool(stringValue)) {...}
  else if (key == TRACER_TYPE && stringValue == THIRD_BODY) {...}
  else 
  {
    species.SetProperty<std::string>(key, stringValue);
  }
}
```

include/micm/configure/solver_config.hpp
```diff
@@ -119,7 +119,7 @@ namespace micm
 
   class JsonReaderPolicy
```
K20shores, line 120:
```suggestion
  class ConfigReaderPolicy
```

Maybe we should rename the class to reflect the changes?

include/micm/configure/solver_config.hpp
```diff
@@ -213,25 +213,30 @@ namespace micm
       // Iterate CAMP file list and form CAMP data object arrays
       for (const auto& camp_file : camp_files)
       {
-        json config_subset = json::parse(std::ifstream(camp_file));
+        json config_subset = YAML::LoadFile(camp_file.string());
```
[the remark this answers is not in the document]
K20shores: There's a using on line 122, that makes `json` be the `YAML::Node` type, but perhaps we should rename that to `object_type` or something? YAML-cpp reads both yaml and json

include/micm/configure/solver_config.hpp
```diff
@@ -119,7 +119,7 @@ namespace micm
 
   class JsonReaderPolicy
   {
-    using json = nlohmann::json;
+    using json = YAML::Node;
```
K20shores, line 122:
```suggestion
    using object = YAML::Node;
```

Maybe we should rename this so that it no longer seems specific to json?

include/micm/configure/solver_config.hpp
```diff
@@ -361,51 +366,66 @@ namespace micm
...
+            species.SetProperty<int>(key, value.as<int>());
+          }
+          else if (!stringValue.empty() && isFloat(stringValue)) {
+            species.SetProperty<double>(key, value.as<double>());
+          }
+          else if (!stringValue.empty() && isBool(stringValue))
+          {
+            species.SetProperty<bool>(key, value.as<bool>());
+          }
+          else
```
K20shores, line 392: It seems that YAML-cpp provides functionality to [check that type](https://github.com/jbeder/yaml-cpp/wiki/Tutorial). We should use that here, maybe? That would remove the need for these one-off type casting functions.

```
switch (node.Type()) {
  case YAML::NodeType::Null: // ...
  case YAML::NodeType::Scalar: // ...
  case YAML::NodeType::Sequence: // ...
  case YAML::NodeType::Map: // ...
  case YAML::NodeType::Undefined: 
}
```

K20shores (changes requested):
It's looking good! But, we need a few more minor changes. Also, please convert a few of the configurations to yaml so that we always have a test that is parsing yaml

include/micm/configure/solver_config.hpp
montythind: We cant use this node.Type as the code becomes very nested with lots of Try, catch statement and making it very complicated to read. 
Here is example what the code will look like. I dont think it pretty. 
if (node.IsScalar()) {
    // It's a single value (int, float, string, bool)
    std::cout << "This is a scalar value." << std::endl;

    // Check for specific types using as<>
    try {
        int intValue = node.as<int>();
        std::cout << "This is an integer: " << intValue << std::endl;
    } 
    catch (const YAML::BadConversion& e) {
        // Not an integer, try the next type
        try {
            double doubleValue = node.as<double>();
            std::cout << "This is a double: " << doubleValue << std::endl;
        } 
        catch (const YAML::BadConversion& e) {
            // Not a double, try string
            std::string stringValue = node.as<std::string>();
            std::cout << "This is a string: " << stringValue << std::endl;
        }
    }
} 

Thoughts?

K20shores: Ah, I didn't understand fully that yaml-cpp doesn't actually know the type. What you've done sounds fine and my comments no longer apply

sjsprecious (approved):
LGTM. Congrats on your first PR to MICM, nice job! @montythind

K20shores (approved):

boulderdaze (approved):
Looks good! Thank you for addressing the comments!

mattldawson (approved):
looks great!

PR #651, opened 2024-09-10
Add const qualifiers

mattldawson:
Adds some const qualifiers to indices in solver functions.

### DIFF
--- a/include/micm/process/process_set.hpp
+++ b/include/micm/process/process_set.hpp
@@ -280,13 +280,13 @@ namespace micm
       auto react_id = reactant_ids_.begin();
       auto prod_id = product_ids_.begin();
       auto yield = yields_.begin();
-      std::size_t offset_rc = i_group * rate_constants.GroupSize();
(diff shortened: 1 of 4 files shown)


sjsprecious (commented):
Thanks @mattldawson . It looks good to me. Do we need to add the same qualifier to the CUDA code as well?

K20shores (approved):

mattldawson:
> Thanks @mattldawson . It looks good to me. Do we need to add the same qualifier to the CUDA code as well?

It probably wouldn't hurt, but it seems like these changes don't affect the CUDA build, so maybe that could be a separate PR?

sjsprecious:
> > Thanks @mattldawson . It looks good to me. Do we need to add the same qualifier to the CUDA code as well?
> 
> It probably wouldn't hurt, but it seems like these changes don't affect the CUDA build, so maybe that could be a separate PR?

Sure. Sounds good to me.

sjsprecious (approved):

PR #654, opened 2024-09-12
Move temporary variables to the State class

### LINKED ISSUE #643: Move local temporary variables into State variable
After discussing with Matt, it may be easier and convenient to move all the local variables (https://github.com/NCAR/micm/blob/main/include/micm/solver/rosenbrock.inl#L19-L30) into `state` variable so that they can be allocated/deallocated at each time step integration (saving memory space rather than carrying them around the host model).

**Ideas**
- create an empty base class (not templated) for the temporary variables.
- create the templated derived class for the actual temporary variables of a specific solver.
- the State class adds as a new data member as the base class type.
- In the GetState function of the solver, we set the new data member directly.

**Acceptance Criteria**
- Move all the local variables into `state` variable, and pass the unit tests.

sjsprecious:
This PR moves the temporary variables from the Backward Euler and Rosenbrock solvers into a separate struct in the `State` class. This enables us to optimize the data allocation and deallocation of those variables without changing the `solve` function itself.

Many thanks to @mattldawson who did the majority of the revision.

All the unit tests passed on Derecho's A100 GPU with nvhpc/24.7 compiler. Somehow I have to relax the tolerance for a specific GPU integration test.

fix #643


sjsprecious:
Superseded by #655

PR #663, opened 2024-09-19
correcting indexing for LU decomposition

### LINKED ISSUE #625: Some data in the State variable may be corrupt
As identified by this issue https://github.com/NCAR/MUSICA-Performance-Comparison/issues/62, if we reuse the `State` variable between different iterations of `solve` function, the results are corrupt and we get `fail to converge` error message later.

Kyle commented that this was also the reason why he had to initialize the LU matrix to zero originally (https://github.com/NCAR/micm/issues/587).

This issue indicates that the LU matrix in `State` variable is not overwritten correctly between iterations (or we should not expect it to be overwritten correctly at all?).

K20shores:
Closes #625 

See the issue for more detail, but this fixes a bug where the L or U matrix would sometimes not be set to the value in the jacobian matrix, leading to a drifting value of the LU matrices. This would lead to incorrect results and eventually the solver would stop converging. We noticed this in the performance repository and learned that we couldn't reuse the state matrix.

We started setting the LU matrices to zero back in #410, but later removed in #594 when we thought it was just an issue around a vectorized normalized error function. It wasn't.

### DIFF
--- a/docker/Dockerfile.llvm
+++ b/docker/Dockerfile.llvm
@@ -1,4 +1,5 @@
 FROM fedora:37
+# We need llvm version 15.0.7
 
 RUN dnf -y update \
     && dnf -y install \
@@ -9,7 +10,7 @@ RUN dnf -y update \
         git \
         make \
         zlib-devel \
-        llvm-devel \
+        llvm-devel \ 
         openmpi-devel \
         valgrind \
     && dnf clean all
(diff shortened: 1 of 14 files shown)


sjsprecious:
I think similar changes are needed for the CUDA code at
https://github.com/NCAR/micm/blob/main/src/solver/lu_decomposition.cu#L59-L64
and
https://github.com/NCAR/micm/blob/main/src/solver/lu_decomposition.cu#L82-L86

PR #666, opened 2024-09-20
Set LU matrices to zero when jacobian is a zero element

### LINKED ISSUE #625: Some data in the State variable may be corrupt

K20shores:
Better version of #663

Closes #625 

See the issue for more detail, but this fixes a bug where the L or U matrix would sometimes not be set to the value in the jacobian matrix, leading to a drifting value of the LU matrices. This would lead to incorrect results and eventually the solver would stop converging. We noticed this in the performance repository and learned that we couldn't reuse the state matrix.

We started setting the LU matrices to zero back in #410, but later removed in #594 when we thought it was just an issue around a vectorized normalized error function. It wasn't.

### DIFF
--- a/include/micm/util/vector_matrix.hpp
+++ b/include/micm/util/vector_matrix.hpp
@@ -228,7 +228,7 @@ namespace micm
       return L;
     }
 
-    void print() const {
+    void Print() const {
       for (std::size_t i = 0; i < x_dim_; ++i)
       {
         for (std::size_t j = 0; j < y_dim_; ++j)
--- a/include/micm/util/vector_matrix.hpp
+++ b/include/micm/util/vector_matrix.hpp
@@ -228,6 +228,17 @@ namespace micm
       return L;
     }
 
+    void print() const {
+      for (std::size_t i = 0; i < x_dim_; ++i)
+      {
+        for (std::size_t j = 0; j < y_dim_; ++j)
+        {
+          std::cout << (*this)[i][j] << " ";
+        }
+        std::cout << std::endl;
+      }
+    }
+
     /// @brief Set every matrix element to a given value
     /// @param val Value to set each element to
     void Fill(T val)
--- a/src/solver/lu_decomposition.cu
+++ b/src/solver/lu_decomposition.cu
@@ -56,23 +56,21 @@ namespace micm
           auto inLU = d_niLU[i];
           for (size_t iU = 0; iU < inLU.second; ++iU)
           {
+            size_t U_idx = d_uik_nkj[uik_nkj_offset].first + tid;
             if (d_do_aik[do_aik_offset++])
             {
-              size_t U_idx = d_uik_nkj[uik_nkj_offset].first + tid;
               size_t A_idx = d_aik[aik_offset++] + tid;
               d_U[U_idx] = d_A[A_idx];
             }
             else {
-              size_t U_idx = d_uik_nkj[uik_nkj_offset].first + tid;
               d_U[U_idx] = 0;
             }
 
             for (size_t ikj = 0; ikj < d_uik_nkj[uik_nkj_offset].second; ++ikj)
             {
-              size_t U_idx_1 = d_uik_nkj[uik_nkj_offset].first + tid;
               size_t L_idx = d_lij_ujk[lij_ujk_offset].first + tid;
               size_t U_idx_2 = d_lij_ujk[lij_ujk_offset].second + tid;
-              d_U[U_idx_1] -= d_L[L_idx] * d_U[U_idx_2];
+              d_U[U_idx] -= d_L[L_idx] * d_U[U_idx_2];
               ++lij_ujk_offset;
             }
             ++uik_nkj_offset;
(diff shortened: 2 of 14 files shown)


test/unit/solver/test_linear_solver_policy.hpp
```diff
@@ -204,6 +204,64 @@ void testRandomMatrix(std::size_t number_of_blocks)
       A, b, x, [&](const FloatingPointType a, const FloatingPointType b) -> void { EXPECT_NEAR(a, b, 1.0e-5); });
```
sjsprecious, line 204: can we use a smaller tolerance here now?

test/unit/solver/test_linear_solver_policy.hpp
sjsprecious, line 1: Also can we initialize the LU matrix with 0 rather than 1.e-30 now?

sjsprecious (commented):
@K20shores Thanks for fixing this bug. It will make me sleep much better for the weekend.

Just a few minor comments but otherwise the PR looks good to me. Great job!

src/solver/lu_decomposition.cu
```diff
@@ -62,6 +62,10 @@ namespace micm
               size_t A_idx = d_aik[aik_offset++] + tid;
               d_U[U_idx] = d_A[A_idx];
             }
+            else {
+              size_t U_idx = d_uik_nkj[uik_nkj_offset].first + tid;
```
boulderdaze, line 66: `Line 61` has the same definition. Maybe we could define it outside if conditional?

K20shores: I was noticing that too. There's a lot of redefinition in these and I was wondering if now would be a good time to remove those

boulderdaze: Yeah I think we can address them in this PR.

boulderdaze (changes requested):
Looks good! I only have a minor change request. I think the test this PR brings in are very valuable.

test/unit/solver/test_linear_solver_policy.hpp
K20shores: yes

include/micm/util/vector_matrix.hpp
```diff
@@ -228,6 +228,17 @@ namespace micm
...
+      for (std::size_t i = 0; i < x_dim_; ++i)
+      {
+        for (std::size_t j = 0; j < y_dim_; ++j)
+        {
+          std::cout << (*this)[i][j] << " ";
+        }
+        std::cout << std::endl;
+      }
+    }
+
```
boulderdaze, line 241: Doesn't look like it is used anywhere, could you confirm it? Since we don't execute clang tidy automatically, could you name the function in CamelCase?

sjsprecious (approved):

src/solver/lu_decomposition.cu
```diff
@@ -85,6 +87,11 @@ namespace micm
               size_t A_idx = d_aki[aki_offset++] + tid;
               d_L[L_idx] = d_A[A_idx];
             }
+            else {
+              size_t L_idx = d_lki_nkj[lki_nkj_offset].first + tid;
+              d_L[L_idx] = 0;
```
boulderdaze, line 92: The same line is found in `line 86`. I think we can define it outside the conditional

src/solver/lu_decomposition.cu
```diff
@@ -85,6 +87,11 @@ namespace micm
...
+            }
+
             for (size_t ikj = 0; ikj < d_lki_nkj[lki_nkj_offset].second; ++ikj)
             {
               size_t L_idx_1 = d_lki_nkj[lki_nkj_offset].first + tid;
```
boulderdaze, line 97: I think this line can be deleted as well

include/micm/util/vector_matrix.hpp
K20shores: It was useful when I was debugging. I'd like to keep it

src/solver/lu_decomposition.cu
K20shores: fixed

boulderdaze: oh I was suggesting for line `94-101`
```C++
  for (size_t ikj = 0; ikj < d_lki_nkj[lki_nkj_offset].second; ++ikj)
  {
    size_t L_idx_1 = d_lki_nkj[lki_nkj_offset].first + tid;   // Delete this 
    ...
    d_L[L_idx] -= d_L[L_idx_2] * d_U[U_idx];    // Update d_L[L_idx_1] to d_L[L_idx] 
    ++lkj_uji_offset;
  }
```

include/micm/util/vector_matrix.hpp
boulderdaze: Yeah that's fine with me. Would you be able to make it in CamelCase though for the consistency in our code base?

boulderdaze (approved):
Thank you for addressing the comments

PR #671, opened 2024-09-27
Remove singular check for LU decomposition and add const qualifier to the CUDA kernels

### LINKED ISSUE #637: Explore whether using pointers to const data as arg to CUDA kernels improves performance
Currently, CUDA kernels get passed `const` structs like [`LuDecomposeParam`](https://github.com/NCAR/micm/blob/cb4d6eb5dd33583c061076b57009da944dfd9119/include/micm/cuda/util/cuda_param.hpp#L34-L57) which include pointers to arrays of indices. However, the array element types are not `const`. Explore whether creating something like a `ConstLuDecomposeParam` struct that has data member pointers to `const` data, and then having this type be used as the argument to CUDA kernels (with appropriate assignment operator overloads) will result in better pre-fetching of data in memory during the execution of the kernel.

Possible definition of new struct:
```
struct ConstLuDecomposeParam
{
  const std::pair<const std::size_t, const std::size_t>* niLU_;
  const char* const do_aik_;
  const std::size_t* const aik_;
  const std::pair<const std::size_t, const std::size_t>* uik_nkj_;
  const std::pair<const std::size_t, const std::size_t>* lij_ujk_;
  const char* do_aki_;
  const std::size_t* aki_;
  const std::pair<const std::size_t, const std::size_t>* lki_nkj_;
  const std::pair<const std::size_t, const std::size_t>* lkj_uji_;
  const std::size_t* uii_;
  std::size_t niLU_size_;
  std::size_t do_aik_size_;
  std::size_t aik_size_;
  std::size_t uik_nkj_size_;
  std::size_t lij_ujk_size_;
  std::size_t do_aki_size_;
  std::size_t aki_size_;
  std::size_t lki_nkj_size_;
  std::size_t lkj_uji_size_;
  std::size_t uii_size_;

  ConstLuDecomposeParam(const LuDecomposeParam& other) :
    niLU_(other.niLU_),
    ...
    {}
};
```
or we can change the existing `LuDecomposeParam` struct with the new one above (without the initializer) and see if it works too.

### LINKED ISSUE #664: Remove the singular check for the Jacobian matrix
Current in MICM, we will check whether the Jacobian matrix is singular or not after `LuDecomposition` (https://github.com/NCAR/micm/blob/main/include/micm/solver/rosenbrock.inl#L251-L252).

Based on my experience, CAM-Chem never did this in practical runs and this check can be costly for GPU runs because a synchronization is needed in order to return the correct value on the GPU (which is not done correctly right now).

Thus it may be easier to remove all the singular check in the MICM CPU & GPU code?

sjsprecious:
This PR:
- removes the singular check for LU decomposition (previously the GPU code is not checking the singularity correctly).
- adds the `const` qualifier to the CUDA kernels (it reduces the time of 64 tests (including all CUDA tests) to 245 seconds now!)

fix #664 
fix #637


mattldawson (approved):
nice!

PR #673, opened 2024-10-02
Copyable state

K20shores:
To fix [255](https://github.com/NCAR/musica/pull/225) in musica, the state needs to be moveable so that store the state properly copies the temporary variables

### DIFF
--- a/include/micm/solver/state.hpp
+++ b/include/micm/solver/state.hpp
@@ -59,6 +59,10 @@ namespace micm
     std::size_t number_of_grid_cells_;
     std::unique_ptr<TemporaryVariables> temporary_variables_;
 
+    /// @brief Constructor with parameters
+    /// @param parameters State dimension information
+    State(const StateParameters& parameters);
+
     /// @brief Copy constructor
     /// @param other The state object to be copied
     State(const State& other)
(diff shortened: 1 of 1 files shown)


mattldawson (approved):

include/micm/solver/state.hpp
```diff
@@ -59,6 +59,10 @@ namespace micm
     std::size_t number_of_grid_cells_;
     std::unique_ptr<TemporaryVariables> temporary_variables_;
 
+    /// @brief Constructor with parameters
+    /// @param parameters State dimension information
+    State(const StateParameters& parameters);
+
```
boulderdaze, line 65: Can't find the definition of this constructor. Is it a placeholder?

include/micm/solver/state.hpp
```diff
@@ -139,6 +182,10 @@ namespace micm
     /// @brief Print state (concentrations) at the given time
     /// @param time solving time
     void PrintState(double time);
+
+    /// @brief Default constructor
+    /// Only defined to be used to create default values in types, but a default constructed state is not useable
+    State();
```
boulderdaze, line 188: Would it make more sense to put them in where the `State(...)` constructors are defined?

boulderdaze (changes requested):
Looks good! Just have a couple minor questions

include/micm/solver/state.hpp
K20shores: Sure, I can move it back. I had made it private briefly which is why it's here. But that can't work becuase musica does need to make a default value for it to make the MICM object there

include/micm/solver/state.hpp
K20shores: it's in the inl file.

boulderdaze: ahh my bad

boulderdaze (approved):

sjsprecious (approved):

PR #679, opened 2024-10-18
Set hstart to timestep

K20shores:
To make it easy to test backward euler in the performance repository, I've added `h_start` to the backward euler parameters. This means we can use the same testing code for rosenbrock or backward euler


sjsprecious (approved):
Thanks for working on this change. Looks great to me!

mwaxmonsky (approved):

boulderdaze (approved):

PR #683, opened 2024-11-08
Change LU decomposition class to be Doolittle-specific

### LINKED ISSUE #682: Rename LU decomposition
In preparation for bringing in the MOZART algorithm for LU decomposition, rename the existing LU decomposition classes to include Doolittle in the name (which is the name of the currently used algorithm).

**Acceptance Criteria**
- C++, CUDA, and JIT LU decomposition classes are renamed to include the name of the algorithm (Doolittle)
- All current tests pass
- Documentation referring to the current classes are updated

**Ideas**
- Could use class names `DoolittleLuDecomposition`, `CudaDoolittleLuDecomposition`, and `JitDoolittleLuDecomposition`

sjsprecious:
This PR renames the LU decomposition class to `Doolittle`, which is the specific implementation for LU decomposition in MICM.

fix #682


include/micm/jit/solver/jit_solver_builder.hpp
```diff
@@ -17,7 +17,7 @@ namespace micm
   ///
   /// JIT-compiled solvers only work with vector-ordered matrices
   template<class SolverParametersPolicy, std::size_t L = MICM_DEFAULT_VECTOR_SIZE>
-  using JitSolverBuilder = SolverBuilder<
+  using JitSolverBuilder_Doolittle = SolverBuilder<
```
K20shores, line 20:
```suggestion
  using DefaultJITSolverBuilder = SolverBuilder<
```

I prefer this instead of adding doolittle to the name. What do you think?

include/micm/cuda/solver/cuda_solver_builder.hpp
```diff
@@ -21,7 +21,7 @@ namespace micm
   ///
   /// GPU solvers only work with vector-ordered matrices
   template<class SolverParametersPolicy, std::size_t L = MICM_DEFAULT_VECTOR_SIZE>
-  using CudaSolverBuilder = SolverBuilder<
+  using CudaSolverBuilder_Doolittle = SolverBuilder<
```
K20shores, line 24:
```suggestion
  using DefaultCUDASolverBuilder = SolverBuilder<
```
I prefer this instead of adding doolittle to the name. What do you think?

examples/example.cpp
```diff
@@ -47,7 +47,7 @@ int main(const int argc, const char* argv[])
   auto chemical_system = solver_params.system_;
   auto reactions = solver_params.processes_;
 
-  auto solver = CpuSolverBuilder<micm::RosenbrockSolverParameters>(solver_params.parameters_)
+  auto solver = CpuSolverBuilder_DoolittleLU<micm::RosenbrockSolverParameters>(solver_params.parameters_)
```
K20shores, line 50:
```suggestion
  auto solver = DefaultCPUSolverBuilder<micm::RosenbrockSolverParameters>(solver_params.parameters_)
```
I prefer this instead of adding doolittle to the name. What do you think?

K20shores (approved):
I have a preference for only adding doolittle to the LU decomposition class and either leaving the solver builder names alone or renaming them to `Default*`. But I can be convinced either way

include/micm/jit/solver/jit_solver_builder.hpp
sjsprecious: Thanks Kyle. I do not like my current name change to the solver builder either. If I add the `Default` prefix to the solver builder as you suggested, what would be the alternative name when we add the new LU decomposition implementation?

K20shores: If it's faster, maybe `_Fast`? That'll at least indicate that something fancy is going on

sjsprecious: That sounds great. For consistency purpose, is it better to put both `Default` and `Fast` as prefix or suffix?

examples/example.cpp
mattldawson: agreed, or just leave it as `CpuSolverBuilder`. If the MOZART LU algorithm is faster, we'll probably just use that in the examples and tutorials

examples/profile_example.cpp
```diff
@@ -34,7 +34,7 @@ namespace fs = std::filesystem;
 using namespace micm;
 
 template<std::size_t L>
-using VectorBuilder = CpuSolverBuilder<RosenbrockSolverParameters, VectorMatrix<double, L>, SparseMatrix<double, SparseMatrixVectorOrdering<L>>>;
+using VectorBuilder = CpuSolverBuilder_DoolittleLU<RosenbrockSolverParameters, VectorMatrix<double, L>, SparseMatrix<double, SparseMatrixVectorOrdering<L>>>;
```
mattldawson, line 37: same as above

include/micm/jit/solver/jit_solver_builder.hpp
mattldawson: Personally, I would just leave this as `JitSolverBuilder`. If the MOZART algorithm is faster, but someone wants to use the Doolittle algorithm (for some reason), they can just specify all the template parameters for the solver builder themselves

mattldawson (changes requested):
looks good! Just a couple comments related to the alias names

include/micm/jit/solver/jit_solver_builder.hpp
sjsprecious: Thanks @mattldawson . I just revert to the original builder name and we can do what you suggest after we implement the MOZART algorithm.

examples/example.cpp
sjsprecious: As Matt suggested, we may just keep the original name for now.

sjsprecious:
Thanks @mattldawson . I just revert the builder to use its original name as suggested.

mattldawson (approved):
nice!

PR #691, opened 2024-12-05
Fixes a bug in the logic used to calculate alpha - jacobian

mattldawson:
Inside the convergence loop of the Rosenbrock solver there may be repeated calls to `AbstractRosenbrockSolver::LinearFactor()`. Because this function overwrites the negative jacobian stored in `state.jacobian_`, the alpha values from successive calls accumulate along the diagonal.

This apparently didn't have enough of an impact to affect the results of the analytical tests, but should be corrected.

I wasn't sure of the best way to test this fix, because the impact was small and we don't currently have unit tests for the `AbstractRosenbrockSolver::Solve()` function because exact outputs are difficult to calculate. If anyone has suggestions for tests, I can add them.

### DIFF
--- a/include/micm/solver/rosenbrock.inl
+++ b/include/micm/solver/rosenbrock.inl
@@ -66,11 +66,17 @@ namespace micm
       stats.jacobian_updates_ += 1;
 
       bool accepted = false;
+      double last_alpha = 0.0;
       //  Repeat step calculation until current step accepted
       while (!accepted)
       {
+        // Compute alpha accounting for the last alpha value
+        // This is necessary to avoid the need to re-factor the jacobian
+        double alpha = 1.0 / (H * parameters_.gamma_[0]) - last_alpha;
+
         // Form and factor the rosenbrock ode jacobian
-        LinearFactor(H, parameters_.gamma_[0], Y, stats, state);
+        LinearFactor(alpha, Y, stats, state);
+        last_alpha = alpha;
 
         // Compute the stages
         for (uint64_t stage = 0; stage < parameters_.stages_; ++stage)
@@ -222,15 +228,13 @@ namespace micm
 
   template<class RatesPolicy, class LinearSolverPolicy, class Derived>
   inline void AbstractRosenbrockSolver<RatesPolicy, LinearSolverPolicy, Derived>::LinearFactor(
-      double& H,
-      const double gamma,
+      const double alpha,
       const auto& number_densities,
       SolverStats& stats,
       auto& state) const
   {
     MICM_PROFILE_FUNCTION();
 
-    double alpha = 1 / (H * gamma);
     static_cast<const Derived*>(this)->AlphaMinusJacobian(state.jacobian_, alpha);
 
     linear_solver_.Factor(state.jacobian_, state.lower_matrix_, state.upper_matrix_);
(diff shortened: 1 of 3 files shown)


sjsprecious (commented):
Thanks @mattldawson for catching this issue. It looks good to me.

Can you lower the tolerance in the integration tests to see if this change makes a difference or not?

include/micm/solver/rosenbrock.inl
```diff
@@ -66,11 +66,17 @@ namespace micm
...
 
       bool accepted = false;
+      double last_alpha = 0.0;
       //  Repeat step calculation until current step accepted
       while (!accepted)
       {
+        // Compute alpha accounting for the last alpha value
+        // This is necessary to avoid the need to re-factor the jacobian
+        double alpha = 1.0 / (H * parameters_.gamma_[0]) - last_alpha;
+
```
K20shores, line 76: I'm not sure this is correct. We used to make a copy of the jacobian and then do a linear factor and continue, so this accumulation never would have happened. I don't think we are meant to be reusing the last alpha value

K20shores: The copy was removed in #646 and then the iteration inside of the linear factor was removed with #671

sjsprecious: Making a copy of Jacobian matrix will make this issue go away but it will take extra memory space and copy time. Doing the change inline is more efficient but we just need to handle the previous last alpha value correctly in this case.

boulderdaze (approved):

include/micm/solver/rosenbrock.inl
K20shores: I guess what I don't understand is why we need the last alpha value. I thought that the value of alpha was meant to be entirely dependent on the current value of `H` and not any previous values.

sjsprecious: Hmm, let me try this way. What `AlphaMinusJacobian` function does is to add the `alpha` value to the diagonal elements of Jacobian matrix (in the current implementation, the Jacobian matrix is modified inline), and the value of `alpha` depends on the current `H` value. If a time-substepping is required, we need to recalculate the `H`, and then `alpha`, and then apply the new `alpha` to the **original** Jacobian matrix. But since the original Jacobian matrix has been modified by the previous alpha value, we then need to subtract the previous alpha value from the diagonal elements to restore the original Jacobian matrix, and then apply the new `alpha` value.

sjsprecious: If we make a copy of Jacobian matrix, we do not need to worry about the previous `alpha` value because we can always access the **original** Jacobian matrix directly. But since the current implementation changes the Jacobian matrix by the `AlphaMinusJacobian` function, we need to restore the original Jacobian matrix first when doing the time-substepping.

K20shores: Ohh, I get it now. The subtraction cancels out the first alpha.

sjsprecious: Exactly! This seems a bug to me but I am surprised that it does not affect the integration tests a lot (I tried to lower the tolerance but it did not succeed). I expect that this should make a difference for some stiff problems that require time-substeppings.

sjsprecious (approved):
LGTM!

PR #698, opened 2024-12-24
Add in-place LU decomposition option for linear solve

### LINKED ISSUE #695: Add an in-place LU decomposition using the MOZART LU algorithm
The current implementation of the MOZART LU decomposition scheme accepts separate A, L, and U matrices. Add an LU decomposition option that can be used with the Backward Euler solver that does an in-place LU decomposition, overwriting the A matrix with the L and U matrix data, and then uses this in the linear solve.

**Acceptance Criteria**
- An in-place LU decomposition and solve is available for use with the Backward Euler solver
- Tests of the Backward Euler solver with the in-place LU decomposition and solve are included for all the analytical test cases.

**Ideas**
- Collect the Jacobian non-zero elements, do the Diagonal Markowitz reordering, add any additional non-zero elements needed by the L and U matrices, and then create a jacobian/L/U combined sparse matrix in the `State`
- Overload the `LinearSolver::Factor()` and `LinearSolver::Solve()` functions with versions that accept a non-const reference to a single sparse matrix.

mattldawson:
Adds an option for doing in-place LU decomposition that overwrites the original A matrix with the combined L and U matrices. Adds a test using in-place LU decomposition with the Backward Euler solver.

closes #695

### DIFF
--- a/include/micm/solver/backward_euler.inl
+++ b/include/micm/solver/backward_euler.inl
@@ -54,7 +54,8 @@ namespace micm
     // if the last attempt to reduce the timestep fails,
     // accept the current H but do not update the Yn vector
 
-    using MatrixPolicy = decltype(state.variables_);
+    using DenseMatrixPolicy = decltype(state.variables_);
+    using SparseMatrixPolicy = decltype(state.jacobian_);
 
     SolverResult result;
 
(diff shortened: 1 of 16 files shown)


sjsprecious (approved):
Thanks @mattldawson for implementing this. Looks good to me.

PR #700, opened 2024-12-27
Filter zeros in solver results

mattldawson:
Adds a filter to the Solver class's Solve function to remove negative results. Also corrects logic in tutorial code for accumulating solver timing output.


sjsprecious (approved):
Looks nice!

PR #713, opened 2025-01-24
Change the matrix digit format to 6 decimal points and 12 width

sjsprecious:
Pure formatting change to make the matrix print output more readable


K20shores (approved):

sjsprecious:
Looks like the format does not work on different platforms and compilers. So just close it without merge.

PR #717, opened 2025-01-29
Fix substepping issue for Rosenbrock solver with inline LU decomposition

### LINKED ISSUE #704: Need to update the workflow for the time substepping of Rosenbrock solver with inline LU algorithm
Since the inline [LU implementation](https://github.com/NCAR/micm/pull/698) is merged, we should be able to use it for the Rosenbrock solver as well.

However, if the inline LU algorithm is used, we can no longer do the trick for the [AlphaMinusJacobian function](https://github.com/NCAR/micm/blob/main/include/micm/solver/rosenbrock.inl#L73-L79) like it currently does because the Jacobian matrix is modified directly. Instead, we need to re-calculate the Jacobian matrix if the time substepping is triggered.

**Acceptance criteria**
- Clean up the settings related to the `last_alpha` variable.
- All the unit tests should pass on CPU and GPU.

**Ideas**

- Move the `SubtractJacobianTerms` function inside the [while loop](https://github.com/NCAR/micm/blob/main/include/micm/solver/rosenbrock.inl#L71) and remove the `last_alpha` variable there.

sjsprecious:
This PR fixes a potential issue of using Rosenbrock solver with inline LU decomposition algorithm, when a substepping happens.

Basically we need to:
- re-compute the Jacobian matrix after each sub step because the Jacobian matrix is now modified by the LU decomposition algorithm in place
- only keep the `last_alpha` code for the non-inline LU decomposition algorithm.

fix #704

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -82,7 +82,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Run CMake
-        run: cmake -S . -B build -G "Visual Studio 16 2019" -A ${{ matrix.architecture }} -T ClangCL
+        run: cmake -S . -B build -D CMAKE_CXX_COMPILER="C:/Program Files/LLVM/bin/clang-cl.exe" -G "Visual Studio 16 2019" -A ${{ matrix.architecture }} -T ClangCL
       - name: Build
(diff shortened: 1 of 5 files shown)


K20shores (approved):

mattldawson (approved):
nice!

PR #724, opened 2025-02-10
moving diagonal elements to state

### LINKED ISSUE #681: Jacobian diagonal elements should be moved to the state
During #668, I noticed that the jacobian diagonal elements are still on the solver and so we are not yet agnostic to the mechanism solved

https://github.com/NCAR/micm/blob/3f0c6077c54eacdf3836595b5128f57f42e8462b/include/micm/solver/rosenbrock.hpp#L57

# Acceptance criteria
- jacobian diagonal elements are removed from the solver

K20shores:
Closes #681

With this, I believe all mechanism specific information is removed from the solvers, at least for the CPU.

Lots of the information copied at construction of the GPU is still mechanism specific though


sjsprecious (approved):
Thanks @K20shores for working on this PR. It looks good to me.

I could confirm that all the tests pass on Derecho's A100 GPU using the `nvhpc/24.11` compiler.

mattldawson (approved):

PR #734, opened 2025-02-21
adding state default constructor test

K20shores:
Adds a test of the state's default constructor, fixes a memory error


sjsprecious (approved):
Thanks @K20shores for identifying and fixing this issue.

PR #737, opened 2025-03-11
correcting more state constructors, adding tests

K20shores:
corrected the move assignment constructor, added test for all copy/move/assignment operations on the state class


sjsprecious (approved):
Thanks @K20shores for fixing this issue and adding the unit test. Looks good to me!

PR #748, opened 2025-03-28
adding a swap method to save on memory copies

K20shores:
Just a small idea I had. Rosenbrock can elide some memory copies if we just swap the pointers of the vector data when we need to copy some memory. Feel free to reject this if we don't think it's useful

### DIFF
--- a/include/micm/solver/rosenbrock.inl
+++ b/include/micm/solver/rosenbrock.inl
@@ -62,6 +62,12 @@ namespace micm
       rates_.AddForcingTerms(state.rate_constants_, Y, initial_forcing);
       stats.function_calls_ += 1;
 
+      // For memory optimization, the initial forcing can be reused for the first stage.
+      // However, we must always copy the initial forcing to the first stage because K[0] 
+      // and the initial forcing are frequently swapped. This ensures the initial forcing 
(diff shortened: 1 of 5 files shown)


include/micm/solver/rosenbrock.inl
```diff
@@ -91,7 +97,7 @@ namespace micm
           double stage_combinations = ((stage + 1) - 1) * ((stage + 1) - 2) / 2;
           if (stage == 0)
           {
-            K[stage].Copy(initial_forcing);
+            K[stage].Swap(initial_forcing);
```
sjsprecious, line 100: If the time step is rejected twice, I think the `K[0]` vector is not initialized to `initial_forcing` any more?

K20shores: I think you're right. I didn't think far enough ahead, I suppose

sjsprecious (approved):
Thanks @K20shores for working on this improvement. I could not think about a exception case that may fail the `swap` change here but @mattldawson or @boulderdaze may be able to. :)

sjsprecious:
I could also confirm that all the CUDA tests pass on Casper's A100 GPU using gcc/12.4.0.

PR #751, opened 2025-04-03
Update CUDA LU decomposition kernel with the CPU optimizations

### LINKED ISSUE #743: Update the Cuda implementation of LU decomposition
Matt implemented the in-place version of LU decomposition for the CPU code. We need to update the CUDA implementation accordingly.

sjsprecious:
This PR updates the CUDA LU decomposition kernel with the in-place Mozart algorithm, which is found an optimal implementation based on the CPU performance evaluation.

I overwrite the original Doolittle LU decomposition CUDA kernel as I think there is no value to keep it. I can bring it back if anyone is still interested in it somehow.

Since the linear solver CUDA kernel is not updated with the in-place algorithm yet, the CUDA integration test is broken at this moment. I will fix it once the linear solver CUDA kernel is udpated.

fix #743


include/micm/solver/lu_decomposition_mozart_in_place.hpp
```diff
@@ -90,7 +90,7 @@ namespace micm
       requires(VectorizableSparse<SparseMatrixPolicy>)
     void Decompose(SparseMatrixPolicy& ALU) const;
 
-   private:
+   protected:
```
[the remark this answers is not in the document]
sjsprecious: Yes I need this change in order to expose the `initialize` function from the CUDA LU decomposition class.

test/unit/solver/test_lu_decomposition_in_place_policy.hpp
```diff
@@ -198,7 +218,7 @@ void testExtremeValueInitialization(std::size_t number_of_blocks, double initial
   CopyToHost<SparseMatrixPolicy>(ALU);
 
   check_results<double, SparseMatrixPolicy>(
-      A, ALU, [&](const double a, const double b) -> void { EXPECT_NEAR(a, b, 1.0e-09); });
+      A, ALU, [&](const double a, const double b) -> void { EXPECT_NEAR(a, b, 1.0e-14); });
```
K20shores, line 221: love this

K20shores (approved):

test/unit/solver/test_lu_decomposition_in_place_policy.hpp
sjsprecious: Looks like Windows does not like this. Revert it back.

mattldawson (approved):
nice!

PR #753, opened 2025-04-04
Update CUDA linear solver with in-place algorithm

### LINKED ISSUE #744: Update the Cuda implementation of linear solver
Matt has implemented the in-place version of linear solver and we need to update the CUDA implementation accordingly.

sjsprecious:
This PR updates the CUDA linear solver kernel with the in-place algorithm, same as the CPU implementation.

It also fixes the broken CUDA integration and linear solver tests.

I confirm that all the CUDA tests pass on Derecho's A100 GPU with `gcc/12.4.0`.

fix #744


mattldawson (approved):
nice!

sjsprecious:
@mattldawson @K20shores why Copilot can not review `*hpp` and `*cpp` files as well? Does it simply hate my PR? :(

K20shores (approved):

PR #755, opened 2025-04-10
Catch-all header

K20shores:
this will allow us to use singular #includes to get any functionality of micm that we wish

Feel free to reject this if it seems unuseful

### DIFF
--- a/include/micm/CPU.hpp
+++ b/include/micm/CPU.hpp
@@ -1,0 +1,10 @@
+// Copyright (C) 2024-2025 University Corporation for Atmospheric Research
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <micm/Process.hpp>
+#include <micm/Solver.hpp>
+#include <micm/System.hpp>
+#include <micm/Util.hpp>
+#include <micm/version.hpp>
--- a/include/micm/GPU.hpp
+++ b/include/micm/GPU.hpp
@@ -1,0 +1,24 @@
+// Copyright (C) 2024-2025 University Corporation for Atmospheric Research
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <micm/cuda/process/cuda_process_set.hpp>
+#include <micm/cuda/process/cuda_process_set.cuh>
+
+#include <micm/cuda/solver/cuda_linear_solver_in_place.cuh>
+#include <micm/cuda/solver/cuda_linear_solver_in_place.hpp>
+#include <micm/cuda/solver/cuda_lu_decomposition.hpp>
+#include <micm/cuda/solver/cuda_lu_decomposition_mozart_in_place.cuh>
+#include <micm/cuda/solver/cuda_lu_decomposition_mozart_in_place.hpp>
+#include <micm/cuda/solver/cuda_rosenbrock.cuh>
+#include <micm/cuda/solver/cuda_rosenbrock.hpp>
+#include <micm/cuda/solver/cuda_solver_builder.hpp>
+#include <micm/cuda/solver/cuda_solver_parameters.hpp>
+#include <micm/cuda/solver/cuda_state.hpp>
+
+#include <micm/cuda/util/cuda_dense_matrix.hpp>
+#include <micm/cuda/util/cuda_matrix.cuh>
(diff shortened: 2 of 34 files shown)


mattldawson (approved):
nice!

boulderdaze (commented):
It reminds me of precompiled headers but it doesn't function like that, right? Is the purpose to have this collection of header files specifically for use in tests? If so, should they instead be relocated to the test directory?

K20shores:
> It reminds me of precompiled headers but it doesn't function like that, right? Is the purpose to have this collection of header files specifically for use in tests? If so, should they instead be relocated to the test directory?

It's so that anyone could `#include <micm/CPU.hpp>` and have everything they need without having to figure out which header files to load. This isn't intended to support precompiled headers, but maybe it would allow us to? Although, if we want that, we may consider moving straight to [c++ modules](https://en.cppreference.com/w/cpp/language/modules).

Also, the next bit of work I need to do is to setup musica to include the cuda solver. I think it's time that we make some better aliases so that the musica code can be smaller. This will allow me to put all of the aliases in one place. It may not be strictly necessary, but I think it's nice

boulderdaze:
> > It reminds me of precompiled headers but it doesn't function like that, right? Is the purpose to have this collection of header files specifically for use in tests? If so, should they instead be relocated to the test directory?
> 
> It's so that anyone could `#include <micm/CPU.hpp>` and have everything they need without having to figure out which header files to load. This isn't intended to support precompiled headers, but maybe it would allow us to? Although, if we want that, we may consider moving straight to [c++ modules](https://en.cppreference.com/w/cpp/language/modules).
> 
> Also, the next bit of work I need to do is to setup musica to include the cuda solver. I think it's time that we make some better aliases so that the musica code can be smaller. This will allow me to put all of the aliases in one place. It may not be strictly necessary, but I think it's nice

I see. My concern with this approach is that users might end up including files that they don't need, which could increase memory usage. I'm not quite sure how this is related to making the musica code smaller. I haven't really seen this practice commonly used in the other codebase, but if you think it's nice to have, I understand.

boulderdaze (approved):

K20shores:
> > > It reminds me of precompiled headers but it doesn't function like that, right? Is the purpose to have this collection of header files specifically for use in tests? If so, should they instead be relocated to the test directory?
> > 
> > 
> > It's so that anyone could `#include <micm/CPU.hpp>` and have everything they need without having to figure out which header files to load. This isn't intended to support precompiled headers, but maybe it would allow us to? Although, if we want that, we may consider moving straight to [c++ modules](https://en.cppreference.com/w/cpp/language/modules).
> > Also, the next bit of work I need to do is to setup musica to include the cuda solver. I think it's time that we make some better aliases so that the musica code can be smaller. This will allow me to put all of the aliases in one place. It may not be strictly necessary, but I think it's nice
> 
> I see. My concern with this approach is that users might end up including files that they don't need, which could increase memory usage. I'm not quite sure how this is related to making the musica code smaller. I haven't really seen this practice commonly used in other codebases, but if you think it's nice to have, I understand.

This would only increase compile time; the compiler only includes binary code which is needed by the program.

boulderdaze:
> > > > It reminds me of precompiled headers but it doesn't function like that, right? Is the purpose to have this collection of header files specifically for use in tests? If so, should they instead be relocated to the test directory?
> > > 
> > > 
> > > It's so that anyone could `#include <micm/CPU.hpp>` and have everything they need without having to figure out which header files to load. This isn't intended to support precompiled headers, but maybe it would allow us to? Although, if we want that, we may consider moving straight to [c++ modules](https://en.cppreference.com/w/cpp/language/modules).
> > > Also, the next bit of work I need to do is to setup musica to include the cuda solver. I think it's time that we make some better aliases so that the musica code can be smaller. This will allow me to put all of the aliases in one place. It may not be strictly necessary, but I think it's nice
> > 
> > 
> > I see. My concern with this approach is that users might end up including files that they don't need, which could increase memory usage. I'm not quite sure how this is related to making the musica code smaller. I haven't really seen this practice commonly used in other codebases, but if you think it's nice to have, I understand.
> 
> This would only increase compile time; the compiler only includes binary code which is needed by the program.

Yeah, that makes sense that linkers will strip out unused symbols, functions, etc.

PR #756, opened 2025-04-11
adding wall loss regression test

K20shores:
When I was investigating the release bug with music box, I ended up recreating the wall loss example in micm to pinpoint what was going on. I figured it would be good to go ahead and save this code as a regression test


boulderdaze (commented):
I don't know what the flow tube is, but if it is a specific mechanism, would it make sense to create a directory for it as we have one for RosebrockChapman?

K20shores:
> I don't know what the flow tube is, but if it is a specific mechanism, would it make sense to create a directory for it as we have one for RosebrockChapman?

I think the reason RosebrockChapman has its own directory is because it has a whole set of hard-coded files specific to that mechanism. The tests I added here just use our API, similar to what we do in the integration tests. I'm fine with putting it in a directory if you'd prefer though.

boulderdaze (approved):

boulderdaze:
> > I don't know what the flow tube is, but if it is a specific mechanism, would it make sense to create a directory for it as we have one for RosebrockChapman?
> 
> I think the reason RosebrockChapman has its own directory is because it has a whole set of hard-coded files specific to that mechanism. The tests I added here just use our API, similar to what we do in the integration tests. I'm fine with putting it in a directory if you'd prefer though.

Okay, I understand

PR #762, opened 2025-04-18
Cuda builder

K20shores:
In support of https://github.com/NCAR/musica/issues/299

Creates GPU aliases


CMakeLists.txt
```diff
@@ -33,10 +35,9 @@ option(MICM_BUILD_DOCS "Build the documentation" OFF)
 option(MICM_ENABLE_LLVM "Build with LLVM support for JIT-compiling" OFF)
 option(MICM_ENABLE_TESTS "Build the tests" ON)
 option(MICM_ENABLE_PROFILE "Profile MICM Solver" OFF)
-set(MICM_GPU_TYPE "None" CACHE STRING "The GPU type being targeted")
+option(MICM_ENABLE_CUDA "Enable CUDA support" OFF)
```
sjsprecious, line 38: This line seems redundant. When a user sets `MICM_GPU_TYPE`, it implies that CUDA is enabled. On the other hand, when `MICM_ENABLE_CUDA` option is enabled, a user still needs to specify what type of NVIDIA GPUs they want to compile.

cmake/dependencies.cmake
```diff
@@ -102,7 +102,7 @@ endif()
 ################################################################################
 # GPU Support
 
-if(NOT ${MICM_GPU_TYPE} STREQUAL "None")
+if(MICM_ENABLE_CUDA)
```
sjsprecious, line 105: If `MICM_ENABLE_CUDA` option is not used, we need to reverse the change here.

include/micm/CPU.hpp
```diff
@@ -11,37 +11,27 @@
...
-      micm::Solver<BackwardEulerStandardType, micm::State<DenseMatrixStandard, SparseMatrixStandard>>;
-
-  using RosenbrockThreeStageBuilder =
-      micm::CpuSolverBuilder<micm::RosenbrockSolverParameters, DenseMatrixVector, SparseMatrixVector>;
-  using BackwardEulerBuilder = micm::CpuSolverBuilder<
-      micm::BackwardEulerSolverParameters,
-      DenseMatrixVector,
-      SparseMatrixVector,
-      micm::LuDecompositionDoolittle>;
+  using DenseMatrixVector = VectorMatrix<double, MICM_DEFAULT_VECTOR_SIZE>;
```
sjsprecious, line 14: Is there a reason to remove `micm::` here?

sjsprecious (changes requested):
Thanks @K20shores . It looks good to me but I have some questions for clarification.

include/micm/CPU.hpp
K20shores: It's already inside of the `micm` namespace

CMakeLists.txt
K20shores: That's true. But we also end up setting this option later. Since we will eventually compile for all GPU types, I'm wondering if our default to should be to target all GPUs if there's no shock to performance. I can revert this though

include/micm/CPU.hpp
sjsprecious: Thanks for the clarification.

CMakeLists.txt
sjsprecious: If that is the case, I think we can just set `MICM_GPU_TYPE` to something like `all-major`.

I do not think that compiling for all GPU types will hurt the performance (maybe missing some architecture-specific optimization), but the binary file size may be larger and the compilation time may be longer. I still think it is the user's responsibility to tell MUSICA/MICM what type of GPUs are you going to use. In addition, if we plan to run on AMD or Intel GPUs in the future, the `MICM_ENABLE_CUDA` option won't work but the `MICM_GPU_TYPE` can still be used to point to the right libraries/compilers.

sjsprecious (changes requested):
Somehow the CUDA code is compiled but the CUDA unit tests are not enabled on Derecho.

boulderdaze (approved):
Looks good to me

sjsprecious (approved):
Thanks @K20shores . All the GPU tests are passed on Derecho's A100 GPUs using `gcc/12.4.0`.

mattldawson (approved):
nice!

CMakeLists.txt
```diff
@@ -33,10 +33,8 @@ option(MICM_BUILD_DOCS "Build the documentation" OFF)
 option(MICM_ENABLE_LLVM "Build with LLVM support for JIT-compiling" OFF)
 option(MICM_ENABLE_TESTS "Build the tests" ON)
 option(MICM_ENABLE_PROFILE "Profile MICM Solver" OFF)
+set(MICM_DEFAULT_VECTOR_SIZE "4" CACHE STRING "Default size for vectorizable matrix types")
```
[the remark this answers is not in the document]
K20shores: It's not needed

PR #764, opened 2025-05-01
Allow creation of State with specified number of grid cells

### LINKED ISSUE #686: Specify number of grid cells when creating State objects
Explore whether it is necessary to specify the number of grid cells when creating a `Solver` (as is currently done), or whether it is possible to just specify number of grid cells when creating instances of `State` from the `Solver%GetState()` function.

If this is possible, it would make it easier to use MICM in CAM-SIMA, which needs an instance of the `Solver` to be available before the number of grid cells being solved is known.

mattldawson:
Removes the requirement that the number of grid cells for one solver be fixed and known at construction. Instead, one solver can be used to create any number of State objects, each with their own specified number of grid cells

closes #686

### DIFF
--- a/include/micm/cuda/solver/cuda_state.hpp
+++ b/include/micm/cuda/solver/cuda_state.hpp
@@ -26,8 +26,9 @@ namespace micm
 
     /// @brief Constructor which takes the state dimension information as input
     /// @param parameters State dimension information
-    CudaState(const StateParameters& parameters)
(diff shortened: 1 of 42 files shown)


mattldawson:
@sjsprecious - I built the NVHPC Docker image but haven't tested it on Derecho yet. Would you be able to give it a try? If you're too busy, let me know and I can try it out. (This PR shouldn't affect the solver performance - it's all changes to creation of the State objects)

boulderdaze (approved):
Looks good!

PR #771, opened 2025-05-09
Add maximum grid cells function

mattldawson:
Adds a function to the Solver class that can be used to get the maximum number of grid cells that can fit in one group of vectorized State matrices.

### DIFF
--- a/include/micm/solver/solver.hpp
+++ b/include/micm/solver/solver.hpp
@@ -83,7 +83,7 @@ namespace micm
 
     /// @brief Returns the maximum number of grid cells per state
     /// @return Number of grid cells
-    /// @details This is the maximum number of grid cells that can that fit
+    /// @details This is the maximum number of grid cells that can fit
     ///          within one group for vectorized solvers. For non-vectorized solvers,
     ///          there is no limit other than the maximum size of a std::size_t.
     std::size_t MaximumNumberOfGridCells() const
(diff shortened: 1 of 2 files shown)


K20shores (approved):

include/micm/solver/solver.hpp
```diff
@@ -80,6 +81,23 @@ namespace micm
       return solver_.Solve(time_step, state, params);
     }
 
+    /// @brief Returns the maximum number of grid cells per state
+    /// @return Number of grid cells
+    /// @details This is the maximum number of grid cells that can that fit
+    ///          within one group for vectorized solvers. For non-vectorized solvers,
+    ///          there is no limit other than the maximum size of a std::size_t.
```
boulderdaze, line 88:
```suggestion
    /// @brief Returns the maximum number of grid cells per state
    /// @return Number of grid cells
    /// @details This is the maximum number of grid cells that can fit
    ///          within one group for vectorized solvers. For non-vectorized solvers,
    ///          there is no limit other than the maximum size of a std::size_t.
```
I am curious, in what occasion do you need to call this function?

boulderdaze (approved):

include/micm/solver/solver.hpp
mattldawson: I noticed it would be useful in the MUSICA library to know how to create a state with the number of grid cells equal to the vector dimension of the vectorized matrices. There wasn't an easy way to access this information before.

boulderdaze: I see, thanks!

PR #798, opened 2025-07-15
Adds features to support cloud chemistry

boulderdaze:
This PR 
- Introduces the `PhaseTransferRateConstant` class
- Updates comments
- Adds overridden `SetConcentration` functions to allow non-species elements to be set in `state` variables
- Adds a data member for non-species elements in the` System` class to support setting them in state variables
- Adds the unit tests for the newly added functions

Note: The phase transfer rate constant header isn't included in the process header yet, since it's a placeholder. The test is also not included but will be added in the future.


include/micm/system/system.hpp
```diff
@@ -16,34 +16,57 @@ namespace micm
...
    public:
-    /// @brief The gas phase is a micm::Phase and determines what species are present.
+    /// @brief The gas phase, defining a set of species present in the system
     Phase gas_phase_;
-    /// @brief This is a catchall for anything that is not the gas phase.
+    /// @brief Additional phases (e.g., aqueous, aerosol), mapped by name and representing non-gas phase
     std::unordered_map<std::string, Phase> phases_;
+    /// @brief Tracks non-phase elements (e.g., number concentrations) associated with a model.
+    ///        Elements are mapped using a prefix specific to the model's name and representation.
+    std::unordered_map<std::string, std::string> others_;
```
boulderdaze, line 34: Hi @mattldawson and @K20shores, do you have any concerns or thoughts about adding these variables to `System` (Also `SystemParameters` )? Since we want to track number concentration, which isn't a `Phase` , I created the `others `map to include them separately.


`State` variables to support modal model
```JSON
"cloud.primary.number_concentartion"
"cloud.primary.aqueous.SO4"
"..."
```

K20shores (approved):

Issue #804, opened 2025-07-29
Generalize Process class to to handle phase-transfer processes


boulderdaze:
The current `Process` structure is designed for gas-phase reactions and assumes clearly defined reactants and products. This doesn't fit well for condensation and evaporation, which involve phase transfer.

To support aerosol processes, the design should be made more flexible:
- Generalize `Process` to handle condensation and evaporation
- Improve extensibility for aerosol modeling

PR #805, opened 2025-08-01
Replace typedef Yield with struct Yield 

boulderdaze:
This PR:
- Replaces `using Yield = std::pair<Species, double>` with `struct Yield` which contains the same data members. This avoids duplicate typedef or `using` statement.
- Uses the explicit variable names instead of generic name(`species`, `coefficient` vs. `first`, `second`)
- Adds preliminary support for multi-phase reactions through `PhaseReaction`, `PhaseTransferReaction` struct. This is a work in progress.
- is one of several small PRs for #804 for easier review.

### DIFF
--- a/include/micm/process/reaction_types.hpp
+++ b/include/micm/process/reaction_types.hpp
@@ -25,10 +25,16 @@ namespace micm
           coefficient_(coefficient)
     {
     }
+
+    Yield(const Species& species)
+        : species_(species),
+          coefficient_(1.0)
+    {
+    }
   };
 
   /// @brief Represents a homogeneous or single-phase reaction with phase-specific
-  ///        reactants, products, and a rate constant
+  ///        reactants, products and a rate constant
   struct PhaseReaction
   {
     Phase phase_;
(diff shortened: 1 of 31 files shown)


K20shores (approved):

include/micm/process/reaction_types.hpp
```diff
@@ -0,0 +1,51 @@
...
+    Species species_;
+    double coefficient_;
+
+    Yield() = default;
+
+    Yield(const Species& species, double coefficient)
+        : species_(species),
+          coefficient_(coefficient)
+    {
+    }
```
mattldawson, line 27: would it make sense to add a constructor that just takes the species and sets the coefficient to 1 by default? I think most of the time, the coefficient will be 1

mattldawson (approved):
Looks good! Just one question, but I'm fine with it as-is also

include/micm/process/reaction_types.hpp
boulderdaze: Yes, that makes sense! updated.

PR #833, opened 2025-08-28
Update CUDA LU decomposition with user-defined vector length

### LINKED ISSUE #825: Update CUDA LU decomposition unit tests with user-defined vector length
Currently the CUDA LU decomposition implementation assumes that the vector length of CUDA sparse matrix is always equal to the number of grid cells. We may want to change it to allow any vector length, just like the update for CUDA sparse matrix (https://github.com/NCAR/micm/pull/814).

**Acceptance Criteria**
- All the unit tests pass
- The CUDA LU decomposition algorithm works with any vector length

**Ideas**
- Add customized GPU tests for CUDA LU decomposition with different vector lengths.
- Follow the changes from https://github.com/NCAR/micm/pull/814 for the LU sparse matrix.

sjsprecious:
This PR updates the CUDA LU decomposition kernel to allow a user-defined vector length.

All the unit tests pass on Derecho's A100 GPUs.

fix #825


mattldawson (approved):
awesome!

PR #838, opened 2025-09-02
Design concept for phase transfer process set

boulderdaze:
This draft is for discussing a design concept to set forcing terms, and later Jacobian terms, for phase transfer processes that involve aerosol models.

@mattldawson, @K20shores, @dwfncar, I'd love to hear your thoughts on the callback-based approach for setting the forcing terms.


boulderdaze:
Closing the draft as we discussed it in the design meeting

PR #840, opened 2025-09-08
Update CUDA ProcessSet unit tests with user-defined vector length

sjsprecious:
This PR updates the CUDA ProcessSet unit tests (AddForcingTerms and SubtractJacobianTerms) with the user-defined vector length.

All the unit tests pass on Derecho's A100 GPU nodes.

fix #839

PR #849, opened 2025-09-23
Temporary fix for uncaught MICM Exceptions in MUSICA

boulderdaze:
MUSICA fails to catch certain exceptions thrown by MICM due to incompatible error handling.
This is a temporary fix. Will address the following issue:
- https://github.com/NCAR/micm/issues/848


boulderdaze:
Addressed in #840

PR #869, opened 2025-11-06
857 pass number of grid cells to cuda rosenbrock

### LINKED ISSUE #857: Pass number of grid cells to cuda rosenbrock
Now that the vector length and the number of grid cells differ for the GPU solver we need to get the number of grid cells to the cuda solver.

The cuda solver wants to know the number of grid cells at construction time.

https://github.com/NCAR/micm/blob/37cb894613b3c820d3b511108d75401ebdaa3dfe/include/micm/cuda/solver/cuda_rosenbrock.hpp#L78-L80

But that cannot happen now. We setup the CPU solver so that it could work with any state. A state can have any number of grid cells, so the interfaces have different things they expect to be true.

**Acceptance criteria**
- A cuda solver can work with any state, regardless of the number of grid cells

**Ideas**
- Fully remove [`CudaRosenbrockSolverParam`](https://github.com/NCAR/micm/blob/37cb894613b3c820d3b511108d75401ebdaa3dfe/include/micm/cuda/util/cuda_param.hpp#L96-L105)
  - place all of this information on the [`RosenbrockTemporaryVariables`](https://github.com/NCAR/micm/blob/main/include/micm/solver/rosenbrock_temporary_variables.hpp)
    - Currently initialized [here](https://github.com/NCAR/micm/blob/main/include/micm/cuda/solver/cuda_rosenbrock.hpp#L77-L85)
    - and maybe only keep the jacboian diagonal elements?
  - The [`NormalizedErrorDriver`](https://github.com/NCAR/micm/blob/50b0dae81c274cfe2558821755c1fd689f0d6075/src/solver/rosenbrock.cu#L268) uses the devstruct, but it doesn't have to. It also would need to use something different if we remove the `CudaRosenbrockSolverParam`
    - inputs to this driver already come from the temporary variables. We need to add vectors for the error input and output on the temporary variables
    - I think this error input/output is a vector the size of the number of grid cells.
      - Assigned in [rosenbrock.cu](https://github.com/NCAR/micm/blob/main/src/solver/rosenbrock.cu#L42-L60). 
      - The size is `number of grid cells` * `number of speices`
      - It seems we only need this information if we are using our custom kernel to compute the error. We wouldn't need to data to compute the errors if we use the cuda error calculation always. We'd only need to add the jacobian information to the temporary variables, I think
        - That would save on the amount of data we copy (which isn't much, but anything helps?), and remove a custom kernel

K20shores:
Closes #857 

The number of grid cells was only needed on construction of the cuda solver to appropriately size the errors.

I moved the errors to the cuda state, which means that GPU solver is now agnostic to any chemistry configured with micm, because it only depends on the cuda state. The cuda state now owns the errors input, output, and jacobian diagonal elements and takes care of copying them to the device

### DIFF
--- a/include/micm/cuda/solver/cuda_rosenbrock.cuh
+++ b/include/micm/cuda/solver/cuda_rosenbrock.cuh
@@ -11,7 +11,7 @@ namespace micm
 {
   namespace cuda
   {
-    /// @brief Compute alpha - J[i] for each element i at the diagnoal of Jacobian matrix
+    /// @brief Compute alpha - J[i] for each element i at the diagonal of Jacobian matrix
     /// @param jacobian_param Dimensions and device data pointers for the Jacobian
     /// @param alpha scalar variable
     /// @param devstruct device struct including the locations of diagonal elements of the Jacobian matrix


include/micm/cuda/solver/cuda_rosenbrock.cuh
```diff
@@ -43,6 +36,8 @@ namespace micm
         const CudaMatrixParam& errors_param,
         const CudaMatrixParam& absolute_tolerance_param,
         const double relative_tolerance,
-        CudaRosenbrockSolverParam devstruct);
+        const std::size_t errors_size_,
+        double* errors_input,
+        double* errors_output);
```
sjsprecious: see the comment above.

include/micm/cuda/solver/cuda_state.hpp
```diff
@@ -19,9 +19,21 @@ namespace micm
...
     CudaMatrixParam absolute_tolerance_param_;
 
+    // for NormalizedError function
+    std::size_t errors_size_;
+    double* errors_input_ = nullptr;
+    double* errors_output_ = nullptr;
+
+    // for AlphaMinusJacobian function
+    std::size_t jacobian_diagonal_elements_size_;
+    std::size_t* jacobian_diagonal_elements_ = nullptr;
```
sjsprecious: so maybe make these as one struct so that we can pass the struct around rather than the raw pointers and size?

include/micm/cuda/solver/cuda_rosenbrock.cuh
```diff
@@ -28,7 +19,9 @@ namespace micm
     void AlphaMinusJacobianDriver(
         CudaMatrixParam& jacobian_param,
         const double& alpha,
-        const CudaRosenbrockSolverParam& devstruct);
+        std::size_t jacobian_diagonal_elements_size,
+        const std::size_t* jacobian_diagonal_elements
```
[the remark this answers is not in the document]
K20shores: I'll do that

PR #886, opened 2025-12-10
removing llvm

K20shores:
Removes the JIT solver

We don't use this solver, and we have to use a very outdated version of llvm. If we still want a JIT solver, we should look into [MLIR](https://mlir.llvm.org/), which seems to be a better option anyway.

I personally have no problem removing this since it's in our git history and should be relatively easy to add back, but as of now, anytime we make a change to the CPU/GPU interfaces, we have to replicate that for the JIT solver that we aren't using and aren't shipping with [musica](https://github.com/NCAR/musica). Maybe it's time we let it go

Let me know your thoughts


sjsprecious:
I agree that it is better to remove the JIT version for now and we can always recover it when we find it is helpful. Thanks for doing it!

Issue #913, opened 2026-02-02
Evaluate refactoring `ConstraintSet` to reduce index indirection overhead


boulderdaze:
`ConstraintSet` currently precomputes global dependency IDs from `variable_map`, then builds a temporary indices vector and passes it to each Constraint at evaluation time. This design preserves decoupling but introduces extra indirection and per-evaluation allocations.

Problem:
- Additional array lookups inside constraints
- Temporary indices vector created for each constraint evaluation
- Less cache-friendly access compared to precomputed global index vectors

ConstraintSet could be refactored to:
- Precompute and store per-constraint global index vectors (like `ProcessSet` does)
- Avoid constructing temporary indices vectors at evaluation time
- Provide a lower-level or optimized evaluation path within ConstraintSet

Issue #912, opened 2026-02-02
Evaluate refactoring `Residual()` to take `State` reference


boulderdaze:
`EquilibriumConstraint::Residual()` takes raw pointers to concentration data and indices. The data is owned by the state object. In the integration test, it’s unclear how the subset of concentrations used for this constraint is constructed and kept consistent with the state.

Refactor `Residual()` to take a const reference to the state object directly, and derive the required concentration indices from that state instead of passing raw pointers.

PR #914, opened 2026-02-02
Review dae 1 constraint classes

boulderdaze:
- Removed md files. 
- Created the following two PRs to address them as separate issues:
  - #912
  - #913

Note:
I don't think the combined three PRs for implementing the `Constraint` class is big. We could aim to request a review from Matt only for the final version that includes rosenbrock integration (see the file diffs in PR #909), instead requesting reviews for all three PRs. I'd like to hear your thoughts on this. @dwfncar, @K20shores 

I've also decide to keep the `name_` variable of `Constraint` class for now even though it seems to be used solely for logging error messages for unknown species. 
If we refactor `Constraint` class described in PR 913, we probably won’t need the name variable for debugging anymore.


boulderdaze:
If you agree with the note mentioned in the PR description, then once this review PR is merged into the David's original branch, his branch should then be merged into the dev-3-rosenbrock intergration PR so Matt can review everything together.

K20shores (approved):

PR #962, opened 2026-04-07
Allow external model constraints

mattldawson:
This PR adds the ability for external models to contribute constraints to the DAE system.

__Additionally:__
* adds a finite-difference jacobian calculator for verifying consistent forcing (or residual) and jacobian functions. This is usable by external models as well for constraints and processes
* introduces an algebraic constraint initialization step to the solver when constraints are provided. This avoids making users calculate consistent initial conditions themselves in most cases.
* also introduces a separate pre-solve step that allows calculation of constraint parameters from state variables. This is used in MIAM for mass conservation where the linear constraint constant is calculated from the sum of the initial species concentrations.

The implementation relied heavily on Copilot, and should be thoroughly evaluated.

__Steps taken to verify Copilot's implementation__
* thorough review of modified source code; high-level review of new tests; only additions to existing test logic, no changes of existing tests
* several integration tests comparing results from fully kinetic systems vs. constrained systems that should reach the same steady state are included
* comparisons of constrained systems where the constraints are provided by an external model vs. those built from existing MICM constraints are also included
* tests using the new finite-difference jacobian checker
* the new external constraint options are used in [MIAM PR 21](https://github.com/NCAR/miam/pull/21) and [MUSICA PR 842](https://github.com/NCAR/musica/pull/842) for providing constraints from MIAM sufficient to recreate a full cloud chemistry mechanism. Everything seems to be working correctly

__Concerns__
* the implementation of the initialization step for the algebraic constraints looks correct to me, but should be looked at by someone more familiar with these solvers. @davidfillmore - would you mind looking over these changes?


include/micm/constraint/constraint_set.hpp
```diff
@@ -321,6 +344,125 @@ namespace micm
...
+      for (const auto& model : external_constraint_models_)
+      {
+        auto alg_names = model.algebraic_variable_names_func_();
+        if (alg_names.empty())
+          continue;
+        auto param_names = model.state_parameter_names_func_();
+        names.insert(param_names.begin(), param_names.end());
+      }
+      return names;
+    }
```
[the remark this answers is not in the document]
mattldawson: Updated to return `std::vector<std::string>` and detect duplicates in `ExternalConstraintParameterNames()`

include/micm/solver/solver_builder.hpp
```diff
@@ -109,6 +111,42 @@ namespace micm
...
+      else if constexpr (HasProcesses<std::decay_t<ExternalModel>>)
+      {
+        external_models_.emplace_back(
+            ExternalModelProcessSet<DenseMatrixPolicy, SparseMatrixPolicy>{ std::forward<ExternalModel>(model) });
+      }
+      else
+      {
+        external_constraint_models_.emplace_back(
+            ExternalModelConstraintSet<DenseMatrixPolicy, SparseMatrixPolicy>{ std::forward<ExternalModel>(model) });
+      }
```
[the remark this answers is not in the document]
mattldawson: updated to accept model by value instead of r-value reference. the models aren't large and this is done once during setup

include/micm/util/jacobian_verification.hpp
```diff
@@ -0,0 +1,213 @@
...
+        {
+          // Forward difference: need f(x) as well
+          forcing_base.Fill(0.0);
+          forcing_func(base_variables, forcing_base);
+          for (std::size_t row = 0; row < num_species; ++row)
+          {
+            result[block][row * num_species + col] = (forcing_plus[block][row] - forcing_base[block][row]) / h;
+          }
+        }
+      }
```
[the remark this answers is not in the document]
mattldawson: this is only used for testing, it doesn't need to be called with a large number of blocks/grid cells

include/micm/solver/rosenbrock.inl
```diff
@@ -383,4 +394,144 @@ namespace micm
...
+
+      if (nan_detected)
+        return SolverState::NaNDetected;
+      if (inf_detected)
+        return SolverState::InfDetected;
+    }
+
+    // Did not converge within max iterations
+    return SolverState::ConstraintInitializationFailed;
+  }
```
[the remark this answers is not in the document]
mattldawson: Modified to throw an error when constraints are included in a CUDA solver build

test/unit/solver/test_constraint_initialization.cpp
```diff
@@ -0,0 +1,284 @@
...
+  state.variables_[0][C_idx] = SimpleConstrainedSystem::K_eq * 0.5;  // C = K_eq * B = consistent
+  state.conditions_[0].temperature_ = 298.15;
+  state.conditions_[0].pressure_ = 101325.0;
+
+  solver.UpdateStateParameters(state);
+
+  double A_before = state.variables_[0][A_idx];
+  double B_before = state.variables_[0][B_idx];
+  double C_before = state.variables_[0][C_idx];
+
```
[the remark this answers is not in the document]
mattldawson: removed

include/micm/solver/solver_builder.hpp
```diff
@@ -99,13 +100,50 @@ namespace micm
     }
 
     /// @brief Add processes from an external model
-    /// @param model The external model
+    /// @param model The external model (taken by value; caller decides whether to copy or move)
     /// @return Updated SolverBuilder
+    /// @deprecated Use AddExternalModel() instead
     template<class ExternalModel>
-    SolverBuilder& AddExternalModelProcesses(ExternalModel&& model)
+    SolverBuilder& AddExternalModelProcesses(ExternalModel model)
```
boulderdaze: This function does what `AddExternalModel ` does for process-only models. Should we remove it instead of marking deprecated?
It could be confusing even for developers to see two similar APIs. (I was confused by it at first as well. ) Since this is still a new addition, removing it now would still be safe.

include/micm/solver/solver_builder.inl
```diff
@@ -309,19 +401,38 @@ namespace micm
...
+      // Set forcing, jacobian, updating state param functions
       // The species map and parameter map are used to set indices in the state variables
       // and custom parameters.
       constraint_set.SetConstraintFunctions(species_map, params_map, jacobian);
+      constraint_set.SetExternalModelConstraintFunctions(params_map, species_map, jacobian);
+
+      // Add external constraint parameter update functions to the pipeline
+      auto ext_constraint_param_funcs = constraint_set.GetExternalUpdateStateParamFunctions();
+      update_state_param_funcs.insert(
+          update_state_param_funcs.end(), ext_constraint_param_funcs.begin(), ext_constraint_param_funcs.end());
```
boulderdaze: Not a big deal, but for consistency across the code base,
`ext_constraint_param_funcs` could be inserted after `constraint_param_funcs`.

boulderdaze (approved):
Looks good! I have two suggestions, though they should be easy to address if you agree.
You also added tests comparing `ProcessSet` analytical Jacobian against finite-difference approximation. I'm curious what led to adding finite differences at this stage with the introduction of constraints?

include/micm/util/jacobian_verification.hpp
```diff
@@ -0,0 +1,213 @@
...
+  /// where variables has shape [num_blocks x num_species] and forcing has the same shape.
+  /// Callers should bind any additional arguments (rate constants, state parameters, etc.)
+  /// into the callable via a lambda capture.
+  ///
+  /// Returns a DenseMatrixPolicy of shape [num_blocks x (num_species * num_species)] where
+  /// element [block][row * num_species + col] = df_row/dx_col.
+  ///
+  /// When a perturbation would push a variable below zero, one-sided differences are used instead.
+  template<class DenseMatrixPolicy, class ForcingFunc>
+  DenseMatrixPolicy FiniteDifferenceJacobian(
```
K20shores, line 41: Is this only used in the tests?

K20shores (approved):
It seems fine to me, but it seems that `FiniteDifferenceJacobian` is only used in the tests? Maybe I missed it elsewhere

include/micm/util/jacobian_verification.hpp
mattldawson: yes - it's specifically for testing whether our forcing and jacobian function implementations are consistent with each other. It (expensively) calculates Jacobian contributions based on the forcing function and compares them with what comes out of the Jacobian function to make sure they're consistent

K20shores: Is it necessary? Isn't this mathematically provable?

mattldawson: It's not necessary, but imo it's a pretty valuable test.

Deriving on paper the partial derivatives of the rates is one thing, but implementing them is another. As the equations start to get [more complex](https://open-atmos.github.io/camp/html/sub__model___u_n_i_f_a_c_8c.html#a0301d06b09af90fcfe49f9a1ba7c6742) as we move to mixed and condensed phase processes, the Jacobian functions get very tricky to write, at least in my experience (Claude probably is not as error prone as I am).

On top of being tricky to write, when the forcing and Jacobian functions are out-of-sync, the errors are very subtle and difficult to debug. For example, the solver will often converge but with a few more iterations than it should need, and/or with answers that might be subtly off. Or, the solver fails under some conditions and not other.

Also, if we start to rely more heavily on AI implementations, this is an independent way to give some assurance that it is formulating these functions correctly.

But, I'm fine not including it or not.

dwfncar (approved):

davidfillmore (approved):

include/micm/solver/solver_builder.hpp
mattldawson: Good point. I will remove it. I agree it's confusing, and I just forgot how new it is, so nothing is really using it yet.

mattldawson: removed!

include/micm/solver/solver_builder.inl
mattldawson: reordered!

include/micm/util/jacobian_verification.hpp
mattldawson: I'm gonna merge this in since everyone approved so I can keep moving on the cloud chemistry, but we can revisit this later if you want. It would be easy enough to remove this if it's not needed

PR #966, opened 2026-04-15
Rate constant refactor

K20shores:
- Refactor to share the rate constant functions between cpu/gpu
- Place all rates into a structure of arrays which we presort to encourage cache-reuse
- requires us to use c++ 23


include/micm/solver/solver.hpp
```diff
@@ -165,44 +171,37 @@ namespace micm
...
-      Process::CalculateRateConstants<DenseMatrixType>(processes_, state);
+      // Dispatch to GPU path if the RatesPolicy (e.g. CudaProcessSet) exposes
+      // GpuCalculateRateConstants; otherwise use the CPU path.
+      if constexpr (requires { solver_.rates_.GpuCalculateRateConstants(store_, state); })
+        solver_.rates_.GpuCalculateRateConstants(store_, state);
+      else
+      {
+        ReactionRateStore::EvaluateCpuRates(store_, state);
+        ReactionRateStore::CalculateRates(store_, state);
+      }
```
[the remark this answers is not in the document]
K20shores: Fixed — external update functions now run before rate constant calculation so user-defined and surface types read up-to-date `custom_rate_parameters_`.

include/micm/solver/solver.hpp
```diff
@@ -28,6 +29,7 @@ namespace micm
     System system_;
     std::vector<std::function<void(const std::vector<micm::Conditions>&, DenseMatrixType&)>>
         update_state_parameters_functions_;
+    ReactionRateStore store_;
```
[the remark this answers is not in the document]
K20shores: Fixed — `store_` is now moved in both the move constructor and move assignment operator.

src/CMakeLists.txt
```diff
@@ -65,3 +65,7 @@ if (MICM_ENABLE_PROFILE)
   target_compile_definitions(micm INTERFACE MICM_PROFILE)
 endif()
 
+if (MICM_ENABLE_TESTS)
+  target_compile_definitions(micm INTERFACE MICM_ENABLE_TESTS)
+endif()
+
```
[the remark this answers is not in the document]
K20shores: Valid concern. `MICM_ENABLE_TESTS` is INTERFACE because `cuda_state.hpp` is a public header and test binaries need the definition to see the `CopyToHost` call at compile time. Changing to PRIVATE on the library would require explicitly adding the definition to every CUDA test target. Happy to do that as a follow-up, but leaving it INTERFACE for now to keep the scope contained.

test/unit/process/rate_constant/test_arrhenius_rate_constant.cpp
sjsprecious, line 1: Do we have the cuda tests for each reaction type now?

K20shores: i made claude add a test for this

include/micm/process/rate_constant/rate_constant_functions.hpp
```diff
@@ -0,0 +1,252 @@
...
+{
+
+  /// @brief Calculate Arrhenius rate constants for a contiguous block of reactions.
+  ///        k = A * exp(C/T) * (T/D)^B * (1 + E*P)
+  /// @param params      Array of Arrhenius parameters, length n
+  /// @param n           Number of reactions
+  /// @param temperature Temperature [K]
+  /// @param pressure    Pressure [Pa]
+  /// @param output      Destination array of length n
+  MICM_HOST_DEVICE inline void CalculateArrhenius(
```
sjsprecious: Now I see how you can reuse the same kernel for CPU and GPU. This Macro is a smart idea.

test/integration/analytical_policy.hpp
sjsprecious, line 13: I changed `NUM_CELLS=163840` and updated `L = 163840` in the  `test_cuda_analytical_rosenbrock.cpp`, and the unit test fails.

sjsprecious (changes requested):
Looks like the CUDA integration test will fail for larger grid cells, after calculating the rate constants on the GPU.

test/integration/analytical_policy.hpp
K20shores: not good. I don't have the time right now to fix this since I have to prepare for a tutorial. I can look into this next week

include/micm/process/rate_constant/rate_constant_functions.hpp
K20shores: I still think we should be able to mark this `constexpr` and note even need the macro. That was my understanding, at least. I will investigate further next week

include/micm/process/reaction_rate_store.hpp
```diff
@@ -0,0 +1,377 @@
...
+  #define M_PI 3.14159265358979323846
+#endif
+
+namespace micm
+{
+
+  /// @brief Structure-of-arrays store for all reaction rate constant parameters.
+  ///
+  /// Processes must be sorted by RateConstantTypeOrder before BuildFrom is called.
+  struct ReactionRateStore
```
boulderdaze: Since this is a rate constant, it would be better to make the struct more descriptive
```suggestion
  struct ReactionRateConstantStore
```

include/micm/process/reaction_rate_store.hpp
```diff
@@ -0,0 +1,377 @@
...
+          for (const auto& entry : store.lambda_entries)
+            state.rate_constants_[i_cell][entry.rc_index] = entry.source->lambda_function_(cond);
+        }
+      }
+    }
+
+    /// @brief Calculate all analytic rate constants into state.rate_constants_.
+    ///        Lambda entries are untouched; parameterized multipliers applied last.
+    template<class StatePolicy>
+    static void CalculateRates(const ReactionRateStore& store, StatePolicy& state)
```
boulderdaze: The functions calculate rate constants, not rate, and the results are stored in `state.rate_constants_`. I think the naming should be updated for all four functions:
```yaml
CalculateRateConstants()
CalculateRateConstantsVectorized()
CalculateRateConstantsScalar()
EvaluateCpuRateConstants()
```

include/micm/process/reaction_rate_store.hpp
```diff
@@ -0,0 +1,377 @@
...
+
+  /// @brief Structure-of-arrays store for all reaction rate constant parameters.
+  ///
+  /// Processes must be sorted by RateConstantTypeOrder before BuildFrom is called.
+  struct ReactionRateStore
+  {
+    // ----------------------------------------------------------------
+    // Analytic types (GPU-safe parameter structs)
+    // ----------------------------------------------------------------
+    std::vector<ArrheniusRateConstantParameters> arrhenius;
```
boulderdaze: This differs from our usual trailing underscore `_` convention for member data. I'm okay with it given that the struct has public members, but it does introduce some inconsistency. 
 
```suggestion
std::vector<ArrheniusRateConstantParameters> arrhenius_;
```

test/integration/analytical_policy.hpp
sjsprecious: No problem. I still do not fully understand the CUDA kernel here so we can schedule another live coding session to address this together.

include/micm/process/rate_constant/rate_constant_functions.hpp
sjsprecious: I will be interested to see if `constexpr` can do this magic here.

include/micm/solver/solver.hpp
```diff
@@ -158,51 +166,49 @@ namespace micm
       return processes_;
     }
 
+    const ReactionRateStore& GetStore() const
```
boulderdaze: The function name `Solver::GetStore()` is confusing. Can you make it more descriptive?

```suggestion
    const ReactionRateConstantStore& GetRateConstantStore() const
```

include/micm/solver/solver.hpp
```diff
@@ -158,51 +166,49 @@ namespace micm
...
         update_func(state.conditions_, state.custom_rate_parameters_);
+
+      // Dispatch to GPU path if the RatesPolicy (e.g. CudaProcessSet) exposes
+      // GpuCalculateRateConstants; otherwise use the CPU path.
+      if constexpr (requires { solver_.rates_.GpuCalculateRateConstants(store_, state); })
+        solver_.rates_.GpuCalculateRateConstants(store_, state);
+      else
+      {
+        ReactionRateStore::EvaluateCpuRates(store_, state);
+        ReactionRateStore::CalculateRates(store_, state);
```
boulderdaze: Same observation: 
the CPU specific function `CalculateRates` uses a different naming pattern than the GPU version, `GpuCalculateRateConstants`.

boulderdaze (changes requested):
Looks good! I have a couple of suggestions.

test/integration/analytical_policy.hpp
sjsprecious: It is fixed by [769b752](https://github.com/NCAR/micm/pull/966/commits/769b752188b737710cbc9aefc14af2db914f035c). Thanks for doing it.

boulderdaze (approved):

sjsprecious (approved):
Thanks @K20shores for working on this refactor. Looks good to me.

I really like the `MICM_CONSTEXPR` mechanism that could potentially remove many duplications of CUDA code for calculation.

PR #980, opened 2026-04-30
Set c++ standard to 20

K20shores:
In #966, I thought we needed to use C++23 because that's when constexpr math functions were added [P1383R2](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p1383r2.pdf). 

However, we only need to mark the functions `constexpr` so that [`--expt-relaxed-constexpr`](https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/#expt-relaxed-constexpr-expt-relaxed-constexpr) allows CUDA device code to call them. 

NOAA cannot compile with C++20 on GCC11 because GCC11 does not have `std::format`. See our regression [here](https://github.com/NCAR/MechanismConfiguration/pull/263), so it's unlikely they can use C++23.

However, we apparently don't even need it. Drop the version back to C++20

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,6 +32,10 @@ option(MICM_ENABLE_MEMCHECK "Enable memory checking in tests" OFF)
 option(MICM_BUILD_DOCS "Build the documentation" OFF)
 option(MICM_ENABLE_TESTS "Build the tests" ON)
 option(MICM_BUILD_SHARED_LIBS "Build micm_cuda as shared libraries" OFF)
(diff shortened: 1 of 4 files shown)


sjsprecious (approved):
Thanks @K20shores for figuring it out. It is confusing that C++20 supports it already but the report was in 2023?

K20shores:
> Thanks @K20shores for figuring it out. It is confusing that C++20 supports it already but the report was in 2023?

It's not surprising. The standards just say what is compliant for that year, but there's nothing stopping a previous version from already implementing something

sjsprecious:
Ah thanks for your clarification.

PR #984, opened 2026-05-13
Address warnings as errors

K20shores:
Addressing all compile warnings errors

- adds a cmake option `MICM_COMPILE_WARNING_AS_ERROR`
- when turned on, it sets the equivalent of `-Werror`
- use this option in all of our CI builds
- during this PR, I (and claude) fixed all of the warnings that showed up as errors across all of our compilers
- also addressed a longstanding TODO in `test/regression/RosenbrockChapman/regression_test_dforce_dy_policy.hpp`

Closes #988


include/micm/constraint/types/linear_constraint.hpp
```diff
@@ -108,7 +108,7 @@ namespace micm
       std::size_t row_idx = info.row_index_;
 
       return DenseMatrixPolicy::Function(
-          [coeffs, species_indices, constant, row_idx](auto&& state, auto&& params, auto&& force)
+          [coeffs, species_indices, constant, row_idx](auto&& state, auto&&, auto&& force)
```
boulderdaze: I see that this change across the codebase is intended to remove unused variable warnings, but I think it really hurts readability. A user (or a developer) would have no idea what the argument represents, especially when using `auto`, which further hurts visibility and would require tracing back to the call site.

In some code, the type is obvious without a variable name, for example, (eg. `ConstraintInfo&` ), that would be fine omitting. For generic types like `std::size_t`, omitting the variable name makes the code harder to follow. I don't think silencing the warning should take priority over readability.

K20shores: You bring up a good point. To accept this, I'll have to add `-Wno-unused-parameter` to our list of compile flags. Is that a fair compromise? I prefer having as few warnings as possible in our code

K20shores: Actually, I just realized that the unused paramters comes from `-Wextra`, which I only added in musica. I could relax musica to just `-Wall -Werror` and allow all of these unused paramters

K20shores: Ok, I've reverted the changes that removed unsued paramters. I also added `-Wno-unused-parameter`, even though we don't need it, but at least it carries the idea behind this conversation that we explicitly want to allow it

PR #989, opened 2026-05-19
Remove phase transfer classes and redundant copyright line

boulderdaze:
This PR:
- Removes phase transfer classes that MIAM implements
  - Includes phase transfer process, phase transfer process builder, transfer coefficient, henrys law constant and its test files 
- Removes the redundant copyright line in `README.md`
   - In the LICENSE file (line 189), the data don't match with the one in README, so it was updated to match it.  

As a next step, an issue has been created to remove a single-type variant. Since it now only holds `ChemicalReaction`, it no longer has to be a variant.
- https://github.com/NCAR/micm/issues/990

```C++
using ProcessVariant = std::variant<ChemicalReaction>;
```


include/micm/process/process.hpp
```diff
@@ -19,12 +18,12 @@ namespace micm
   class Process
   {
    public:
-    using ProcessVariant = std::variant<ChemicalReaction, PhaseTransferProcess>;
+    using ProcessVariant = std::variant<ChemicalReaction>;
```
K20shores, line 21: Should we think about removing the variant since we onkly have one type now?

```suggestion
```

include/micm/process/process.hpp
```diff
@@ -19,12 +18,12 @@ namespace micm
...
 
     ProcessVariant process_;
 
     template<typename T>
-      requires std::same_as<std::decay_t<T>, ChemicalReaction> || std::same_as<std::decay_t<T>, PhaseTransferProcess>
+      requires std::same_as<std::decay_t<T>, ChemicalReaction>
```
K20shores, line 26: If we do remove the variant, we no longer need lines like this

K20shores (changes requested):

include/micm/process/process.hpp
boulderdaze: Yeah, I mentioned in the description that this will be handled in a separate issue. Since it blocks the MIAM PR, I wanted to remove `HenrysLawConstant` first and address it in a separate PR later.
- #990

include/micm/process/process.hpp
boulderdaze: Same comment as above

K20shores (approved):

PR #998, opened 2026-05-27
Enabling clang tidy CI failures

K20shores:
Closes #988 

- enables clang tidy to cause failures
- places all checks in the dockerfile so they can easily be run on local computers
- nothing is addressed yet because there's too many to comfortably do in one PR
  - #997 lists all known failures that will be fixed in future PRs

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,24 +1,4 @@
 ---
-# All checks are currently disabled (Checks: '-*').
-# Enable them incrementally in separate PRs, one group at a time.
-#
-# Intended check groups to enable (each requires a clean-up PR first):
-#
-#   readability-braces-around-statements
-#   readability-identifier-naming  (with CheckOptions below)
-#   readability-* (remaining)
-#   modernize-pass-by-value
-#   modernize-use-emplace
-#   modernize-use-auto
-#   modernize-return-braced-init-list
-#   modernize-deprecated-headers
-#   modernize-* (remaining)
-#   performance-unnecessary-value-param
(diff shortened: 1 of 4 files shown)


.clang-tidy
```diff
@@ -1,52 +1,103 @@
...
+  - key:             readability-identifier-naming.VariableCase
+    value:           lower_case
+  - key:             readability-identifier-naming.ParameterCase
+    value:           aNy_CasE
   - key:             readability-identifier-naming.MemberCase
     value:           aNy_CasE
   - key:             readability-identifier-naming.MemberSuffix
     value:           _
+  - key:             readability-identifier-naming.ConstantCase
+    value:           lower_case
```
boulderdaze: I know at least one constant `GAS_CONSTANT` is not lower case.

K20shores: true, do you have a preference for how we should name these? We have said in the past that we adhere to the google style guide, but they have a [different naming convention](https://google.github.io/styleguide/cppguide.html#Constant_Names) for constants. I don't actually care what pattern we you, so happy to make clang tidy do whatever you like most

boulderdaze: I think we can just use uppercase since the constants defined in `util/constants.hpp` use uppercase, though I could be wrong in other part of the code. In that case, we might want  to disable this check. I also don’t have a preference about code style.

K20shores: @boulderdaze we don't have to disable it, I think we can just set its value to upper case

boulderdaze (approved):

PR #1004, opened 2026-05-28
Refactor Solver Builder API to avoid passing ExternalModel object twice

boulderdaze:
System takes model in its constructor, but the solver builder also requires calling AddExternalModel(model). This makes the builder API redundant and confusing.
```C++
// Current
auto solver = CpuSolverBuilder<RosenbrockSolverParameters>(
                    RosenbrockSolverParameters::FourStageDifferentialAlgebraicRosenbrockParameters())
                    .SetSystem(System(gas_phase, model));
                    .AddExternalModel(model)
                    .Build();

// This PR
auto solver = CpuSolverBuilder<RosenbrockSolverParameters>(
                    RosenbrockSolverParameters::FourStageDifferentialAlgebraicRosenbrockParameters())
                    .SetSystem(System(gas_phase));
                    .AddExternalModel(model)
                    .Build();
```

This PR:
- Makes `AddExternalModel` responsible for capturing variable/parameter names and size of external model state.
- `System` defines a gas system only.
- `SolverBuilder::AddExternalModel` stores both the state-query wrapper and the compute wrappers. During `Build()`, it updates the `species_map` with external model contributions before constructing the rates/constraint objects.
- Renames variables to maintain consistency
```yaml
external_models                     -> external_process_sets
external_constraint_models_         -> external_constraints_
                                    -> external_systems_ (added)
external_model_forcing_functions_   -> external_forcing_functions_
external_model_jacobian_functions_  -> external_jacobian_functions_
```
- Removes struct `SystemParameters` that only contains gas_phase now
- Closes #983

### DIFF
--- a/include/micm/system/system.hpp
+++ b/include/micm/system/system.hpp
@@ -2,143 +2,61 @@
 // SPDX-License-Identifier: Apache-2.0
 #pragma once
 
-#include <micm/external_model.hpp>
 #include <micm/system/phase.hpp>
-#include <micm/system/species.hpp>
 
 #include <functional>
-#include <memory>
-#include <set>
 #include <string>
-#include <unordered_map>
 #include <vector>
 
 namespace micm
 {
 
-  struct SystemParameters
-  {
-    /// @brief  @brief The gas phase
-    Phase gas_phase_{};
-    /// @brief External models (e.g., aerosol models) that provide additional components to the system
-    std::vector<ExternalModelSystem> external_models_{};
-  };
-
-  /// @brief Represents the complete chemical state of a grid cell
-  ///        Includes the gas phase and other associated phases, each with their own set of species.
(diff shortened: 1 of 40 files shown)


include/micm/system/system.hpp
K20shores, line 1: I wonder if we even need this anymore. It's purpose was to hold all of the phases, all of the models, but now we've reduced it to only the gas phase. Maybe we remove the system and just set a gas phase. This would be a major breaking api change, but I don't really see much utility to system

K20shores (approved):

include/micm/system/system.hpp
boulderdaze: I had a similar thought that `system` just becomes a wrapper around `gas_phase` and mainly exists to avoid breaking the current APIs. 
Issue was created:
- #1005

PR #1045, opened 2026-07-23
Remove lower and upper matrix policies

mattldawson:
This PR removes the `LMatrixPolicy` and `UMatrixPolicy`. We added these originally thinking we might be able to improve performance by using different matrix orderings for lower and upper matrices. In the end, it had little effect (@sjsprecious - is this how you remember it also?).

Currently, they would prevent the refactoring of linear solver functions to use the matrix row functions (since the matrices passed to a single function must be of the same matrix policy type). Plus, fewer template parameters is always nice. I don't think we ever exposed these through the MUSICA APIs (they have default values), so their removal shouldn't cause downstream problems (I think).


boulderdaze (approved):

K20shores (approved):
Love it when we can delete things

sjsprecious (approved):
@mattldawson I vaguely recalled that this change had little effect on the overall performance, plus we have the in-place algorithm that runs faster now. Thus it should be safe to remove these policies. Thanks for working on this.

PR #1047, opened 2026-07-27
Fix cumulative alpha in multi-step non-in-place LU path

dwfncar:
- non-in-place LU path reuses alpha-shifted Jacobian across attempts of a rejected step; shifts diagonal by delta alpha only
- last_alpha stored the per-attempt delta, should be cumulative shift in the matrix
- third and later attempts of a step factored [(alpha_k + alpha_1) M - J] instead of [alpha_k M - J]
- corrupted stages feed go into the error estimate; so wrong steps silently accepted or could increase rejection frequency
- introduced in #691; in-place and CUDA builders unaffected
- fix: track cumulative shift in last_alpha


sjsprecious (approved):
Thanks @dwfncar for catching this bug. I think at some point we will remove the non-in-place version, so hopefully we can get rid of the whole logic here.

boulderdaze (approved):

K20shores (approved):
what a surprisingly simple fix

K20shores (changes requested):
Are there any tests we could add to catch this?

K20shores (approved):
Once the clang-tidy issues are fixed, ready to merge

PR #1048, opened 2026-07-27
Refactor solver functions to use `MatrixPolicy::Function()`

mattldawson:
This ginormous PR replaces pairs of solver functions specialized to matrix orderings with single functions that use the generalized `MatrixPolicy::Function()`. This is needed for the planned Kokkos-backed solver. The only major development step remaining is to finish the full implementation of Kokkos-backed dense and sparse matrix classes.

I was concerned about maintaining the performance of the original code, so I had Copilot write a benchmark test using the Chapman mechanism with many grid cells that can be used to check the total number of instructions in a full solve with standard and vector-ordered matrix types (vector1, 2, 4, 8, 128). It turned out to be super useful as I went through each function refactor. I could use it to take before and after snapshots to ensure I wasn't adding additional instructions during the refactor. This showed up several needed improvements to the `MatrixPoilcy::Function()` and related functions (described below). I also had Copilot create an action that uses this benchmark to gate-keep PRs. Any PR that introduces additional instructions to the solve (compared to main) will fail the test. You can then decide if the additional instructions are acceptable or not. This PR fails this gate-keeping test. If you look at the action logs you can see the metrics:
```
Run pr/scripts/compare_chapman.py base.txt pr.txt --tolerance "$TOLERANCE"

kind                 base              pr           delta   delta%
FAIL: PR adds instructions to the hot path for: standard
standard      576,025,916     611,346,438     +35,320,522   +6.13%  <-- regression
vector1       713,530,569     613,011,430    -100,519,139  -14.09%
vector128     236,670,620     220,045,464     -16,625,156   -7.02%
vector2       521,633,765     406,257,254    -115,376,511  -22.12%
vector4       391,729,268     314,970,571     -76,758,697  -19.59%
vector8       303,868,717     263,484,576     -40,384,141  -13.29%
Error: Process completed with exit code 1.
```
There are decreases for all but the standard matrix ordering. I can try to hunt these down, but I wasn't sure if it would be worth it since we don't use the standard matrix when we care about performance. But, I can if needed.

I think the refactoring improves the readability of the solver functions (there are lines that almost directly match the pseudo-code algorithms in the class headers).

**Update**
- @sjsprecious would it be possible to run the performance comparison tests for the main branch and this PR?

## Specific Updates
* swapped exception throwing for assertions in matrix functions that end up in the solver hot-path
* introduced `Grouped...` views that bake-in indexing to avoid repeated index calculation
* removed some unneeded multiply/divide steps in indexing math
* introduced `Copy()` and `Fill()` functions that can be used in `MatrixPolicy::Function()` that under-the-hood use `std::copy_n` and `std::fill_n` for large (>18) vector dimensions
* Modified `ForEachRow()` to calculate on padded cells and remove a conditional block; introduced a `ForEachRowStrict()` function that retains the conditional block to exclude padded cells from the calculation when needed (e.g., in reductions like `NormalizedError()`). `ForEachRow()` is faster, so should be used whenever possible.
* removed `Vectorizable` concept. It's no longer needed!

### How I used AI
* I refactored the solver functions by hand (no AI)
* I used Copilot for some debugging and to identify performance regressions
* Copilot wrote the benchmarking test and action, with light review by me
* Copilot wrote the updates to matrix functions and tests, with moderate review by me


docs/performance.md
sjsprecious, line 1: I love the detailed doc, as well as this micro-benchmark!

include/micm/util/sparse_matrix_standard_ordering_compressed_sparse_column.hpp
```diff
@@ -352,17 +463,88 @@ namespace micm
...
+          storage = value;
+      }
+
+      /// @brief Copy a sparse-block value into the caller-owned block-variable temp.
+      template<BlockVariableView Dst, GroupedSparseMatrixBlockView Src>
+      [[gnu::always_inline]]
+      void Copy(Dst&& dst, Src&& src)
+      {
+        auto& storage = dst.Get();
+        if constexpr (requires(std::size_t i) { storage[i]; })
```
K20shores: I'm surpsied clang tidy isn't catching the lack of braces around these statements. It's either not used in any of our tests, or maybe clang tidy is misconfigured.

include/micm/util/sparse_matrix_standard_ordering_compressed_sparse_column.hpp
```diff
@@ -352,17 +463,88 @@ namespace micm
...
+          storage[0] = value;
+        else
+          storage = value;
+      }
+
+      /// @brief Copy a sparse-block value into the caller-owned block-variable temp.
+      template<BlockVariableView Dst, GroupedSparseMatrixBlockView Src>
+      [[gnu::always_inline]]
```
K20shores, line 512: Do we know that this is always better? I guess this almost always results in a larger binary size, with the hopeful trade-off that the lack of a stack function call is likely too expensive on that hot path (which I believe is true). However, it seems possible that the instruction cache could potentially be hurt. It's possible that the number of instructions is smaller, but we could still have a large runtime because we end up waiting for the instruction data. 

I'm not sure if there's an easy way to know for sure

include/micm/util/sparse_matrix_vector_ordering_compressed_sparse_column.hpp
```diff
@@ -450,14 +577,147 @@ namespace micm
...
+        if constexpr (L >= 16)
+        {
+          std::fill_n(view.group_base_ + view.block_offset, L, value);
+        }
+        else
+        {
+          T* dst = view.group_base_ + view.block_offset;
+          for (std::size_t i = 0; i < L; ++i)
+            dst[i] = value;
+        }
```
K20shores, line 610: Do we know why `std::fill_n` isn't always better?

include/micm/util/sparse_matrix_standard_ordering_compressed_sparse_column.hpp
K20shores: i see it did catch this

include/micm/util/sparse_matrix_standard_ordering_compressed_sparse_column.hpp
mattldawson: We don't. This just makes it a little closer to the original implementation, and improved the instruction count and local timing tests (although I didn't explicitly test every function with and without the `always_inline`).

I forgot to put this in the original PR comment, but I was hoping @sjsprecious could run the performance comparison tests before we merge this PR in. I agree that the instruction count and the local timing estimates I used during development are only rough approximations.

include/micm/util/sparse_matrix_vector_ordering_compressed_sparse_column.hpp
mattldawson: This one was just based on the instruction count and local timing tests. Vector-ordered matrices with `L=1,2,4,8` all did better with the `for` loop, but `L=128` did better with `std::fill_n`. This is another case where the full performance comparison tests might help

include/micm/process/process_set.hpp
```diff
@@ -359,296 +352,159 @@ namespace micm
...
-      auto cell_state = state_variables[i_cell];
-      auto cell_forcing = forcing[i_cell];
-      auto react_id = reactant_ids_.begin();
-      auto prod_id = product_ids_.begin();
-      auto yield = yields_.begin();
-
-      for (std::size_t i_rxn = 0; i_rxn < number_of_reactants_.size(); ++i_rxn)
+    const std::size_t n_rxn = number_of_reactants_.size();
+    DenseMatrixPolicy::Function(
+      [this, n_rxn](auto&& forcing_view, const auto&& state_view, const auto&& rc_view)
```
sjsprecious, line 356: So I will use the `AddForcingTerms` function as an example to show my ignorance and better understand what you are doing here.

First question: what are `forcing_view`, `state_view` and `rc_view` here? How are they differing from the original `forcing` and `state` variables?

include/micm/process/process_set.hpp
```diff
@@ -359,296 +352,159 @@ namespace micm
...
+        for (std::size_t i_rxn = 0; i_rxn < n_rxn; ++i_rxn)
         {
-          const std::size_t row_id = react_id[i_react];
-          if (!is_algebraic_variable_[row_id])
+          // Calculate the rate as the rate constant times the product of all reactant concentrations
+          rc_view.Copy(rate, rc_view.GetConstColumnView(i_rxn));
+          for (std::size_t i_react = 0; i_react < number_of_reactants_[i_rxn]; ++i_react)
           {
-            cell_forcing[row_id] -= rate;
+            state_view.ForEachRow(
```
sjsprecious, line 368: Please correct me if I misunderstood anything here (I am almost sure I did):

- for the vector matrix layout, the `ForEachRow` function is likely to loop over the vector length and calculate the rate? Here `rate` is a matrix rather than a scalar? Where is the loop over the number of group/block?

- for the standard matrix layout, the loop over number of the grid cells should be the outermost but now it is gone. How does this new implementation make it happen equivalently?

Feel free to point me to the actual implementation which may help me understand better.

sjsprecious (commented):
@mattldawson I like this work, especially when it simplifies lots of interface.

I do not feel qualified to judge whether the implementation is correct or not here, so I am fine as long as all the existing CI tests pass.

I have some specific questions about the new implementation and hopefully it won't take too much time for you to clarify. It will help me gain a better idea about how the Kokkos work will proceed.

docs/performance.md
mattldawson: I'm glad! I really liked using it during the refactor. I think it also would be pretty easy to copy into and use in other repos.

include/micm/process/process_set.hpp
mattldawson: They are "views" into the matrix. The lambda that is passed to `DenseMatrixPolicy::Function()` is called one or more times, depending on the matrix ordering.

For a standard-ordered matrix with 10 rows, the lambda is called 10 times (once per row), and the views are of a single row of the matrix. For a vector-ordered matrix with L=5 and 10 rows, the lambda is called twice, and the views are of 5-row chunks of the matrix data.

You can think of `DenseMatrixPolicy::Function()` as replacing the outer loop in the forcing functions.
For standard-ordered matrices: https://github.com/NCAR/micm/blob/72d41e1605cfac8be638bbfba07c6659ffb8a907/include/micm/process/process_set.hpp#L364-L366

For vector-ordered matrices: https://github.com/NCAR/micm/blob/72d41e1605cfac8be638bbfba07c6659ffb8a907/include/micm/process/process_set.hpp#L429-L431

The lambda function takes the place of the contents of that outer loop.

include/micm/process/process_set.hpp
mattldawson: It's actually a little different than that. (See comment above about `DenseMatrixPolicy::Function()` replacing the outer loop in the forcing function first).

The `ForEachRow()` function operates on each row in the view. For standard-ordered matrices, the views are of a single row and so `ForEachRow()` just operates on a single set of scalars (the row elements) and there is no looping. For vector-ordered matrices, `ForEachRow()` operates on each row in the L-sized set of row of the view.

For standard-ordered matrices, `ForEachRow()` takes the place of lines like this: https://github.com/NCAR/micm/blob/72d41e1605cfac8be638bbfba07c6659ffb8a907/include/micm/process/process_set.hpp#L382

For vector-ordered matrices, `ForEachRow()` takes the place of lines like this: https://github.com/NCAR/micm/blob/72d41e1605cfac8be638bbfba07c6659ffb8a907/include/micm/process/process_set.hpp#L450-L453
(including the loop over cells in the L-sized group)

mattldawson: For standard-ordered dense matrix, here is the implementation: https://github.com/NCAR/micm/blob/8dd08b2b64750cd01ad09802c08cbc086dee468a/include/micm/util/matrix.hpp#L758-L763

...and for vector-ordered dense matrix: https://github.com/NCAR/micm/blob/8dd08b2b64750cd01ad09802c08cbc086dee468a/include/micm/util/vector_matrix.hpp#L779-L801

The vector-ordered implementation has a compile-time check of whether view are of vector-like things (not our matrices). In this case it has to check for the exact end of the rows in the matrix because the vectors probably aren't exact multiples of the vector dimension (L). When all the views are matrices, it just operates on all cells (including any padded cells to fill out the last L group) This is faster bc `L` is known at compile time, but the actual number of rows of the matrix is not.

include/micm/util/matrix.hpp
```diff
@@ -452,14 +436,7 @@ namespace micm
...
     {
-      if (column_index >= y_dim_)
-      {
-        throw MicmException(
-            MICM_ERROR_CATEGORY_MATRIX,
-            MICM_MATRIX_ERROR_CODE_ELEMENT_OUT_OF_RANGE,
-            "Column index " + std::to_string(column_index) + " out of range for matrix with " + std::to_string(y_dim_) +
-                " columns");
-      }
+      assert(column_index < y_dim_ && "column index out of range");
```
boulderdaze, line 440: I see this `assert` is replacing `MicmException` in quite a few places now. 
This would prevent errors form propagating to the downstream applications that use MICM. 
The current approach of constructing the MicmException is intended to propagate the errors to MUSICA while providing context through `Category()` and `Code()`. 
Errors caught by `assert`  just write to stderr, with just literal strings something like: 
```yaml
Assertion failed: column_index < y_dim_ && "column index out of range"
```
 where `column_index < y_dim` is meaningless to the downstream users, and they would have no context about what went wrong.

mattldawson: I agree that it would be nice to keep the exceptions, but they added considerable (~20% if I remember correctly) overhead to the runtime of the solve. I don't think we could justify this cost for 3D model runs.

If we really wanted to throw exceptions for users when they call these functions, we could maybe template them with something like:
```c++
template<bool Unsafe=false>
void DoSomethingDangerous()
{
  if constexpr (!Unsafe)
  {
    [exception logic]
  }
  ...
}
```
then in the solver functions, we could always call `DoSomethingDangerous<true>();`

I think in the solver functions there mostly wouldn't be a way for a bad user configuration to throw one of these exceptions (all the matrices are created in the state builder).

What do you think?

include/micm/process/process_set.hpp
sjsprecious: Thanks @mattldawson for your detailed explanation. That makes much sense now. So can you point me to the place where `DenseMatrixPolicy::Function()` does the magic to decide how many times a lambda function is called?

sjsprecious: Also where are `forcing_view`, `state_view` and `rc_view` defined for this lambda function?

include/micm/util/matrix.hpp
boulderdaze: Ah, that's quite an overhead. I think it's a reasonable trade off, given that we are also supporting the 3D models.

boulderdaze (approved):
Thank you for addressing my review comments. I understand that there is a change coming in, but I will approve this as we discussed in the comment threads.

include/micm/util/sparse_matrix_standard_ordering_compressed_sparse_column.hpp
sjsprecious: @mattldawson I will run the MUSICA performance test for the `main` branch and your PR, and report back the numbers here later.

include/micm/process/process_set.hpp
mattldawson: sorry, I missed this earlier. The arguments passed into `Function()` after the lambda are matrices and/or vectors. Those get turned into views and passed to the lambda function.
https://github.com/NCAR/micm/blob/580cfbf5c685223ab4aa18466162890ce50b41f3/include/micm/process/process_set.hpp#L413-L416

K20shores (approved):

include/micm/process/process_set.hpp
mattldawson: ...and also missed the question above. The looping over L-sized chunks is in `Function()`:
https://github.com/NCAR/micm/blob/6a349af429a7fc1d1d4008b7dc259c4ea0c6a043/include/micm/util/vector_matrix.hpp#L1229-L1289

PR #1049, opened 2026-07-28
Use a consistent way to define variable precision

sjsprecious:
This PR refactors MICM to use consistent type aliases throughout the codebase for floating-point precision and indexing, enabling support for both single and double precision builds. 

**Key Changes:**

- Introduces `Real` type alias (double/float based on `MICM_USE_DOUBLE` CMake option) and `Index` type alias (std::size_t) in `include/micm/util/types.hpp`
- Replaces all hardcoded `double` with `Real` across CPU/GPU solvers, matrices, constraints, and rate constants
- Replaces all hardcoded `std::size_t` with Index for consistency
- Adds precision-specific test cases (float/double) in constraints, differential equations, and Jacobian verification
- Fixes precision-dependent numerical issues in tests (tolerance adjustments, CUDA kernel precision)

Note that some stiff problems are skipped for float precision since it is either not accurate enough to resolve it or requires a too large tolerance.

All the changes are done by Claude under my supervisory.

closed #1029


include/micm/util/jacobian_verification.hpp
```diff
@@ -176,24 +183,26 @@ namespace micm
   JacobianComparisonResult CheckJacobianSparsityCompleteness(
       const SparseMatrixPolicy& analytical_jacobian,
       const DenseMatrixPolicy& fd_jacobian,
-      std::size_t num_species,
-      double threshold = 1.0e-6)
+      Index num_species,
+      // The threshold must sit above finite-difference roundoff on a genuinely-zero entry, which is
+      // far larger in float; otherwise a true zero is misreported as an undeclared dependency.
+      Real threshold = std::is_same_v<Real, double> ? static_cast<Real>(1.0e-6) : static_cast<Real>(1.0e-2))
```
K20shores, line 189: `1.0e-2` seems very large

include/micm/util/matrix.hpp
```diff
@@ -26,7 +27,7 @@ namespace micm
   };
 
   /// @brief A 2D array class with contiguous memory
-  template<class T = double>
+  template<class T = Real>
```
K20shores, line 30: I wonder if we even need a templated type here now

include/micm/util/vector_matrix.hpp
```diff
@@ -28,7 +29,7 @@ namespace micm
   ///
   /// The template arguments are the type of the matrix elements and the size of the number
   /// of rows per group.
-  template<class T, std::size_t L = MICM_DEFAULT_VECTOR_SIZE>
+  template<class T, Index L = MICM_DEFAULT_VECTOR_SIZE>
```
K20shores, line 32: same here, I wonder if we need the templated type anymore

include/micm/util/matrix.hpp
K20shores: oh wait we do for our tests. Some of our tests use `std::string` to make a point. Hmm. We could consider removing those I guess

test/integration/test_equilibrium_constraint.cpp
```diff
@@ -91,8 +94,12 @@ TEST(EquilibriumIntegration, SetConstraintsAPIWorks)
   // Verify UpdateStateParameters works
   solver.UpdateStateParameters(state);
 
-  double expected_K_eq = ComputeEquilibriumConstant(K_eq, -2400.0, 300.0);
-  EXPECT_NEAR(state.custom_rate_parameters_[0][B_C_eq_idx], expected_K_eq, 1e-10);
+  micm::Real expected_K_eq = ComputeEquilibriumConstant(K_eq, -2400.0, 300.0);
+  // Scale the tolerance by machine epsilon so double keeps ~1e-16 strictness while float passes.
```
K20shores, line 98: I don't know what this means

test/tutorial/CMakeLists.txt
```diff
@@ -5,8 +5,19 @@ create_standard_test(NAME README_example SOURCES test_README_example.cpp)
...
+# overflows float (FLT_MAX = 3.40e38). The parameter becomes +inf, the falloff kernel returns NaN,
+# and every solve reports NaNDetected. The magnitude is inherent to the unit conversion for a
+# third-order reaction, so the mechanism cannot be expressed in single precision without rescaling
+# the tutorial -- and rescaling would invalidate the concentration table published alongside it in
+# docs/source/user_guide/. Restrict them to double-precision builds instead.
+if(MICM_USE_DOUBLE)
+  create_standard_test(NAME rate_constants_no_user_defined_example_by_hand SOURCES test_rate_constants_no_user_defined_by_hand.cpp)
+  create_standard_test(NAME rate_constants_user_defined_example_by_hand SOURCES test_rate_constants_user_defined_by_hand.cpp)
+endif()
+
```
K20shores: I'm reasonably confident that I wrote these tutorials and chose random values. I'd prefer we use realistic values rather than removing the tests on single precision

test/unit/solver/test_linear_solver_in_place_policy.hpp
```diff
@@ -121,22 +133,34 @@ void TestDenseMatrix()
...
+      });
 }
 
 template<class MatrixPolicy, class SparseMatrixPolicy, class LinearSolverPolicy>
-void TestRandomMatrix(std::size_t number_of_blocks)
+void TestRandomMatrix(micm::Index number_of_blocks)
 {
+  // Ill-conditioned dense lognormal test matrices: a single-precision solve legitimately loses too
+  // many significant digits for the reconstruction check to hold. Skipped in single precision (for
+  // both CPU and GPU callers); reconstruction accuracy is verified in the double-precision build.
```
K20shores, line 151: Does this mean that random test systems can't be solved, but that somehow chemistry systems are often conditioned and don't face this issue?

test/unit/solver/test_rosenbrock.cpp
```diff
@@ -151,8 +156,10 @@ TEST(RosenbrockSolver, CanSetTolerances)
   auto foo = micm::Species("foo");
   auto bar = micm::Species("bar");
 
-  foo.SetProperty("absolute tolerance", 1.0e-07);
-  bar.SetProperty("absolute tolerance", 1.0e-08);
+  // Property setters are typed on micm::Real; a bare double literal would be an unsupported type in a
+  // single-precision (Real=float) build and throw "Species: ...". Store the tolerances as micm::Real.
```
K20shores: This comment only appears here, but we do this everywhere. Please remove this comment.

include/micm/util/types.hpp
```diff
@@ -0,0 +1,20 @@
...
+
+namespace micm
+{
+
+// Real-precision typedef.  Default ON via the MICM_USE_DOUBLE cmake option.
+#ifdef MICM_USE_DOUBLE
+  using Real = double;
+#else
+  using Real = float;
+#endif
```
K20shores: Because this is a compile-time value, someone including micm headers directly without cmake would now be getting single precision by default instead of double precision

K20shores: maybe

```
#if !defined(MICM_USE_DOUBLE)
    #error "MICM_USE_DOUBLE must be defined (link the micm CMake target)"
#endif
```

or invert it to an opt-in MICM_USE_SINGLE so the header default matches the CMake default.

include/micm/cuda/util/cuda_param.hpp
```diff
@@ -2,66 +2,68 @@
...
   ProcessInfoParam* jacobian_process_info_ = nullptr;
-  std::size_t* jacobian_reactant_ids_ = nullptr;
-  std::size_t* jacobian_product_ids_ = nullptr;
-  double* jacobian_yields_ = nullptr;
-  std::size_t* jacobian_flat_ids_ = nullptr;
+  micm::Index* jacobian_reactant_ids_ = nullptr;
+  micm::Index* jacobian_product_ids_ = nullptr;
+  micm::Real* jacobian_yields_ = nullptr;
+  micm::Index* jacobian_flat_ids_ = nullptr;
   uint8_t* is_algebraic_variable_ = nullptr;
```
mattldawson: Should you add a
```
using Bool = uint8_t;
```
also?

mattldawson (approved):
Looks good! this was a `Real`ly long PR.

Side note: I think the number of tests that needed to be tweaked or skipped for single-precision is a preview of what it would be like trying to solve real chemistry systems in single-precision

include/micm/cuda/util/cuda_param.hpp
sjsprecious: Thanks @mattldawson for catching it! It should be fixed in https://github.com/NCAR/micm/pull/1049/commits/ebf2967e1f5070248245d6202e9e2291e567bbe6.

include/micm/util/jacobian_verification.hpp
sjsprecious: Correct. And `1.e-6` is also very large for double precision too.

include/micm/util/matrix.hpp
sjsprecious: Good catch. Hmm, this seems like a question I am always puzzled: when to use template variable and when to use a type alias?

test/integration/test_equilibrium_constraint.cpp
sjsprecious: So the original version uses `1e-10` as tolerance, which is fine for double precision but too strict for float precision. This changes make it precision-aware.

test/tutorial/CMakeLists.txt
sjsprecious: I see. I will use a different value for single precision and see if it works.

include/micm/util/matrix.hpp
K20shores: That's why I wondered if we needed the alias. Since we have matrices templated on a datatype, the algorithms using the matrics could use the templated type to define its variables. It's unclear to me whether we really need everything to be single- or double-precision. As this PR stands, we can't do mixed-precision solving, at least not with a lot of granularity. For example, we might be fine with storing all of the data in single precision on the matrices, but maybe we want the algorithms themselves to use local data in double precision.

Not sure what to do with that statement. I think I need more time to think through things

test/unit/solver/test_linear_solver_in_place_policy.hpp
sjsprecious: Well, my understanding here is that with double precision, we can still solve the matrix even if it comes from a random system and is ill-conditioned. But single precision is not accurate enough to provide a reasonable solution.

I do not think the real chemistry system in our model is preconditioned. It just happened to work for decades. Maybe our matrix is diagonal-dominant, which makes the solver more stable?

test/unit/solver/test_rosenbrock.cpp
sjsprecious: Ah good catch. I will remove it.

include/micm/util/matrix.hpp
sjsprecious: Correct. That is my concern after doing this PR, also aligned with Matt's comment above.

If we really can not solve the chemistry system with single precision, most of this work seems unnecessary.

boulderdaze (approved):

include/micm/util/types.hpp
sjsprecious: @K20shores it should be addressed in https://github.com/NCAR/micm/pull/1049/commits/1ff713e924ffdaaaac8514cbe43749e8a20b76f8.

test/unit/solver/test_rosenbrock.cpp
sjsprecious: removed in https://github.com/NCAR/micm/pull/1049/commits/1ff713e924ffdaaaac8514cbe43749e8a20b76f8.

test/tutorial/CMakeLists.txt
sjsprecious: @K20shores Claude found a different set of values and they seemed to work with single precision. See changes at https://github.com/NCAR/micm/pull/1049/commits/1ff713e924ffdaaaac8514cbe43749e8a20b76f8. They look quite different from the original values so let me know if you have any questions about it.

include/micm/util/matrix.hpp
mattldawson: In my experience with CAMP, the most important thing to have high precision for was the production and loss terms in the forcing function. I ended up using separate `long double` vectors to collect production and loss terms separately, then took the difference and returned the net forcing in double precision. In MICM, we don't currently collect the terms separately like this, but if you wanted to try solving in single precision, you might consider implementing that change, with the production and loss terms stored in double precision.

K20shores (approved):
I'm a bit wary of this, only because so much changes for something we seem unable to actually support. However, it's reasonably easy to undo, and it allows us to test everything out.

PR #1054, opened 2026-08-05
Speed up and correct the Oregonator test

K20shores:
The Oregonator sub-step, introduced in #970, was in the shared helper, so all 30 Rosenbrock variants took 30000 solver calls per output interval. It is now a parameter that only backward Euler sets.

The rate constants also did not map onto `stiff/orego/equation.f`, the file that produced the reference values, so the test compared against a different problem with a 9.8e-3 error floor. They are now derived from `s`, `q`, and `w`.

Rosenbrock tolerance 1e-2 -> 1e-6. Backward Euler 2e-2 -> 5e-3. Full `ctest` 55.0 s -> 10.8 s.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/test/integration/test_analytical_backward_euler.cpp
+++ b/test/integration/test_analytical_backward_euler.cpp
@@ -314,28 +314,28 @@ TEST(AnalyticalExamples, HIRES)
 
 TEST(AnalyticalExamples, Oregonator)
 {
-  // The oregonator (k2=1.6e9) is an extremely stiff oscillator with a period of ~48s.
-  // One large step (H=30*tau) causes Newton to converge to the wrong attractor, giving
-  // ~61000x relative error. Sub-stepping at tau/1000 (~0.00016s) lets backward Euler track
-  // the slow manifold with O(H) first-order accuracy (~1.2% relative error).
-  // The output interval is 30 * tau, so 30000 sub-steps give a step size of tau/1000.
+  // The oregonator (k2 = 1.6e9) is an extremely stiff oscillator with a period of about 48 s.
+  // One large step (H = 30 * tau) makes Newton converge to the wrong attractor and gives a
+  // relative error of about 61000. Sub-steps of tau/1000 let backward Euler track the slow
+  // manifold with O(H) first-order accuracy. The output interval is 30 * tau, so 30000
+  // sub-steps give a step size of tau/1000, and a relative error of about 3e-3.
   constexpr micm::Index kOregonatorSubsteps = 30000;
 
-  TestAnalyticalOregonator(backward_euler, 0.02, kOregonatorSubsteps);
-  TestAnalyticalOregonator(backard_euler_vector_1, 0.02, kOregonatorSubsteps);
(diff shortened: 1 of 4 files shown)


test/integration/test_analytical_backward_euler.cpp
```diff
@@ -314,25 +314,28 @@ TEST(AnalyticalExamples, HIRES)
...
-  TestAnalyticalOregonator(backward_euler_vector_mozart_in_place_3, 0.02);
-  TestAnalyticalOregonator(backward_euler_vector_mozart_in_place_4, 0.02);
+  // The oregonator (k2 = 1.6e9) is an extremely stiff oscillator with a period of about 48 s.
+  // One large step (H = 30 * tau) makes Newton converge to the wrong attractor and gives a
+  // relative error of about 61000. Sub-steps of tau/1000 let backward Euler track the slow
+  // manifold with O(H) first-order accuracy. The output interval is 30 * tau, so 30000
+  // sub-steps give a step size of tau/1000, and a relative error of about 3e-3.
+  constexpr micm::Index kOregonatorSubsteps = 30000;
+
+  TestAnalyticalOregonator(backward_euler, 5e-3, kOregonatorSubsteps);
```
sjsprecious, line 324: Looks like the `solver_relative_tolerance` is not provided here and the `1e-9` default value will be used. Is this the reason why you need 30000 substeppings here? Since the overall relative tolerance is `5e-3`, I wonder whether we can use a larger `solver_relative_tolerance` to reduce the number of substeps and further speed up the tests here. Of course if you think the current tests run fast enough, we can discard this comment.

sjsprecious (approved):
Neat! I love any changes that make our unit tests run faster and more accurately.

Just one minor comment but nothing should hold it from being merged.

test/integration/test_analytical_backward_euler.cpp
K20shores: based off of two singular runs

Oregonator runtimes:

| solver relative tolerance | runtime (ms) |
|-------------------------|--------------|
| 1e-9                                  | 2918              |
| 1e-5                                  | 2903              |
| 1e-2                                  | 2904              |

@sjsprecious it doesn't seem to have much of an effect, surprisingly.

sjsprecious: that means the solver converges really fast? Interesting.

K20shores: @sjsprecious yes, that seems true. So now the long time I guess is that it's taking so many substeps

sjsprecious: Do we really need 30000 substeps to get the answer correct here? Or reducing to 3000 leads to an error larger than the tolerance?

K20shores: @sjsprecious we can get down to 18000 steps before we have to drop the tolerance any lower. That runs in 1806 ms

sjsprecious: Neat! 1.6x speedup within ~30 minutes.

PR #1075, opened 2026-08-21
Fix build, CI and packaging for Kokkos deps

### LINKED ISSUE #1073: Build, CI, packaging improvement for Kokkos deps
When the PR https://github.com/NCAR/micm/pull/967 is merged, Claude found some issues about CI workflows:

- **packaging/CMakeLists.txt** is untouched. `micm_kokkos` is never installed or exported, and `micm` now interface-links `Kokkos::kokkos` with no `find_dependency(Kokkos)` in **micmConfig.cmake.in** — so a Kokkos-enabled install is unconsumable downstream.
- `MICM_USE_KOKKOS` is defined only on `micm_kokkos` while `micm` also links Kokkos, so a binary mixing both targets compiles the same header-only inlines with `[=]` in one TU and `KOKKOS_LAMBDA` in another — an ODR violation.
- `gpu-bench-comment.yml` cannot run: missing permissions, wrong artifact name, series names matching no store step.
- `ctest --rerun-failed` in a fresh tree is a no-op, but on a persistent self-hosted runner a stale `LastTestsFailed.log` would silently reduce the run to previously-failing tests only.

sjsprecious:
Closed #1073 

This PR is generated by Claude under my supervisory.

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1,0 +1,43 @@
+MICM -- Model Independent Chemistry Module
+Copyright (C) 2023-2026 University Corporation for Atmospheric Research
+
+This product is licensed under the Apache License, Version 2.0.  See the
+LICENSE file for the full text.
+
+================================================================================
+Third-party components
+================================================================================
+
+MICM does not vendor third-party source in this repository.  The components below
+are fetched at configure time by CMake (see cmake/dependencies.cmake) and are
+governed by their own licenses.  Redistributing a MICM build that includes them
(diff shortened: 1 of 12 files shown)


NOTICE
K20shores: oh, this seems like something we should have had all along

K20shores (approved):

mattldawson (approved):
looks good to me!

boulderdaze (approved):

PR #1080, opened 2026-08-24
Add Kokkos support to external model API

mattldawson:
This PR updates the external model API to allow use of Kokkos-backed matrices with external models.

In draft as a relate MIAM PR is prepared.

### DIFF
--- a/include/micm/solver/solver_builder.hpp
+++ b/include/micm/solver/solver_builder.hpp
@@ -5,6 +5,7 @@
 #include <micm/constraint/constraint.hpp>
 #include <micm/constraint/constraint_set.hpp>
 #include <micm/constraint/types/equilibrium_constraint.hpp>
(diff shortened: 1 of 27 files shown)


include/micm/solver/solver_builder.hpp
```diff
@@ -34,16 +37,22 @@ namespace micm
...
       class RatesPolicy,
       class LuDecompositionPolicy,
       class LinearSolverPolicy,
-      class StatePolicy>
+      class StatePolicy,
+      class... ExternalModels>
   class SolverBuilder
   {
+    // Allow builders of any external-model pack to move state between specializations
+    template<class, class, class, class, class, class, class, class...>
```
K20shores: we really have to figure out if there's a better way to make the solver builder, lol

PR #1081, opened 2026-08-25
Fix Kokkos Kernel GPU performance

### LINKED ISSUE #1074: GPU performance problem for Kokkos functions
When the PR https://github.com/NCAR/micm/pull/967 is merged, Claude found some GPU performance issue about the Kokkos functions implementation.

Parallelism is `league_size = n_cells / L`, with all intra-team work as `TeamThreadRange(team, L)`. Team size is `Kokkos::AUTO` and never tied to `L`, and nothing in the backend ever queries `team_size_max`. The public default is `L = MICM_DEFAULT_VECTOR_SIZE = 4` — a CPU SIMD constant — so `micm::KokkosRosenbrock` launches teams that AUTO will size at ≥32 threads to do 4 lanes of work. Most of every block idles.

Raising `L` doesn't fix it, because `KokkosRowVariable`/`KokkosBlockVariable` are `Kokkos::Array<T, L>` per thread while each thread touches one lane — 1 KB per thread at `L = 128`, which spills. So the layout is squeezed from both ends, and CI benchmarks `vector128` while unit tests only cover `L ∈ {1,2,3,4}`.

Also:
- `AddToDiagonal` rebuilds the host offset table and allocates + uploads a device View on every Backward Euler iteration.
-  `Fill`/`Copy` use fencing `deep_copy` overloads (~8-10 device-wide syncs per attempted step); and a compile-time-constant loop bound is wrapped in a `KokkosScalarView`, costing two Kokkos allocations plus a fencing 8-byte H2D copy per call.

sjsprecious:
Closed #1074 

Also it fixes some broken tests for single precision.

This PR is done by Claude under my supervisory.


include/micm/kokkos/util/kokkos_sparse_matrix.hpp
```diff
@@ -286,6 +292,27 @@ namespace micm
...
+    /// @brief Team size for a launch whose intra-team work is TeamThreadRange(team, L).
+    ///
+    /// Kokkos::AUTO sizes a team from the functor's resource use, which is unrelated to L.
+    /// Since every intra-team loop here is TeamThreadRange(team, L), a team wider than L
+    /// has threads with no iterations to run -- on a GPU that is most of the block sitting
+    /// idle through the team's barriers.  Request L instead, clamped to what the backend
+    /// permits for this functor.  Cached per functor type: the query is a host-side
+    /// occupancy calculation whose answer does not change between launches.
+    template<typename Functor>
+    static int TeamSizeForL(const Functor& team_functor)
```
K20shores: could we share this and the one in the dense matrix? Maybe put it in a different header file?

include/micm/solver/rosenbrock.inl
```diff
@@ -85,7 +85,15 @@ namespace micm
...
-      if (((present_time + 0.1 * H) == present_time) || (H <= parameters.round_off_))
+      // Stagnation guard.  The first clause is the scale-relative test and does the work once
+      // present_time > 0; the second is its counterpart at present_time == 0, where no step is
+      // small enough to be absorbed.  Both are compared against a step size in seconds, so the
+      // floor has to be one too: round_off_ is dimensionless machine epsilon, ~2e-16 in double
+      // (far below any h_min, so the clause never fires) but ~1.2e-7 in single, where on its own
+      // it would abort every problem whose internal step has to go below a tenth of a
+      // microsecond -- h_min is the step floor the integrator actually works to, and steps at
+      // h_min are force-accepted below, so progress is guaranteed either way.
+      if (((present_time + 0.1 * H) == present_time) || (H <= std::min(parameters.round_off_, h_min)))
```
K20shores: interesting. I guess original rosenbrock was only formulated for double precision then

test/integration/analytical_policy.hpp
```diff
@@ -1759,6 +1770,14 @@ void TestAnalyticalOregonator(
     {
       micm::Real dt = std::min(max_substep, time_step - actual_solve);
       auto result = solver.Solve(dt, state);
+      // A solve that ends unconverged returns only a fraction of the sub-step it was given, so
+      // final_time_ is not bounded below by an ulp of actual_solve the way dt is. Once it falls
+      // under half an ulp the accumulator stops moving and this loop spins forever. Break instead,
+      // and let the accuracy assertions below report what the short integration produced.
+      if (actual_solve + result.stats_.final_time_ == actual_solve)
```
K20shores: This is a user error, and I think we should remove this. Everytime we call solve we really should be checking if the solver converged or not. We don't do that in all of our tests because I guess in double precision it didn't matter. Please remove this and all similar code and instead capture the result of solve and check if it converged

K20shores (changes requested):
A couple of code changes. Also please rewrite all of the LLM generated comments that were added. It hurts my brain to read them and I don't really know what it means

include/micm/kokkos/util/kokkos_sparse_matrix.hpp
sjsprecious: Thanks for this comment. I added a new header file at https://github.com/NCAR/micm/pull/1081/commits/0b7fc7f7670e451d632226089d65a90439b59321.

include/micm/solver/rosenbrock.inl
sjsprecious: I guess so.

test/integration/analytical_policy.hpp
sjsprecious: Agreed. I have added a check for the convergence for all similar code at https://github.com/NCAR/micm/pull/1081/commits/0b7fc7f7670e451d632226089d65a90439b59321.

sjsprecious:
> A couple of code changes. Also please rewrite all of the LLM generated comments that were added. It hurts my brain to read them and I don't really know what it means

Sorry about that. I have cleaned up most LLM-generated comments.

K20shores (approved):

PR #1083, opened 2026-08-26
Fix/dae constraint tolerance measure

dwfncar:
## Numerical Issue

- Convergence tested against a fixed absolute residual, max|G| < 1e-10
- The size of a residual depends on how the constraint was written, so
  multiplying a constraint row by a constant changes the answer
- Large row scales never converge. Number-density QSSA constraints hold
  |G| near 1e-8, so Solve returns ConstraintInitializationFailed, 0 steps
- Small row scales report success while the constraints are still
  violated, which the integrator later reports as StepSizeTooSmall

## Fix

- Test the size of the Newton correction instead of the residual:
      scale_a = atol_a + rtol * max(|z_a|, |z_a + delta_a|)
      q       = max_a |delta_a| / scale_a  <=  constraint_init_tolerance_
- Has no units, unaffected by multiplying a constraint row by a constant
- A residual of exactly zero is still accepted without factoring
- The last allowed update is now tested, so a state that satisfies the
  constraints after that update is no longer reported as a failure


dwfncar:
Issued PR
https://github.com/NCAR/micm/pull/1083

sjsprecious (approved):

K20shores (approved):

